PR #28, opened 2015-01-16
Install roles in roles directory.

wizonesolutions:
This is to ensure the desired versions of roles are installed; the user might already
have some old ones in the global role path.

An alternative way to approach this would be to ignore `ansible.cfg` and provide an example along with an explanation. Then people can set `roles_path` (and other options) to whatever they want.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .vagrant/
 vagrant_ansible_inventory_default
 config.yml
+roles
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,0 +1,2 @@
+[defaults]
+roles_path = ./roles


geerlingguy:
Interesting idea... but why not recommend that the global roles be at the latest versions?

wizonesolutions:
People might be using them in other playbooks. If, however, roles respect
semver and ansible-galaxy understands semver, then this could be feasible.

geerlingguy:
I'm going to close this PR for now... I think I'd rather people have the roles installed globally, and I don't want to add anything to install them local to this playbook (for now), unless there's a simple way it can be configured easily, through local variables.

PR #40, opened 2015-03-13
Use quotation marks for property values

mchelen:
It seems like Drush is fine with some non-quoted property values, but has problems with others. Here is an example where I modified this default make file and ended up with problems:
https://gist.github.com/mchelen/0a2aa697ca208e94c684

For the sake of consistency, maybe it is best to use quotation marks for all values?


geerlingguy:
Makes sense. Unquoted floating-point-like numbers can always cause strange issues with different YAML parsers.

PR #64, opened 2015-04-02
Clarify credentials for logging into site

alexfinnarn:
I initially tried "drupal/drupal" for a username and password combo since it was used for MySQL credentials, but I didn't see anywhere where the Drupal admin credentials were mentioned. This final part could be a little confusing for people just getting into setting up VM local development like myself.


geerlingguy:
Thanks, this is good to have in the docs!

geerlingguy:
I added the ability to configure the account credentials in `config.yml` (I don't know why I didn't set it up that way originally!), and shortened the directions in the user guide.

PR #88, opened 2015-05-01
Fix reference to old vars/main.yml file in README.

rocketeerbkw:


geerlingguy:
Wow, how did that slip through all this time? That was a very old change. Thanks for updating!

rocketeerbkw:
Guess it's true nobody reads the README ;)

Issue #39, opened 2015-03-05
Bash command to clone and run


rgoodie:
In my alias.sh script (that auto-loads through my bash profile), I've added a function called `newd` clones this repo, uses sed to change configuration to Drupal 7 if desired, renames the VM to something other than drupaltest.dev, and starts `vagrant up`. Nothing to ground breaking but I thought I would share it since I use this all of the time to spin up new sites. I also don't need to remember the exact repo name. 

Such as `$ newd 7 test-fieldformatter` or `$ newd 8 drupal8test`

https://gist.github.com/rgoodie/9966f30b404a4daa59e1

geerlingguy:
@rgoodie - Nice! I might have to try the script out a couple times and offer some suggestions for improvment (especially to make sure it doesn't need modifications when certain things (like the drush make file) is changed.

I may add a link to it in the README as well, as an option for managing instances of the dev vm :)

rgoodie:
Use as you see fit. Any suggestions welcomed.

geerlingguy:
I've added a page to the Wiki for `newd` specifically, but I'm planning on expanding that page to include a few other demonstration scripts and extensions to help people with a Drupal VM-based workflow.

PR #103, opened 2015-05-19
Centos compatibility

JorisVanEijden:
Add CentOS 6 compatibility.

Tested on CentOS 6.
Not working: phpmyadmin (remi phpmyadmin creates `/etc/phpMyAdmin` folder but role expects `/etc/phpmyadmin`)
Not tested: solr, mailhog

Config.yml changes I needed:
php_opcache_enabled_in_ini: false (see [ansible-role-php #39](https://github.com/geerlingguy/ansible-role-php/issues/39))
xhprof documentroot: `"/usr/share/pear/xhprof_html"`


geerlingguy:
I'm liking this a lot! One initial note (hopefully I'll get more time to go into the individual lines of code later):

I'd like to have a variable in `config.yml` for "OS" or something like that, and set the box based on that. Additionally, we need to switch the box for Parallels too (or just abandon those users for centos ;). That way you still never need to change the Vagrantfile and all changes are in `config.yml`.

JorisVanEijden:
CentOS 7 is failing on issue [ansible-role-mysql #9](https://github.com/geerlingguy/ansible-role-mysql/issues/9)

JorisVanEijden:
Ubuntu 12 is failing due to old ansible (1.4.4) on geerlingguy/ubuntu1204 box (I think).

README.md
```diff
@@ -8,7 +8,7 @@ http://www.drupalvm.com/
 
 This project aims to make spinning up a simple local Drupal test/development environment incredibly quick and easy, and to introduce new developers to the wonderful world of Drupal development on local virtual machines (instead of crufty old MAMP/WAMP-based development).
 
-It will install the following on an Ubuntu 14.04 linux VM:
+It will install the following on an Ubuntu 14.04 or CentOS 6 linux VM:
```
geerlingguy: Let's just leave it as "Ubuntu 14.04 (by default)"

Vagrantfile
```diff
@@ -18,7 +18,14 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   config.vm.network :private_network, ip: vconfig['vagrant_ip']
   config.ssh.insert_key = false
 
-  config.vm.box = "geerlingguy/ubuntu1404"
+  vconfig['vagrant_os'] ||= "ubuntu14"
```
geerlingguy: For this, I'd rather add a config option `vagrant_box`, and default it to `geerlingguy/ubuntu1404`. That way users could fork my own Packer configurations and create their own unique boxes—this would especially come in handy if they wanted to use Vagrant to provision on other cloud providers too...

So this line would just be `config.vm.box = vconfig['vagrant_box']`

Vagrantfile
```diff
@@ -75,8 +82,14 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   end
 
   # Parallels.
-  config.vm.provider :parallels do |p|
-    p.vm.box = "parallels/ubuntu-14.04"
+  config.vm.provider :parallels do |p, override|
```
geerlingguy: Same thing as above here—just use `vconfig['vagrant_box']`, and Parallels users will just have to specify their own box... maybe we can add a comment to that effect in config.yml, or I could just add a note in the Wiki somewhere.

example.config.yml
```diff
@@ -1,4 +1,10 @@
 ---
+# Vagrant OS
+# Virtual machine operating system.
+# Currently supported values are: centos6 and ubuntu14
+# Experimental values are: centos7 and ubuntu12
+vagrant_os: ubuntu14
```
geerlingguy: See notes above - we'll want to use `vagrant_box` here.

provisioning/tasks/init-redhat.yml
```diff
@@ -0,0 +1,29 @@
...
+  yum: "name={{ item }} state=installed"
+  with_items:
+    - "curl"
+    - "python-pycurl"
+    - "@development"
+
+- name: Install postfix so Drupal can send emails.
+  yum: name=postfix state=installed
+
+# Use remi yum repository for PHP.
```
geerlingguy: I have a `geerlingguy.repo-remi` role to install the remi repo; so I'd prefer to just add that in the main playbook conditionally, instead of adding the extra steps here.

provisioning/tasks/drupal.yml
```diff
@@ -13,14 +13,14 @@
 
 - name: Generate Drupal site with drush makefile.
   command: >
-    drush make -y {{ drush_makefile_path }} --no-gitinfofile
+    {{ drush_path }} make -y {{ drush_makefile_path }} --no-gitinfofile
```
geerlingguy: Is this a new variable? I'm fuzzy on it, but if it's definitely defined in all circumstances, I'm okay with using the var here. Otherwise, this seems unrelated to the overall PR.

provisioning/playbook.yml
```diff
@@ -38,7 +40,8 @@
 
   tasks:
     - include: tasks/www.yml
-    - include: tasks/mysql.yml
+    - include: tasks/apparmor.yml
```
geerlingguy: +1

prod/example-prod.config.yml
```diff
@@ -1,4 +1,9 @@
 ---
+# Vagrant OS
+# Virtual machine operating system.
+# Currently supported values are: centos6 and ubuntu14
+vagrant_os: ubuntu14
```
geerlingguy: Same as note above in config.yml

example.config.yml
```diff
@@ -73,8 +79,10 @@ installed_extras:
   - xdebug
   - xhprof
 
-# Add any extra apt packages you'd like to install.
+# APT: Add any extra apt packages you'd like to install.
 extra_apt_packages: []
```
geerlingguy: Let's just change this to `extra_packages`. We can do a minor version bump on the overall project, and people will then be able to choose to update the variable if they're using it already.

Then update the other included files to use `extra_packages`.

geerlingguy:
I finally had some more time to run through all the changes, and it's looking very good—can you make the rest of the changes I noted, and then I'll work on merging and touching up anything I can to get Ubuntu 12.04 and CentOS 7 working sooner rather than later?

This is one small missing piece that's going to make Drupal VM waaaay more awesome-r!

provisioning/tasks/drupal.yml
JorisVanEijden: drush_path comes from the drush role. I had to use it on CentOS 6 because drush was not in the path.

geerlingguy: Sounds good!

geerlingguy:
Since this PR was a bit off from master, I grabbed your patch, applied it, fixed all the conflicts, tested everything further, then went ahead and added a new Wiki page, where we can track specific config changes required per OS/box: [Using Different Base OSes with Drupal VM](https://github.com/geerlingguy/drupal-vm/wiki/Using-Different-Base-OSes).

geerlingguy:
CentOS 7 should soon work well, too... final hurdle is almost complete: https://github.com/geerlingguy/ansible-role-mysql/pull/38

PR #105, opened 2015-05-21
Improved example

JorisVanEijden:
Getting a working copy of the drupal project means you always get the default branch.
You should never specify `version` on git downloads, only branches or commits.


geerlingguy:
Thanks! I'm going to add another line of docs with specifics for how to change core version in the download section too.

Issue #16, opened 2014-11-05
Can't find role php-pecl


rgoodie:
This morning I re-grabbed this (via `git reset --hard && git pull`). Upon launching a provision (`vagrant up && vagrant provision`) I got the following error. 

```
~/vagrant/drupal-dev-vm$ vagrant provision
[default] Running provisioner: ansible...
ERROR: cannot find role in /home/r/vagrant/drupal-dev-vm/provisioning/roles/geerlingguy.php-pecl or /home/r/vagrant/drupal-dev-vm/provisioning/geerlingguy.php-pecl or /etc/ansible/roles/geerlingguy.php-pecl
Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.
```

The Drupal make file was left at default 8 settings (untouched)

PR #110, opened 2015-05-24
Move requirements.txt into provisioning directory.

JorisVanEijden:
Move requirements.txt into provisioning directory.
Do not merge until [JJG-Ansible-Windows pull #16](https://github.com/geerlingguy/JJG-Ansible-Windows/pull/16) is merged and the version in this repo is updated from upstream.


geerlingguy:
Had to rewrite the patch because of other changes, but thanks, this makes the root directory that much nicer!

PR #111, opened 2015-05-26
Add Adminer as an alternative db tool.

m1n0:
Hi I think it would be nice to have [Adminer](http://www.adminer.org) as a lightweight one-file alternative to PHPmyadmin - this should be easy enough as there already ansible role for this and it requires no configuration at all.

### DIFF
--- a/example.config.yml
+++ b/example.config.yml
@@ -76,6 +76,7 @@ installed_extras:
   # - solr
   - xdebug
   - xhprof
+  # - adminer
 
 # Add any extra apt packages you'd like to install.
 extra_apt_packages: []
@@ -126,3 +127,6 @@ php_xdebug_default_enable: 0
 solr_version: "4.10.4"
 solr_xms: "64M"
 solr_xmx: "128M"
+
+# Adminer settings
+adminer_install_dirs: ["/var/www"]
--- a/provisioning/playbook.yml
+++ b/provisioning/playbook.yml
@@ -32,6 +32,7 @@
     - { role: geerlingguy.mailhog, when: '"mailhog" in installed_extras' }
     - { role: geerlingguy.java, when: '"solr" in installed_extras' }
     - { role: geerlingguy.solr, when: '"solr" in installed_extras' }
+    - { role: tersmitten.adminer, when: '"adminer" in installed_extras' }
 
     # Roles for security and stability on production.
     - { role: geerlingguy.security, when: extra_security_enabled }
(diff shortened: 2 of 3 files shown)


geerlingguy:
@m1n0 - Definitely considering this. Especially since the apt packages for phpmyadmin are perpetually broken, and installing from source is a nightmare.

Since I don't use any web-based db admin tool (I prefer Sequel Pro or the good ol' command line for one-offs), I don't have much skin in the game. But I'd consider replacing phpmyadmin entirely, if others think that's a wise move...

m1n0:
Thanks for your feedback - what I really like about Adminer is that it requires no installation or configuration, it is really lightweight, fast and has extremely simple GUI - and all this is packed in one file, which can be easily put into gitignore on a project.

davidlukac:
I prefer Adminer over phpMyAdmin, it's much faster than phpMyAdmin, has clearer interface and it's easier to use. Desktop SWs are great and I use them as well (atm. MySQL Workbench, or connecting directly from PhpStorm), but it’s always good to have something web based at hand out of the box that doesn’t need installation/configuration and is there right away for whoever spins up the VM - last time I checked phpMyAdmin was broken :-/  (`#1146 - Table 'phpmyadmin.pma_recent' doesn't exist`).

geerlingguy:
In that case, I'm thinking of just swapping out phpMyAdmin for Adminer entirely. Looks decent enough to me, and deployment is definitely simpler :)

I'm going to work on testing things so the install is seamless and easily accessible, then I will also update the Wiki.

JorisVanEijden:
+1 for adminer

geerlingguy:
@m1n0 - I'm forking the role you referenced above (see: https://github.com/geerlingguy/ansible-role-adminer) and working on making it play nice with Apache (and hopefully Nginx soon, too), so you can just access `hostname/adminer` instead of having to symlink `adminer.php` all over the place.

m1n0:
@geerlingguy sure, makes sense - /adminer will most probably wont conflict with anything in drupal anyway :)

geerlingguy:
Fixed in above commit—sorry I couldn't figure out a way to easily get part of the authorship assigned to you, @m1n0; but I'm already liking Adminer way better than I ever liked phpMyAdmin... all the same features, but way less bloat and annoying configuration!

m1n0:
No worries, I am glad you like it.

PR #116, opened 2015-06-06
"config.yml missing" even when it exists

opdavies:
When trying to run a Vagrant command (e.g. "vagrant ssh") from a different directory in the project aside from the root one where the Vagrantfile lives, the command will fail even though config.yml exists.

> Path: /Users/opdavies/Code/Current/Personal/dcbristol15/Vagrantfile
Message: Configuration file not found! Please copy example.config.yml to config.yml and try again.

I've added some extra code to Vagrantfile to make sure that we're looking in the correct place for config.yml, rather than in the directory that we're in when we run the command.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -4,10 +4,11 @@ VAGRANTFILE_API_VERSION = "2"
 
 # Use config.yml for basic VM configuration.
 require 'yaml'
-if !File.exist?('./config.yml')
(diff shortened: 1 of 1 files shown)


geerlingguy:
Thanks! I guess my stubborn ways of always changing directory back to the root directory have made it so I didn't catch this :)

PR #147, opened 2015-07-03
Allow Vagrantfile to be loaded from separate script

oxyc:
I wanted to include this repository as a sub directory in my own project and delegate the Vagrantfile from my project root.

```ruby
dir = File.dirname(__FILE__) + '/'
load dir + "drupal-vm/Vagrantfile"
```
Currently this does not work as the provisioning paths are relative.


geerlingguy:
Running some tests locally just to make sure things still work normally in a couple different scenarios, but this does look like a great addition to make the VM even more flexible, thanks!

PR #149, opened 2015-07-04
Add configuration option to allow drush make to overwrite an existing codebase

oxyc:


oxyc:
On second thought, maybe this is unnecessarily flexible? In my own use case where I need this (together with a sql-drop) I could just run my own script before initalizing the provisioning.

geerlingguy:
I'm a little torn on this, as I've also considered adding this kind of thing before, but my main concern is safety here—if someone built up a codebase and didn't have it in Git, or completely encapsulated in their makefile, I don't want them to accidentally overwrite it.

But it might be worth adding an additional provisioning playbook that wipes the contents of the Drupal VM generated codebase folder _and_ empties the generated database, that could be easily run within or outside the VM for more quick resetting (I currently just drop the db manually, delete the folder while on the host machine, than `vagrant provision` again to regenerate a copy of the site).

oxyc:
Aye, I currently do the same in my project `Makefile`.

But I like the idea of including a separate playbook for wiping–in the process of playing around I _actually_ managed to wipe out my entire repo with a faulty `make`-target. Luckily I had everything in git. Simply to prevent accidents from your own solutions, an example playbook would be good :+1:

PR #150, opened 2015-07-04
Move up installation of any extra packages

oxyc:
So that you for example could install `unzip` and extract zip files in your drush makefile.


geerlingguy:
Makes sense to me.

PR #151, opened 2015-07-04
Add drupal install arguments configuration option

oxyc:
Would be useful to allow additional arguments to be passed to the `drush site-install` command.

In my case I'd like to install the commerce kickstart distribution without the demo store.

### DIFF
--- a/example.config.yml
+++ b/example.config.yml
@@ -49,6 +49,9 @@ drupal_mysql_user: drupal
 drupal_mysql_password: drupal
 drupal_mysql_database: drupal
 
+# Additional arguments or options to pass to `drush site-install`.
+drupal_site_install_extra_args: []
+
 # Cron jobs are added to the root user's crontab. Keys include name (required),
 # minute, hour, day, weekday, month, job (required), and state.
 drupalvm_cron_jobs: []
--- a/provisioning/tasks/install-site.yml
+++ b/provisioning/tasks/install-site.yml
@@ -15,6 +15,7 @@
     --account-name={{ drupal_account_name }}
     --account-pass={{ drupal_account_pass }}
     --db-url=mysql://{{ drupal_mysql_user }}:{{ drupal_mysql_password }}@localhost/{{ mysql_databases[0].name }}
+    {{ drupal_site_install_extra_args | default([]) | join(" ") }}
     chdir={{ drupal_core_path }}
   notify: restart webserver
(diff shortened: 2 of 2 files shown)


geerlingguy:
Does this work fine if you have no extra arguments in the list?

oxyc:
I'll manually test it in a bit to verify.

oxyc:
Earlier it bailed out if the variable was `undefined`, now it's fine both if it's `undefined` or `[]`.

oxyc:
@geerlingguy any changes you'd like for this pull request to get accepted?

example.config.yml
```diff
@@ -49,6 +49,11 @@ drupal_mysql_user: drupal
 drupal_mysql_password: drupal
 drupal_mysql_database: drupal
 
+# Additional arguments or options to pass to drush site install.
+drupal_install_arguments: []
```
geerlingguy: Can you rename the variable to `drupal_site_install_extra_args`?

example.config.yml
```diff
@@ -49,6 +49,11 @@ drupal_mysql_user: drupal
 drupal_mysql_password: drupal
 drupal_mysql_database: drupal
 
+# Additional arguments or options to pass to drush site install.
```
geerlingguy: Can you change this to read:

```
# Additional arguments or options to pass to `drush site-install`.
```

provisioning/tasks/install-site.yml
```diff
@@ -15,6 +15,7 @@
     --account-name={{ drupal_account_name }}
     --account-pass={{ drupal_account_pass }}
     --db-url=mysql://{{ drupal_mysql_user }}:{{ drupal_mysql_password }}@localhost/{{ mysql_databases[0].name }}
+    {{ drupal_install_arguments | default([]) | join(" ") }}
```
geerlingguy: (same thing with var name above).

example.config.yml
```diff
@@ -49,6 +49,11 @@ drupal_mysql_user: drupal
...
+drupal_install_arguments: []
+  # - "--locale=uk"
```
geerlingguy: Since this variable will be used infrequently, please leave out the examples here; if enough people ask, we can add it to a new docs page on 'site installation options' or something along those lines, but I'd rather save the two lines in the example config file.

oxyc:
Updated according to your suggestions. I can do a final manual site-install test tomorrow though.

PR #155, opened 2015-07-08
vagrant-auto_network plugin support

shawnachieve:
This update makes a small modification to the Vagrantfile to detect if the user has the vagrant-auto_network plugin installed and uses that to automatically assign an IP address to the VM rather than having to explicitly set an IP in the config.yml file.  This is useful for users that need to run multiple VM as it helps avoid IP address collisions or having to track which IPs are used by which VMs.


geerlingguy:
This looks like a nice improvement and would hopefully resolve some small networking issues some users have (especially when their local area networks use the 192.168.x prefix!). I might modify the documentation a little after merging, and I'll do a little manual testing myself prior to merging. Thanks!

geerlingguy:
Just a note to myself—if/when this is merged, also needs a mention in docs somewhere, maybe under a new networking section (would be good alongside the livereload/forwarded ports configuration too).

PR #157, opened 2015-07-08
Expanded Local Dev config example

shawnachieve:
This PR provides a new example.config.yml file that demonstrates how to configure the VM with the following features:

* Accessing the site at both drupalvm.dev and www.drupalvm.dev
* SSL Support on both drupalvm.dev and www.drupalvm.dev
* Enabled Apache mod_expires so that varnish caching of static files can be tested/configured.
* Pre-configured XDebug to work with PHPStorm or Sublime on the host computer.

The above are basically the changes I find myself making to the default config.yml file every time I spin up a new VM.  So I'm hoping that adding a new example.config.yml file will help out other people that may want these features.


geerlingguy:
@shawnachieve - Could you actually make these changes directly to the normal `example.config.yml` file, and then I can pick which chunks to add to that file; most of the changes should probably be in the general examples file.

shawnachieve:
I've merged my separate example.config.yml file in to the base example.config.yml file.

shawnachieve:
FYI, it looks like using the apache_mods_enabled directive is having some unintended side effects.  When that directive is used to enable mod_expires it seems that it drops several of the default mods that are suppose to be enabled by the apache ancible script (e.g. enabling mod_expires removes mod_ssl and mod_rewrite). 

I'm investigating this further and will update the PR as soon as I figure out what's going on.

example.config.yml
```diff
@@ -64,11 +64,38 @@ configure_local_drush_aliases: true
...
+# Optionally, enable/disable additional apache modules.
+# By default, all modules required by Drupal are enabled, so only
+# Additional modules need to be listed here.
+apache_mods_enabled:
+  - expires.load
+  - ssl.load
+  - rewrite.load
+# apache_mods_disabled: []
+
+# Enable SSL Support on the Drupal site vhosts.
```
geerlingguy: Can you file a separate PR for the SSL vhosts (and remove it from this PR)? I want to make sure that this works correctly across different OSes and provides a seamless experience out of the box, and I'd rather tackle that separately.

geerlingguy:
@shawnachieve - Can you also split out the xdebug changes? Thinking about this further, I'd like to have some more time to do some performance comparison with the defaults before/after, and it might be a case where we add some extra documentation, but don't actually change the defaults (except to add some things that people might want to change _when_ they enable xdebug).

geerlingguy:
@shawnachieve - Actually, scratch that—as I'm working on cleaning up some documentation, I decided to also merge in the changes to `example.config.yml` that I think we should add for now, for XDebug.

So, please remove those changes from this PR, and I think it'll be good to go for a merge!

example.config.yml
```diff
@@ -64,11 +64,25 @@ configure_local_drush_aliases: true
 
 # Apache VirtualHosts. Add one for each site you are running inside the VM. For
 # multisite deployments, you can point multiple servernames at one documentroot.
+# NOTE: We create separate vhosts for www and non-www URLs so that the
+#       hostsupdater plugin will create host file entries for both.
+# NOTE: We can't create a vhost with a servername of  "www.{{ drupal_domain }}"
```
geerlingguy: Unfortunately, this breaks some of the intuitive behavior of the config.yml file, since now users will have to change the domain in multiple places.

For now, I'd be okay with users having to manually add in `www.` for the domain, and I'd like to make it so hostsupdater can get all the domains even if they use variables like `{{ varname }}`.

example.config.yml
```diff
@@ -64,11 +64,25 @@ configure_local_drush_aliases: true
...
   - {servername: "xhprof.drupalvm.dev", documentroot: "/usr/share/php/xhprof_html"}
   - {servername: "pimpmylog.drupalvm.dev", documentroot: "/usr/share/php/pimpmylog"}
 
+# Optionally, enable/disable additional apache modules.
+# NOTE: Uncommenting this block overrides the default set of modules to be
+#       enabled.  So we need to make sure we include mod_ssl and mod_rewrite
+#       along with our additional module(s) such as expires.
+# apache_mods_enabled:
+#  - expires.load
+#  - ssl.load
```
geerlingguy: I think we can just explicitly enable these three mods, since they're required on a large number of sites anyways, and won't cause any problems or degradations if not included.

shawnachieve:
Synced the PR with master and removed the www vhost entry.  Now the PR just contains the apache modules code.

geerlingguy:
Thanks!

Issue #137, opened 2015-06-30
Enhance DrupalVM with Behat testing


nerdstein:
Look at an Ansible Role that brings in Behat testing.

Composer libraries to consider:

    "behat/behat":                  "3.0.*",
    "behat/mink":                   "1.6@stable",
    "behat/mink-extension":         "*",
    "behat/mink-goutte-driver":     "*",
    "behat/mink-selenium2-driver":  "*",
    "behat/mink-zombie-driver":     "*",
    "behat/mink-browserkit-driver": "*",
    "drupal/drupal-extension":      "*",

geerlingguy:
Can you give a little more background in terms of what Drupal VM should do specifically that would be helpful for all Drupal projects (rather than something that would be run locally for a particular Drupal site inside the VM, e.g. running composer install after provisioning and doing any other required setup).

E.g. are there any particular libraries or packages that Drupal VM should support so that Behat testing configuration is simpler for individual Drupal projects?

nerdstein:
This is a great question. My gut tells me we should just install the infrastructure initially and see how people use it. 

I think there is a lot of discussion happening on how to best leverage it. This is just getting the ball rolling.

geerlingguy:
I would need something more concrete to make it a default configuration for Drupal VM; for now, all the mentioned libraries can be installed globally by adding them to the `composer_global_packages` variable.

nerdstein:
Is this variable within config.yml?  How is this set for DrupalVM?

geerlingguy:
@nerdstein - See https://github.com/geerlingguy/drupal-vm/blob/master/example.config.yml#L119-L120 for an example in the default config (it's commented by default).

nerdstein:
Thanks

kevinquillen:
If you just bring in "drupal/drupal-extension" in composer, it itself pulls in all the required libraries necessary for testing Drupal (it didnt used to in previous versions).

kevinquillen:
Be aware though that the Behat 3.0 docs are not completed and there are notable changes from 2.x.

arknoll:
If you want to actually do selenium tests you need to:

1. Download selenium server: 
curl -o selenium-server-standalone-2.46.0.jar http://selenium-release.storage.googleapis.com/2.46/selenium-server-standalone-2.46.0.jar

2. Run selenium server: 
DISPLAY=:1 xvfb-run java -jar selenium-server-standalone-2.46.0.jar

3. Download firefox: 
[...]

PR #161, opened 2015-07-09
Add Selenium and Firefox to vm to power Behat tests

arknoll:
Relates to https://github.com/geerlingguy/drupal-vm/issues/137

### DIFF
--- a/example.config.yml
+++ b/example.config.yml
@@ -93,6 +93,7 @@ installed_extras:
   - varnish
   - xdebug
   - xhprof
+  - selenium
 
 # Add any extra packages you'd like to install.
 extra_packages: []
@@ -155,3 +156,6 @@ php_xdebug_default_enable: 0
 solr_version: "4.10.4"
 solr_xms: "64M"
 solr_xmx: "128M"
+
+# Selenium configuration
+selenium_version: 2.46.0
--- a/provisioning/requirements.txt
+++ b/provisioning/requirements.txt
@@ -1,3 +1,4 @@
+arknoll.selenium
 geerlingguy.adminer
 geerlingguy.apache
 geerlingguy.composer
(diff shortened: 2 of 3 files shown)


example.config.yml
```diff
@@ -155,3 +156,6 @@ php_xdebug_default_enable: 0
 solr_version: "4.10.4"
 solr_xms: "64M"
 solr_xmx: "128M"
+
+# Selenium configuration
+server_version: 2.46.0
```
geerlingguy: Can this be updated (upstream) to a variable with the name of the role prefixed (e.g. `selenium_server_version`)? Similar to Drupal modules, it's best to use a prefixed variable name so there are no conflicts between roles (and to clarify what, exactly, the variable is used for in a large config file).

provisioning/requirements.txt
```diff
@@ -21,3 +21,4 @@ geerlingguy.repo-remi
 geerlingguy.security
 geerlingguy.solr
 geerlingguy.varnish
+arknoll.selenium
```
geerlingguy: I've been adding roles in alpha order - can you stick your role at the top of the file instead? (OCD, I know... ;).

example.config.yml
geerlingguy: Actually, it looks like the variable _is_ `selenium_version` upstream... is that correct?

geerlingguy:
One other question: Would you be willing to (or would you accept a PR from me to) make this work on RHEL/CentOS? Currently all other configuration in the VM works across platforms, so people can switch to other base boxes if they would like, and I don't want to take away that ability if they want to use selenium.

provisioning/requirements.txt
arknoll: Fixed.

example.config.yml
arknoll: yeah, it should have been selenium_version. Fixed.

arknoll:
@geerlingguy let me give it a shot to get it working on RHEL/CentOS.

geerlingguy:
@arknoll - Great! let me know when that's done, and I'll un-postpone this PR.

arknoll:
@geerlingguy Done and tested on CentOS https://galaxy.ansible.com/list#/roles/4340.

geerlingguy:
Great! I'll merge and do some more testing tonight.

geerlingguy:
Verified that behat/selenium/firefox/etc. works great across all supported CentOS and Ubuntu versions.

PR #162, opened 2015-07-09
SSL Config Example

shawnachieve:


example.config.yml
```diff
@@ -69,6 +69,19 @@ apache_vhosts:
...
 
+# Enable SSL Support on the Drupal site vhosts.
+# See https://github.com/geerlingguy/ansible-role-apache for additional options.
+# Update the file paths below if you are using a custom SSL certificate.
+apache_vhosts_ssl:
+  - {
+    servername: "{{ drupal_domain }}",
+#    serveralias: "www.{{ drupal_domain }}",
+    documentroot: "{{ drupal_core_path }}",
+    certificate_file: "/etc/ssl/certs/ssl-cert-snakeoil.pem",
```
geerlingguy: One difficulty here: this cert isn't available on CentOS (at least, not by default), and I'd like to have all examples OOTB work with either Ubuntu or CentOS (latest).

Maybe we can leave this example conf commented? Or at least rewrite comments a little to mention this fact (and add a note to the docs on different OSes).

shawnachieve: I can definitely do a little rewriting of the comments.  We could also just add either a separate example config and/or a wiki page on SSL support as short term fixes.

Since it's almost a given that in the near future all sites are going to need to be SSL encrypted, I think it would make sense to be able to have the drupal-vm support SSL out of the box.  Is there a good Ansible SSL library that you recommend?

What if we modified the provisioning script so that we generate a self-signed SSL certificate for "*.{{ drupal_domain }}" as part of the provisioning and reference that certificate in the vhost config?

aweingarten: @geerlingguy: tested this locally.   Worked like a charm.  

e.g. 
```
 - {servername: "{{ drupal_domain }}", documentroot: "{{ drupal_core_path }}", certificate_file: "/var/www/cert/wildcard.crt", certificate_key_file: "/var/www/cert/wildcard.key"}
```

geerlingguy: @shawnachieve @aweingarten - I like the idea of auto-generating a self-signed cert for {{ drupal_domain }} (at a minimum), but I fear it'll start getting a little more difficult to get it working well across platforms and in a way that the cert is valid everywhere (at least in a short timeframe).

I think, as a first step, I want to add some really good/focused docs to the docs site, then we can work on a more comprehensive solution.

kostajh:
@aweingarten how did you generate the cert in `/var/www/cert`? Would be great to get these docs  updated with a CentOS 7 example and merged

aweingarten:
@kostajh, I simply created the cert manually on my host machine in the `cert` directory and then mapped it into the vm.

oxyc:
This can now be closed https://github.com/geerlingguy/drupal-vm/pull/438.

PR #165, opened 2015-07-10
Add swappiness task

willvincent:
This PR adds a task to allow configuration of the swappiness value of the guest OS. By default, Linux sets this value to 60, which means some swapping will always occur.  Swapping slows down the system, and in a virtual guest environment this very quickly becomes a performance issue. When the guest swaps, everything usually slows down, both inside the guest os and on the host. Setting swappiness to 0 will cause the linux kernel to only swap to avoid out of memory situations, but even a higher value (that is still considerably lower than the default) like 10 will result is a dramatic increase to responsiveness of the virtual machine in most cases -- especially when configured with more limited resources.

### DIFF
--- a/example.config.yml
+++ b/example.config.yml
@@ -156,10 +156,11 @@ solr_version: "4.10.4"
 solr_xms: "64M"
 solr_xmx: "128M"
 
-# By default, linux sets the swappiness level to 60, this means that some swapping will occur.
-# Swapping will slow down any system, but in a virtual environment, when the guest stars
-# swapping, things will become even more slow than normal. As such, setting the swappiness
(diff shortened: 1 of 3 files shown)


willvincent:
For reference: https://lonesysadmin.net/2013/12/11/adjust-vm-swappiness-avoid-unneeded-disk-io/

geerlingguy:
If I add swappiness configuration, I'd rather add that as part of a larger `performance` role rather than piecemeal, and just leave it out by default. For the majority of use cases, swappiness will be one of the last things that needs extra optimization (vs. making sure MySQL and PHP are using good memory settings).

For now, I'd rather approach the idea of allowing extra tasks to be included in the provisioning via a custom tasks file (like `config.yml` and `drupal.make.yml`).

Issue #169, opened 2015-07-11
Move documentation to Read the Docs


geerlingguy:
GitHub's Wiki is a bit of a pain to use, and doesn't really offer much over just maintaining the documentation inside the project itself. Therefore, I'd like to keep the docs in code, and use Read the Docs to display them at the URL: http://docs.drupalvm.com/

  - [x] Create read the docs account
  - [x] Choose a pattern: docs in current codebase, or new `drupal-vm-docs` git project? (I'm favoring docs-in-main-repo, since I hate messy cross-dependencies and outdated docs). (Picked: putting docs into a `docs` directory).
  - [x] Move the documentation pages from the [Wiki](https://github.com/geerlingguy/drupal-vm/wiki) to the codebase
  - [x] [Set up a cname](https://read-the-docs.readthedocs.org/en/latest/faq.html) for the docs (e.g. `docs.drupalvm.com`)
  - [x] [Set up GitHub webhooks](https://read-the-docs.readthedocs.org/en/latest/webhooks.html) so changes are pushed automatically
  - [x] Redirect each wiki page to the new docs pages.
  - [x] Update project README to point to new documentation pages (and add prominent link to docs.drupalvm.com)
  - [x] Update www.drupalvm.com to point to new documentation pages (and add prominent link to docs.drupalvm.com)
  - [x] Organize the docs similar to organization on the Wiki

For structure/setup, see the [mkdocs build documentation](https://read-the-docs.readthedocs.org/en/latest/builds.html#mkdocs).

nerdstein:
+1 for main repo

geerlingguy:
@nerdstein - Sounds good! Updated the issue summary/checkboxes above.

geerlingguy:
@nerdstein - FYI, I've updated the wiki (https://github.com/geerlingguy/drupal-vm/wiki/PHP-7-on-Drupal-VM), please make sure you pull the latest changes :)

geerlingguy:
@nerdstein added docs to the repo as part of https://github.com/geerlingguy/drupal-vm/pull/172, so onward with this issue!

geerlingguy:
This issue is complete. Yay!

PR #172, opened 2015-07-14
Moving docs

nerdstein:
1. Moves docs to `/docs` directory
2. Stubs out mkdocs.yml with sane defaults
3. Renames Home.md to index.md for mkdocs standard


geerlingguy:
Excellent! I'll merge now, then we can work on sorting things and updating docs as part of #169 .

PR #176, opened 2015-07-15
Add easy port forwarding configuration and Livereload documentation

oxyc:
Add support for port forwarding and add documentation about using livereload, as one example for when it can be used.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -52,6 +52,15 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     end
   end
 
+  if vconfig.include?('vagrant_forwarded_ports')
+    for forwarded_port in vconfig['vagrant_forwarded_ports']
+      config.vm.network "forwarded_port",
+        guest: forwarded_port['guest'],
+        host: forwarded_port['host'],
+        protocol: forwarded_port['protocol']
+    end
+  end
+
   for synced_folder in vconfig['vagrant_synced_folders'];
     config.vm.synced_folder synced_folder['local_path'], synced_folder['destination'],
       type: synced_folder['type'],
--- a/example.config.yml
+++ b/example.config.yml
@@ -26,6 +26,9 @@ vagrant_synced_folders:
     type: nfs
     create: true
 
+vagrant_forwarded_ports: []
+  # - { guest: 35729, host: 35729, protocol: "tcp" }
+
 # Memory and CPU to use for this VM.
 vagrant_memory: 1024
 vagrant_cpus: 2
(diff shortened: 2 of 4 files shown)


Vagrantfile
```diff
@@ -33,6 +33,13 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     end
   end
 
+  for forwarded_port in vconfig['vagrant_forwarded_ports']
```
geerlingguy: Can you wrap this entire set of config in an `if` statement and check if `vagrant_forwarded_ports` is a key in the `vconfig` array? For those updating, I don't want `vagrant up` to fail if the variable is not defined in `config.yml`.

And this would allow us more flexibility in terms of not having it in the default config.yml, but still available if someone wants it.

geerlingguy:
Is it not possible to use the same IP address as the VM (just open up the desired port on the firewall) to do livereload straight into the VM, without having to explicitly add forwarded ports in Vagrant?

Vagrantfile
oxyc: :+1:

oxyc:
Hmm true, I guess it works in the case
- You use the Drupal module (loading `http://192.168.88.88:35729`).
- You run the livereload server on the host machine.

In my scenario I'm running grunt on the VM (as I don't want to require node/ruby for our teams dev computers), and in my cases I see the drupal module as a useless dependency when you only need port forwarding.

But I've been so set it my ways, maybe this is not a common scenario and the documentation isn't needed.

oxyc:
@geerlingguy do you want me to rebase without the livereload docs/references?

example.config.yml
```diff
@@ -22,6 +22,12 @@ vagrant_synced_folders:
     type: nfs
     create: true
 
+# A list of forwarded ports, which allow you to access a port on your host
```
geerlingguy: Since this seems like a less-common use case, let's leave the extra explanation out, and just leave `vagrant_forwarded_ports: []` in the file, with one example line commented out (that way it's not explicitly targeted at livereload).

example.config.yml
```diff
@@ -117,6 +123,7 @@ firewall_allowed_tcp_ports:
   - "8080"
   - "8443"
   - "8983"
+  - "35729"
```
geerlingguy: Let's leave the default set of ports as-is, and just make sure the documentation page for Livereload mentions adding the specified port to the firewall allowed TCP ports.

oxyc:
Updated the logic commit according to your suggestions. Docs still need to be revised, guess they should really be about port forwarding.

oxyc:
@geerlingguy want me to revise the docs or just leave this feature out? I personally don't use it anymore and as no one else has requested in half a year it might just be an unnecessary feature.

geerlingguy:
@oxyc - Let's just leave it out—it's one of those frontend things where now there are 20 different ways (gulp, grunt, livereload... use Ruby, use Node.js, etc.), and I'd rather let devs figure out how they want to do it on their own.

PR #178, opened 2015-07-16
add support for custom synced folder mount options

oxyc:
I read your article about rsync vs nfs–a very interesting read. I tried switching to it but in the end I decided to stick to NFS due to the long sync delays (annoying while editing css for example).

To get instant sync (and hopefully some other performance benefits) I needed to add the following mount options to the synced_folders `nolock,vers=3,udp,noatime,actimeo=1'`

Hence this patch.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -40,7 +40,8 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       rsync__exclude: synced_folder['excluded_paths'],
       rsync__args: ["--verbose", "--archive", "--delete", "-z", "--chmod=ugo=rwX"],
       id: synced_folder['id'],
-      create: synced_folder.include?('create') ? synced_folder['create'] : false
+      create: synced_folder.include?('create') ? synced_folder['create'] : false,
+      mount_options: synced_folder.include?('mount_options') ? synced_folder['mount_options'] : []
   end
 
   if is_windows


shawnachieve:
FYI, we use another vagrant plugin to handle rsync mounts: https://github.com/smerrill/vagrant-gatling-rsync

This plugin is much more efficient/faster at syncing a large codebase like Drupal than the built-in rsync handler.  Typically code changes are synced into the vm with this plugin in about 0.25 seconds (usually faster than I can hit save and switch over to my browser to test the update).

I haven't tried this plugin yet with the Drupal VM (we're using it on our custom vagrant environments).  Here is the code we have in our Vagrantfile for the sharing:

```
  for synced_folder in vconfig['vagrant_synced_folders'];
    excluded_paths = synced_folder['exclude_paths']
    # puts "RSync excluded_paths: #{excluded_paths}"
    config.vm.synced_folder synced_folder['local_path'], synced_folder['destination'],
      type: synced_folder['type'],
      rsync__auto: "true",
      rsync__exclude: excluded_paths,
      rsync__args: ["--verbose", "--archive", "--delete", "-z", "--chmod=ugo=rwX"],
      id: synced_folder['id']
  end

  if Vagrant.has_plugin?("vagrant-gatling-rsync")
    config.gatling.latency = 2.5
    config.gatling.time_format = "%H:%M:%S"
  end  

  # Turn on auto-rsync after the VM boots.
  {
    [:up, :resume, :reload] => "vagrant gatling-rsync-auto",
  }.each do |command, trigger|
    config.trigger.after command, :stdout => true do
     info "Executing #{trigger}"
     run "#{trigger}"
    end
  end
```

oxyc:
@shawnachieve thanks! It improved my performance quite a bit. In the end I settled with having my gulp watch tasks rsync The changed files specifically, and now it's as fast as native.

I still think it's worth having this pull request though. Mount options are even mentioned in the docs.

geerlingguy:
Thanks! And @shawnachieve - Would you mind opening a new issue (or filing a PR, if you feel so inclined) to suggest adding support for gatling rsync? I may be willing to add support to Drupal VM, as it looks like it could be an even better option than NFS.

shawnachieve:
Yup. I'll open a PR once I've had a chance to test and verify the appropriate config for use with drupal-vm.

PR #182, opened 2015-07-28
Use drupal_domain for extra vhosts

opdavies:
Use the drupal_domain variable to the extra vhost servernames.

Also, added some extra spaces to make it easier to read. :)

### DIFF
--- a/example.config.yml
+++ b/example.config.yml
@@ -66,8 +66,8 @@ configure_local_drush_aliases: true
 # multisite deployments, you can point multiple servernames at one documentroot.
 apache_vhosts:
   - {servername: "{{ drupal_domain }}", documentroot: "{{ drupal_core_path }}"}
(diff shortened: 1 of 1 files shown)


geerlingguy:
Unfortunately, when we do this, it makes it so the domains aren't added properly via vagrant-hostsupdater... can you think of a simple way to get hostsupdater to work with this style of variable-built subdomains?

shawnachieve:
I just added a PR that allows this to be supported: https://github.com/geerlingguy/drupal-vm/pull/184

opdavies:
@geerlingguy Interesting. I hadn't noticed this before.

@shawnachieve Great. I'll check it out. :)

geerlingguy:
I'm going to move forward with @shawnachieve's PR, and close this one. Thanks for reminding me to keep the variables flexible/consistent, @opdavies !

PR #187, opened 2015-08-03
Update documentation for xhprof and pimpmylog urls

justAChris:
Updates the documentarians for xhprof and pimpmylog extras. The vhost aliases have changed at some point from the http://local. urls and the supporting documentation should match.


geerlingguy:
Nice catch!

PR #188, opened 2015-08-03
Avoid Drupal installer exception for xdebug nesting level

weitzman:
```
Exception 'Drupal\Core\Installer\Exception\InstallerException' with message 'Xdebug settings: xdebug.max_nesting_level is set to 255.

Set xdebug.max_nesting_level=256 in your PHP configuration as some pages in your Drupal site will not work when
this setting is too low.' in /var/www/drupal/core/includes/install.core.inc:2292
```


fluxsauce:
Having explicitly made this configuration change on other systems, :shipit: IMHO

geerlingguy:
Haha, wow... such a specific integer. Merging shortly.

PR #190, opened 2015-08-04
Fix formatting and add note about permission settings on some log files.

sidneygijzen:
I was configuring PML and stumbled upon some small issues. 

Regarding the added note about permissions; I wasn't sure on how you would like to handle this. I chose to add a note to the docs, because I think it's something very depending on one's purpose of using Drupal VM (as a local dev env, mimic a production env, etc).


geerlingguy:
Thanks, looks great!

PR #191, opened 2015-08-06
Set example mysql_wait_timeout to 300.

cthos:
## Motivation
mySQL's default wait timeout is crazy high. (see also http://www.eliotk.net/12/21/mysql-wait_timeout-default-is-set-too-high/)

No need to keep connections open for 8 hours by default.

## Notes
I thought about putting this in the mysql playbook as a default, but since it _is_ mySQL's default, I figured this makes more sense.

### DIFF
--- a/example.config.yml
+++ b/example.config.yml
@@ -143,6 +143,7 @@ composer_home_path: '/home/vagrant/.composer'
 mysql_root_password: root
 mysql_slow_query_log_enabled: true
 mysql_slow_query_time: 2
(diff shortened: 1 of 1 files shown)


geerlingguy:
For a typical LAMP stack, that timeout is pretty insane... but for many applications that persist connections to MySQL (Python, Node, Java, etc... even a few small PHP apps I've worked with), that timeout isn't _too_ outlandish. But it's still way higher than I normally set.

This change looks good to me!

Issue #186, opened 2015-08-02
apache locale error?


mray:
Being totally new to this i followed the installation procedure up to the command `vagrant up`, `vagrant provision` does not seem to help.  
Up until this error fills my console:

```
TASK: [geerlingguy.apache | Ensure Apache is installed.] ********************** 
failed: [drupalvm] => (item=apache2,apache2-mpm-prefork,apache2-utils) => {"failed": true, "item": "apache2,apache2-mpm-prefork,apache2-utils"}
stderr: perl: warning: Setting locale failed.
perl: warning: Please check that your locale settings:
	LANGUAGE = "en_US:",
	LC_ALL = (unset),
	LC_TIME = "de_DE.UTF-8",
	LC_MONETARY = "de_DE.UTF-8",
	LC_CTYPE = "C",
	LC_ADDRESS = "de_DE.UTF-8",
	LC_TELEPHONE = "de_DE.UTF-8",
	LC_NAME = "de_DE.UTF-8",
	LC_MEASUREMENT = "de_DE.UTF-8",
	LC_IDENTIFICATION = "de_DE.UTF-8",
	LC_NUMERIC = "de_DE.UTF-8",
	LC_PAPER = "de_DE.UTF-8",
	LANG = "C"
    are supported and installed on your system.
perl: warning: Falling back to the standard locale ("C").
locale: Cannot set LC_ALL to default locale: No such file or directory
dpkg: error processing archive /var/cache/apt/archives/apache2_2.4.12-1+deb.sury.org~trusty+5_amd64.deb (--unpack):
 error setting ownership of `/var/www/html.dpkg-new': Operation not permitted
Errors were encountered while processing:
 /var/cache/apt/archives/apache2_2.4.12-1+deb.sury.org~trusty+5_amd64.deb
E: Sub-process /usr/bin/dpkg returned an error code (1)

stdout: Reading package lists...
Building dependency tree...
Reading state information...
apache2-mpm-prefork is already the newest version.
apache2-utils is already the newest version.
Suggested packages:
  www-browser apache2-doc apache2-suexec-pristine apache2-suexec-custom
The following NEW packages will be installed:
  apache2
0 upgraded, 1 newly installed, 0 to remove and 102 not upgraded.
7 not fully installed or removed.
Need to get 0 B/146 kB of archives.
After this operation, 461 kB of additional disk space will be used.
(Reading database ... 63059 files and directories currently installed.)
[...]

PR #192, opened 2015-08-08
Issue #186: Do not accept locale environment variables passed over SSH

oxyc:
Not sure where you'd prefer having this?


geerlingguy:
Looks good to me!

PR #193, opened 2015-08-09
added config.yml and drupal.make.yml for ip project

rchadha:


geerlingguy:
Drupal VM is meant to be highly configurable/flexible in many different situations, so I'm not including any particular `config.yml` or `drupal.make.yml` files in the repository. Upstream repositories could add these in, or generate them, if preferred. If there are particular environments that would be better mirrored by Drupal VM with a few small changes in config.yml (e.g. cloud hosting environments), I would be happy to consider merging that in along with some documentation about the particular environment.

Issue #62, opened 2015-03-31
Add extra software task


camoatech:
Add ways to install bower, nodejs, grunt, htop. Maybe a role that calls apt to install these packages and another for npm packages?

geerlingguy:
I could maybe add an optional `extra-packages` variable in `config.yml`, and if anything's configured in that list, it would be installed via `apt`. I tend to shy away from adding extra installations that rely on node, ruby, etc., because usually people want to use them for build tools, and that should usually be done on the host machine (imo). But if there's a valid use case, I'm open to most anything.

geerlingguy:
Others have also requested something like this. I'm thinking I could use `geerlingguy.nodejs` along with a custom variable (if empty, Node is not installed), and `geerlingguy.ruby` along with another custom variable (again, if empty, Ruby is not installed).

This way, people like me who hate having a huge kludge of non-PHP stuff installed on the VM can do so, but I still keep my lean, mean, Drupal machine :)

geerlingguy:
I've added the ability to add extra apt packages with the new `extra_apt_packages` variable.

PR #195, opened 2015-08-12
Add a bit more explanation to the extra_packages variable

sidneygijzen:
Today I ran into the extra_packages variables in my config.yml file. I had a hunch that it could be of use to me, but I had to track it back a bit to `provisioning/tasks/extras.yml` and to [this ticket](https://github.com/geerlingguy/drupal-vm/issues/62) to understand what it exactly does. 

In my case it was perfect to provision the unzip package (for usage in downloading libraries by Drush make files). So, I propose to add a bit more explanation to the variable to make it more clear on how to use it and what it expects.

### DIFF
--- a/example.config.yml
+++ b/example.config.yml
@@ -100,7 +100,9 @@ installed_extras:
   - xdebug
   - xhprof
 
-# Add any extra packages you'd like to install.
+# Add any extra apt or yum packages. Use the apt or yum package name to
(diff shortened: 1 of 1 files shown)


geerlingguy:
I like the idea—sometimes it's best to have a little more explicit direction, especially since the name of the variable is generic and doesn't include 'apt' or 'yum' in it. I'm going to merge this work then try to make the documentation line a little more succinct.

PR #202, opened 2015-08-18
Pass the port number from the Vagrant autogenerated inventory

andreiashu:
The ansible vagrant plugin automatically writes an inventory file with the ssh-config options needed to connect to the vm instance. When one has more than 1 VMs running on the host the chance in this PR will ensure that vagrant provision will still work.

Inventory file sits at .vagrant/provisioners/ansible/inventory/vagrant_ansible_inventory

There's also a "ansible_ssh_host" variable which I suspect would be a better fit for the 'remote-host' value rather than what we have now? This would mean that even if one does not setup their hosts file on the host, drush aliases will still work for that website.


geerlingguy:
This works great, and is a non-disruptive change that helps power users... so thanks!

PR #205, opened 2015-08-19
Fixes #203: Only use Ondrej repo for Ubuntu 12.04 when installing 5.5.

### LINKED ISSUE #203: Weird issue "Unmet dependency with apache2-mpm-prefork" expecting trusty5, but using trusty1
Downloaded drupalvm to a new Macbook. Upon vagrant up, I kept getting an error about a version mismatch in the apache2 version being installed and the one it is told to use. I have only changed the VM memory limit in the config.yml. I also deleted the machine and tried from scratch, this error is hit consistently. Vagrant 1.7.4, VirtualBox 4.30.x

geerlingguy:
See: #203

### DIFF
--- a/provisioning/tasks/init-debian.yml
+++ b/provisioning/tasks/init-debian.yml
@@ -18,7 +18,7 @@
 
 - name: Add repository for PHP 5.5.
   apt_repository: repo='ppa:ondrej/php5'
-  when: php_version == "5.5"
+  when: php_version == "5.5" and ansible_distribution_version == "12.04"
 
 - name: Add repository for PHP 5.6.
   apt_repository: repo='ppa:ondrej/php5-5.6'


kevinquillen:
Does this mean if we change the PHP version to 5.6, it will get around this issue temporarily?

geerlingguy:
@kevinquillen - Possibly. Could you try that and see what happens? I would, but I happen to be in another city with very weak internet, so even reloading/reprovisioning a VM locally would take a couple hours :(

zrpnr:
@kevinquillen changing to 5.6 gives the same error, this patch only fixes #203 when using `php_version: "5.5"`

geerlingguy:
@zrpnr - Yep, I forgot that I changed the default to 5.6. I'll work on a better fix.

Issue #185, opened 2015-08-02
Add ability for public/bridged networking


caschbre:
It would be great if there was a configuration option in config.yml to allow for public/bridged networking. I often need to connect directly to my guest from a device other than the host for a variety of reasons, including testing from a mobile device.  I am also starting to play with setting up my dev VMs on an ubuntu server and a bridged network would make it easier to connect from my laptop/desktop.

geerlingguy:
This is definitely a feature I'll consider adding—networking configuration is not quite flexible with Drupal VM, but there's really no reason for that other than I almost never use non-private networking in my normal work.

See related: #176

caschbre:
I started toying with how to make this a configuration option.  One thing I've run into is when setting up a public_network then NFS synced folder fails.  Here's the error message I get when doing a vagrant up.

"NFS requires a host-only network to be created.
Please add a host-only network to the machine (with either DHCP or a
static IP) for NFS to work."

geerlingguy:
@caschbre can you add an additional network interface? If you just have one interface and switch it from private to public, then it would cause issues with NFS. But if you add another interface for public networking, it should be fine.

caschbre:
@geerlingguy I'm learning this stuff as I go. :)  Can you elaborate on what that would entail? Is that something I do in the Vagrantfile?

geerlingguy:
You would need to add an additional network configuration inside the Vagrantfile, like:

```
config.vm.network "public_network"
```

And put that right before the line that reads `config.ssh.insert_key = false`. Then do a `vagrant up` or `vagrant reload`, and you should have the primary network on your host bridged into the VM.

You can also specify a particular network interface or IP address for one of your host machine's network interfaces (e.g. bridge to WiFi, or a particular network adapter). See https://docs.vagrantup.com/v2/networking/public_network.html

geerlingguy:
I'm thinking the simplest way to make this a configurable option in `config.yml` is to add that line with an if condition for whether a new configuration item like `vagrant_enable_public_network` is true. I should reiterate the warning on the [Vagrant docs public network](https://docs.vagrantup.com/v2/networking/public_network.html) page, though—when you bridge the VM to your computer's network, anyone may be able to access the publicly-available ports, and easily log into the very-insecurely-configured `vagrant` user account and gain root access at least inside the VM.

This box (as is the case with most Vagrant configurations) is configured first for local development convenience, then for security, so keep that in mind :)

caschbre:
[...]

PR #217, opened 2015-08-28
Add ability for public/bridged network #185

caschbre:


example.config.yml
```diff
@@ -10,6 +10,15 @@ vagrant_hostname: drupalvm.dev
 vagrant_machine_name: drupalvm
 vagrant_ip: 192.168.88.88
 
+# If you need to access this guest from a device other than the host you will
+# want to enable public networking. Optionally provide a static IP value.
+# Warning! Vagrant boxes are insecure by default and by design, featuring public
+# passwords, insecure keypairs for SSH access, and potentially allow root access
+# over SSH.
+# https://docs.vagrantup.com/v2/networking/public_network.html
+vagrant_enable_public_network: false
```
geerlingguy: Let's simplify this a bit. Instead of a bool variable + the IP variable, let's do something like:

```
# Allow Drupal VM to be accessed via a network interface on your host by setting
# the IP address here. Vagrant boxes are insecure by default, so be careful
# exposing Drupal VM to the outside world. You've been warned!
vagrant_public_ip: ""
```

Then, see my notes in the logic in the Vagrantfile for the other side of this.

Vagrantfile
```diff
@@ -29,6 +29,15 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   else
     config.vm.network :private_network, ip: vconfig['vagrant_ip']
   end
+
+  if vconfig['vagrant_enable_public_network']
```
geerlingguy: Let's simplify this a bit by just using `vagrant_public_ip`. If it has a value, and the value is 0.0.0.0, do the first condition, else if it has a value, do the second condition:

```ruby
  if !vconfig['vagrant_public_ip'].empty? && vconfig['vagrant_public_ip'] == "0.0.0.0"
    config.vm.network :public_network
  elsif !vconfig['vagrant_public_ip'].empty?
    config.vm.network :public_network, ip: vconfig['vagrant_public_ip']
  end
```

caschbre:
@geerlingguy Good feedback... the changes have been made and pushed up.

rhuffstedtler:
@caschbre - Am I understanding correctly that if you want public networking, but you want it to use DHCP for the address that one should set it to 0.0.0.0?

geerlingguy:
@rhuffstedtler - Yes, that is correct. I'll add an extra line in the config.yml comment directing people to: https://docs.vagrantup.com/v2/networking/public_network.html

PR #220, opened 2015-09-01
Added spaces

opdavies:
Added spaces inside the apache vhost objects(?) so that they are consistent with the rest of the codebase.


geerlingguy:
Looks good, thanks!

PR #222, opened 2015-09-02
added drush extras to the extras task

frob:
This will add the ability to add global drush modules through the config.yml

In the example I am adding registry_rebuild to the vagrant box, but this could also be used to add module builder or any other drush module that isn't necessarily a part of a drupal site.

### DIFF
--- a/example.config.yml
+++ b/example.config.yml
@@ -110,6 +110,8 @@ extra_security_enabled: false
 drush_version: master
 drush_keep_updated: true
 drush_composer_cli_options: "--prefer-dist --no-interaction"
+drush_global_modules:
+  - registry_rebuild
 
 firewall_allowed_tcp_ports:
   - "22"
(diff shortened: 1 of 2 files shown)


geerlingguy:
@frob - If added, I'd rather have this added to the `geerlingguy.drush` role itself (rather than Drupal VM) so it could be used by anyone using the role.

However, besides registry-rebuild, I haven't seen enough people use `drush dl` for things outside a Drupal docroot enough to warrant including the config in normal usage. And registry-rebuild is typically used as an 'quick fix' tool that can be downloaded as needed (imo).

frob:
I am looking at https://github.com/geerlingguy/ansible-role-drush and I have no idea where I would put that. The reason I didn't think it would be there to begin with is, I think of registry rebuild as more of a dev tool to use when I get a project that is completely unorganized.

geerlingguy:
@frob see: https://github.com/geerlingguy/ansible-role-drush/pull/8

PR #223, opened 2015-09-02
Make v.gui configurable within config.yml

opdavies:


frob:
What does this do?

opdavies:
If it's set to true, it opens VirtualBox and displays the VM in a GUI
rather than a headless mode. It's useful for if you need to debug issues
with a VM booting up or starting, which happens occasionally.

-- 

Oliver

geerlingguy:
@opdavies - FYI, since you can open the tool (VirtualBox, VMware, or Parallels) and boot a VM manually (don't need to use `vagrant up`), you can then use GUI mode manually. I'd rather not add this setting, as I want to encourage the use of headless VMs to mirror the experience you'd have with prod (servers are cattle, not pets ;) ).

Because there's a simple manual workaround (boot the VM in the VM tool instead of through Vagrant), I'm going to close this PR.

opdavies:
@geerlingguy That's fair. I hadn't thought of manually booting the VM from VirtualBox before, and always added or updated the value in Vagrantfile. I'll do it that way in the future. :)

PR #234, opened 2015-09-07
Added file provisioner to Vagrantfile to copy know_hosts from host to…

ravbaba:
It copies know_hosts from host to VM so you are avoiding a message which you need to confirm when you are trying to connect first time via ssh with a particular host.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -61,6 +61,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     config.vm.provision "ansible" do |ansible|
       ansible.playbook = "#{dir}/provisioning/playbook.yml"
       ansible.sudo = true
+      ansible.extra_vars = {
+        known_hosts_path: vconfig['known_hosts_path']
+      }
     end
   # Provision using shell provisioner and JJG-Ansible-Windows otherwise.
   else
--- a/provisioning/tasks/sshd.yml
+++ b/provisioning/tasks/sshd.yml
@@ -5,3 +5,18 @@
     regexp: "^AcceptEnv LANG"
     state: absent
     validate: "/usr/sbin/sshd -T -f %s"
+
+# Copy over known_hosts.
+- name: Check if known_hosts file is present.
+  stat: "path={{ known_hosts_path }}"
+  register: known_hosts
+  connection: local
+  sudo: no
+
+- name: Copy known_hosts file from host into Drupal VM.
+  copy:
+    src: "{{ known_hosts_path }}"
+    dest: ~/.ssh/known_hosts
+    mode: 0644
+  sudo: no
+  when: known_hosts.stat.exists
--- a/example.config.yml
+++ b/example.config.yml
@@ -22,6 +22,10 @@ vagrant_synced_folders:
     type: nfs
     create: true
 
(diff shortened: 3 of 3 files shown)


oxyc:
Cool :+1:

Vagrantfile
```diff
@@ -63,6 +63,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       mount_options: synced_folder.include?('mount_options') ? synced_folder['mount_options'] : []
   end
 
+  # Provision using file provisioner to add known_hosts from host to guest.
+  config.vm.provision "file", source: "~/.ssh/known_hosts", destination: "~/.ssh/known_hosts"
```
geerlingguy: Will this work correctly on Windows? I haven't used the file provisioner with a file that may or may not exist, so haven't seen the behavior if the file doesn't exist.

ravbaba: @geerlingguy you are right, I haven't test it on Windows and I can guess it won't work on it. I can move it to an ansible task just tell me where you prefer to place it then I can check if the file exists or not.

geerlingguy:
@ravbaba - You could probably do something like the following, maybe inside `sshd.yml`:

```yaml
    # Copy over known_hosts.
    - name: Check if known_hosts file is present.
      stat: path=~/.ssh/known_hosts
      register: known_hosts
      connection: local
      sudo: no

    - name: Copy known_hosts file from host into Drupal VM.
      copy:
        src: ~/.ssh/known_hosts
        dest: ~/.ssh/known_hosts
        mode: 0644
      sudo: no
      when: known_hosts.stat.exists
```

Please see if that works. If so, that's probably a little cleaner, and should work fine on Windows hosts.

ravbaba:
@geerlingguy - I am going to check it on ios and ubuntu if I find some time today and on Windows but during the weekend. Thank you for the example ;)

ravbaba:
@geerlingguy - I've just tested it on my mac and it works for me, I added some modification to your code so please let me know if you are ok with it. I added your code to my PR.

ravbaba:
I tested it on ubuntu and it works as well, didn't manage to test it on windows.

Vagrantfile
```diff
@@ -61,6 +61,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     config.vm.provision "ansible" do |ansible|
       ansible.playbook = "#{dir}/provisioning/playbook.yml"
       ansible.sudo = true
+      ansible.extra_vars = {
```
geerlingguy: This configuration is not needed, as Ansible picks up any configuration from config.yml automatically.

provisioning/tasks/sshd.yml
```diff
@@ -5,3 +5,18 @@
     regexp: "^AcceptEnv LANG"
     state: absent
     validate: "/usr/sbin/sshd -T -f %s"
+
+# Copy over known_hosts.
```
geerlingguy: This comment can be left out; it's obvious enough from the two tasks what's going on :)

ravbaba:
@geerlingguy - sorry, I've just noticed your comments. Thank you for merging and cleaning up the PR.

PR #240, opened 2015-09-14
Add variable to use ansible in vm as a preference.

joestewart:
Adding a variable to use ansible installation in the vm even if ansible exists on the host.

It allows to set a variable in the config file to choose to run ansible in the vm.

```
# ansible installed in vm preferred
ansbile_in_vm_preferred: true
```

This should change nothing for existing installations and config files.  Although the example config file could have the variable defined as false.

```
# ansible installed in vm preferred
ansbile_in_vm_preferred: false
```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -64,7 +64,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   end
 
   # Provision using Ansible provisioner if Ansible is installed on host.
-  if which('ansible-playbook')
+  if which('ansible-playbook') && !vconfig['ansbile_in_vm_preferred']
     config.vm.provision "ansible" do |ansible|
       ansible.playbook = "#{dir}/provisioning/playbook.yml"
       ansible.sudo = true


geerlingguy:
@joestewart - What would be the use case for this? IMO, using Ansible inside the VM is a crutch/very non-ideal, and the only reason I added it was so Windows users could still use Drupal VM. Unlike Chef/Puppet/CF/etc., Ansible's major strength is that you _don't_ need to kludge up your managed servers/VMs with anything extra for management.

I'd rather avoid adding another variable and extra logic to the Vagrantfile unless there's a very compelling reason to avoid using the already-installed copy of Ansible.

joestewart:
I understand.  I didn't suggest it when the automatic check went in to see if I had much of a use case.

The first use case is selfish.  To be able to test installation like the Windows users but on my machine without changing anything else.

The other was a support issue.  I had a user with ansible problems on the host.  There was an old version of ansible and a couple of related problems.  This variable allows to use the vagrant box immediately and fix the ansible problem as time allowed.

One benefit of allowing install in the vm is the ease of getting new users up and running and no other host requirements.

geerlingguy:
For the first use case, I often do the same, but I just go to the Vagrantfile and switch around the if condition so it runs the provisioner inside the VM regardless.

For the other use case, I'd rather still help the user to get onto the latest version of Ansible on the host, and fix problems there, because ideally, Drupal VM is a starting point for using Ansible to automate other infrastructure or local environments, not just Drupal VM. If someone wants to only use it for Drupal VM, then it might be better to uninstall Ansible entirely (or not install it in the first place).

The Quick Start Guide doesn't mention Ansible installation at all, except for faster setup/better overall experience. It's not a requirement :)

PR #242, opened 2015-09-14
Display DrupalVM logo using Markdown.

joestewart:
Simple change to display the DrupalVM using markdown.

Seemingly not needed, but the logo will then display OK on Bitbucket.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-<p align="center"><img src="https://raw.githubusercontent.com/geerlingguy/drupal-vm/master/docs/images/drupal-vm-logo.png" alt="Drupal VM Logo" /></p>
+![Drupal VM Logo](https://raw.githubusercontent.com/geerlingguy/drupal-vm/master/docs/images/drupal-vm-logo.png)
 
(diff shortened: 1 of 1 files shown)


geerlingguy:
I'd like to keep the logo centered, though—for BitBucket, does the logo not display at all when the markdown includes plain HTML markup?

Markdown's original standard dictates plain HTML tags should still be allowed as a fallback. If BitBucket isn't following the standard, then that should be fixed on their end :P

geerlingguy:
@joestewart - FYI you were the 300th person to star the repo. If there were some award I could give you, I'd send it your way :D

joestewart:
I guess close this one too :)

Markdown doesn't center as far as I know.  And Bitbucket is using some kind of safe mode to disallow using the "p" and "img" html.

300, building a good community.  Thanks again.

joestewart:
Correct, on Bitbucket it does not display at all and instead shows the html.

```<p align="center">Drupal VM Logo</p>```

geerlingguy:
I'd rather have forks on BitBucket look nice too... the centering is more an OCD passion than anything else for me, so since the logo still displays, at least, and doesn't look bad, I'm fine merging this.

Issue #174, opened 2015-07-14
Add ability to provide custom shell scripts to run after provisioning


geerlingguy:
It would be great if a user could add a custom Ansible tasks file (or maybe even extra roles, and we add the `roles` directory to `.gitignore`), that would be run if present, ignored if not present.

Maybe call it `extra_tasks.yml` and have it included from the project root directory (if present). This isn't supported OOTB by Ansible, but this might be a viable approach (default to an empty file if the extra tasks file isn't present): http://stackoverflow.com/a/29583251/100134

geerlingguy:
This will need a documentation page, since usage will not be 100% intuitive/discoverable.

geerlingguy:
I did a good deal of testing for this on Ansible 2.0.0 (still not the stable release yet), and it looks like the following code should do the trick, since 2.x includes files during playbook runs, and doesn't precompile everything:

```yaml
    # Include extra tasks file if it's present.
    - name: Check if extra_tasks.yml is present.
      stat: path=extra_tasks.yml
      register: extra_tasks_file
      connection: local

    - include: extra_tasks.yml
      when: extra_tasks_file.stat.exists
```

geerlingguy:
Postponed, because I can't see a way to do this until Ansible 2.x is publicly available. This will make a nice blog post though!

geerlingguy:
Hmm... another thought I just had—I can include a user-provided shell script in the same way, but since it's not an Ansible `include`, it will work with any old version of Ansible, just via `when`... So I think I'll do that for now, then maybe also add the custom Ansible tasks list as well once 2.0 is stable/out for a while.

oxyc:
Once you upgrade to 2.0, it would be nice if we removed the `#{dir}` prefix from the `config.yml` as well, and pass `Dir.pwd` on to ansible for including the file.

That would make it possible to use this repo as a git submodule or otherwise an unmodified tree together with 7e87005c5

geerlingguy:
Getting extremely close to the finish line here; instead of an ansible playbook, I'm just allowing users to add a list of shell scripts to run after provisioning is complete. I'm already going to use this for a few of my instances where I need to copy Solr search core configuration in place and restart solr as part of setup!

geerlingguy:
Woohoo! See the [docs](http://docs.drupalvm.com/en/latest/extras/scripts/) for information on how to provide scripts to run post-provision. This is working great, and allows zero, one, or as many shell scripts you want!

[...]

PR #244, opened 2015-09-16
Copy Drupal Apache Solr Search conf files into the correct directory.

joestewart:
This should place the files from the Drupal Apache Solr module into the conf directory.

### DIFF
--- a/provisioning/tasks/extras.yml
+++ b/provisioning/tasks/extras.yml
@@ -8,3 +8,30 @@
   yum: "name={{ item }} state=installed"
   with_items: extra_packages | list
   when: ansible_os_family == 'RedHat' and extra_packages | length
+
+- name: Set solr_conf_directory for apachesolr version 3.x.
+  set_fact:
+    solr_conf_directory: "{{ solr_home }}/conf/"
+  when: "'solr' in installed_extras and solr_version.split('.')[0] == '3'"
+
+- name: Set solr_conf_directory for apachesolr version 4.x.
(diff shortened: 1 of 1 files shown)


geerlingguy:
@joestewart - I really, really like this idea, but I don't want to necessarily add it to Drupal VM at this point; a few problems are that this currently only works with the `apachesolr` module, though a significant number of users would like to use `search_api_solr`, and when Drupal 8's Search API Solr module comes around, we'd need to add additional logic to use the D8 copies of the Solr config.

At this time, I think having the process documented separately (docs.drupalvm.com) is good enough, and soon, once #174 is complete, users could add scripts like `post-provision.sh` or `post-provision.yml` that does extra project-specific setup like configuring Solr after the main provisioning is complete.

Thanks!

mchelen:
How about providing this as an option in `config.yml` that is disabled by default? I'm running into the same issue except with `search_api_solr`, and there may be enough users who need some type of Solr integration to make it worthwhile.

joestewart:
I think this use case is the example post provision script - https://github.com/geerlingguy/drupal-vm/blob/master/examples/scripts/configure-solr.sh

mchelen:
@joestewart Ah yeah I see, and it's listed commented out in the `config.yml`. Thanks!

geerlingguy:
Yep! Just added that a week or two ago :)

mchelen:
@geerlingguy Awesome, very handy!

PR #248, opened 2015-09-20
Updated requirements to use YAML.

thom8:
[DEPRECATION WARNING]: Non yaml files for role requirements. This feature will be removed in a future release. Deprecation warnings can be disabled by setting
deprecation_warnings=False in ansible.cfg.


geerlingguy:
@thom8 - I'm assuming you're using Ansible 2.x or the latest `devel` version from the Git repo, correct? This option _is_ deprecated in 2.0, but not in the current stable release. Nevertheless, future proofing FTW!

geerlingguy:
Also, it looks like you'll need to fix the following line inside `.travis.yml`:

```yaml
  - 'ansible-galaxy install -r provisioning/requirements.txt'
```

geerlingguy:
This also needs to be updated to include the `geerlingguy.nginx` role. Sorry I didn't get to this earlier!

thom8:
Yes this warning is only displaying for ansible 2.0.0 and above.

Updated ```.travis.yml``` and added the ```nginx``` role.

PR #254, opened 2015-09-23
Support Nginx access logs.

joestewart:


geerlingguy:
If no access log is defined per server block, it should log using the default config for Nginx, which I think is set to /var/log/nginx/access.log—is that not adequate for most use cases?

joestewart:
I'll try again but wasn't seeing anything written.

geerlingguy:
@joestewart - Note that, by default, Nginx doesn't write logs to disk until a buffer is full; you can simulate a buffer flush by restarting nginx (`sudo service nginx restart`); see if the log is populated after you restart Nginx. If so, buffering is turned on.

geerlingguy:
(See: http://nginx.org/en/docs/http/ngx_http_log_module.html)

joestewart:
Yes, it was the buffering.  I hadn't noticed it until I had added the template change.  

The released version is working well.

BTW, the Nginx and nodejs support in 2.1.0 is great.

geerlingguy:
@joestewart - I do what I can :)

I'll be hopefully getting time to write a few blog posts on Nginx + Drupal in the next month, too; it'll be on Drupal planet!

PR #269, opened 2015-10-14
Add task and variable to disable pre-installed Drupal modules.

afbrogneaux:
With this functionnality, we can disable pre-installed Drupal modules that are no longer useful because of other modules enabled with the variable `drupal_enable_modules`. For instance, we have to disable `toolbar` when using `admin_menu`, otherwise two toolbars are visible.

Example of configuration (4 modules to disable):
```
drupal_disable_modules: [ 'color help overlay toolbar' ]
```

### DIFF
--- a/example.config.yml
+++ b/example.config.yml
@@ -50,6 +50,7 @@ drupal_domain: "drupalvm.dev"
 drupal_site_name: "Drupal"
 drupal_install_profile: standard
 drupal_enable_modules: [ 'devel' ]
+drupal_disable_modules: []
 drupal_account_name: admin
 drupal_account_pass: admin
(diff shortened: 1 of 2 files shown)


geerlingguy:
In this particular case, I would argue it's better to write your own install profile if you want to have a particular configuration with certain modules disabled. The 'install modules' part is just so that if you're doing a bunch of debugging of a few particular modules, you can get them enabled quickly. For a site setup that requires a particular set of modules and configurations, an install profile would be a better route to take.

opdavies:
I would suggest looking at the [Master module](https://www.drupal.org/project/master) for defining which modules should be installed or uninstalled.

PR #277, opened 2015-10-19
Add excluded_paths

opdavies:
Add an empty excluded_paths array for vagrant_synced_folders, and added some more comments with suggestions about what to exclude.


geerlingguy:
I think this may be better suited as a mention in documentation, because most people use NFS and don't need to set excluded_paths. Can you open a separate PR adding the same notes and an example inside https://github.com/geerlingguy/drupal-vm/blob/master/docs/extras/syncing-folders.md ?

opdavies:
Sure. :smile:

PR #279, opened 2015-10-19
Updates to examples and syntax highlighting.

scottreese:
* Updated behat.yml to include the needed contexts to load and use the DrupalExtension driver.
* Updated behat.yml to include region mapping for content region so the DrupalExtension could find it.
* Updated examples to point to features/web since that is what behat.yml is set up for.
* Added syntax to code fences to help fix formatting on the web page.

### DIFF
--- a/docs/extras/behat.md
+++ b/docs/extras/behat.md
@@ -34,12 +34,17 @@ Using the default Drupal site as an example (it's installed in `/var/www/drupal`
 
   1. Create a `behat.yml` file inside the docroot of your site (e.g. create this file alongside the rest of the Drupal codebase at `/var/www/drupal/behat.yml`), with the following contents:
 
-    ```
+    ```yaml
     default:
       suites:
         web_features:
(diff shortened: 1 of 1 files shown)


geerlingguy:
Looks great to me—much cleaner, thanks!

scottreese:
Still doesn't appear to be rendering properly on this page: http://docs.drupalvm.com/en/latest/extras/behat/ even though it looks fine here on github. :/

geerlingguy:
I remember running into that issue on readthedocs before—can you open up a new issue in the issue queue mentioning this problem (and linking to that page) and I'll commit a few different tests to see how to get it working?

The problem is it's a code block in a list, and every markdown parser treats those differently. I think there might be another example _somewhere_ in the docs.

scottreese:
Ah, yes, I was just coming to that conclusion when you commented. Issue created here: https://github.com/geerlingguy/drupal-vm/issues/280.

PR #281, opened 2015-10-20
Add documentation and example of rsync excluded_paths

opdavies:
Here's a new PR, following on from #277.


opdavies:
Thanks. :smile:

PR #292, opened 2015-11-01
Ruby coding style fixes via rubocop in Vagrantfile

nemeth:
I ran `rubocop --auto-correct` on Vagrantfile to be more ruby coding style compliant and made some minor style fixes. The `.rubocop.yml` is there to skip two remaining offenses (long lines and [indentation](https://github.com/geerlingguy/drupal-vm/blob/master/Vagrantfile#L65)).

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,36 +1,38 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
-VAGRANTFILE_API_VERSION = "2"
+VAGRANTFILE_API_VERSION = '2'
+
+require 'yaml'
 
 # Cross-platform way of finding an executable in the $PATH.
 def which(cmd)
   exts = ENV['PATHEXT'] ? ENV['PATHEXT'].split(';') : ['']
   ENV['PATH'].split(File::PATH_SEPARATOR).each do |path|
-    exts.each { |ext|
+    exts.each do |ext|
       exe = File.join(path, "#{cmd}#{ext}")
       return exe if File.executable?(exe) && !File.directory?(exe)
-    }
+    end
   end
-  return nil
+  nil
 end
 
 # Use config.yml for basic VM configuration.
-require 'yaml'
 dir = File.dirname(File.expand_path(__FILE__))
-if !File.exist?("#{dir}/config.yml")
+begin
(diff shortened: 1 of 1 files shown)


.rubocop.yml
```diff
@@ -0,0 +1,7 @@
+# Allow long lines.
```
geerlingguy: Since this project is not primarily a Ruby project, I'd rather avoid adding Ruby-specific tooling/config for it; can you take this file out of the PR?

Vagrantfile
```diff
@@ -1,36 +1,38 @@
...
   ENV['PATH'].split(File::PATH_SEPARATOR).each do |path|
-    exts.each { |ext|
+    exts.each do |ext|
       exe = File.join(path, "#{cmd}#{ext}")
       return exe if File.executable?(exe) && !File.directory?(exe)
-    }
+    end
   end
-  return nil
+  nil
```
geerlingguy: I guess this is more proper Ruby syntax, but it's so grating to see no `return` when that's what the function's doing... is this kosher?

Vagrantfile
```diff
@@ -113,6 +115,5 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   end
 
   # Set the name of the VM. See: http://stackoverflow.com/a/17864388/100134
-  config.vm.define vconfig['vagrant_machine_name'] do |d|
-  end
+  config.vm.define vconfig['vagrant_machine_name']
```
geerlingguy: I guess I'm asking the obvious, but will these two syntaxes result in the exact same behavior? (e.g. the machine name will still be set appropriately)

Vagrantfile
nemeth: yeah, it is recommended to "[avoid return](Avoid return where not required for flow of control) where not required for flow of control"

.rubocop.yml
nemeth: I removed it.

geerlingguy:
Sorry for letting this linger so long... can you run it again and post the latest fixes, and I'll test locally and merge if it's all good?

geerlingguy:
Updated and fixed in #472 - thanks @oxyc and @nemeth, much nicer Ruby syntax!

Issue #189, opened 2015-08-04
rsync__args and guest generated files


sidneygijzen:
hello,

As I work a lot with Zen subthemes, I use a.o. Bundler, SASS and Compass to generate my CSS files on the guest VM. At some point I sync back the files to the host, for which I use the [vagrant rsync-back plugin](https://github.com/smerrill/vagrant-rsync-back). However I noticed that all files were 666 and all folders 777 after rsync'ing (back). This is caused by the last argument in ```rsync__args``` parameter: ```"--chmod=ugo=rwX" ```. Once I removed that argument all permissions are kept intact, and thus I don't have to reset all permissions after rsyncing back from guest to host.

Up til now I haven't run into any troubles (e.g. I can run ```bundle exec compass compile``` without sudo'ing). However, I do use [some extra settings](http://docs.drupalvm.com/en/latest/extras/syncing-folders/#permissions-related-errors) as mentioned in the Drupal VM docs. My synced_folder config (Vagrantfile) is as follows:
```ruby
for synced_folder in vconfig['vagrant_synced_folders'];
    config.vm.synced_folder synced_folder['local_path'], synced_folder['destination'],
      type: synced_folder['type'],
      rsync__auto: "true",
      rsync__exclude: synced_folder['excluded_paths'],
      rsync__args: ["--verbose", "--archive", "--delete", "-z"],
      id: synced_folder['id'],
      create: synced_folder.include?('create') ? synced_folder['create'] : false,
      owner: "vagrant",
      group: "www-data",
      mount_options: synced_folder.include?('mount_options') ? synced_folder['mount_options'] : ["dmode=775,fmode=664"]
  end
```
I haven't tried all possible file generating actions on the guest VM yet, but I figured I would report it to see if more users have run into this. I'm also of course curious about how other users are solving this! :)

oxyc:
I've been using the following for a while and seems to work as well.

- I've also removed rsync compression (should be faster when it's local network)
- Set it to use `--numeric-ids` but I doubt it has any difference in performance even if I read someone mention it.
- Further I removed vagrants recursive chown (as it chowned the files folder as well).

I agree that this should either be configurable with `synced_folder['rsync_args']` or set to a saner default.

```rb
  for synced_folder in vconfig['vagrant_synced_folders'];
    config.vm.synced_folder synced_folder['local_path'], synced_folder['destination'],
      type: synced_folder['type'],
      rsync__auto: "true",
[...]

PR #293, opened 2015-11-01
add support for overriding any synced_folder option

oxyc:
I mentioned this earlier in https://github.com/geerlingguy/drupal-vm/issues/189#issuecomment-139809306.

This would allow the user to override any option passed to `config.vm.synced_folders`. By default it would still work exactly as is does now. Currently I'm using these options for most of my projects for better file permissions using rsync (the merge is not deep by choice, so that you can eg remove specific `rsync__args`).

```rb
options_override:
  rsync__chown: false
  rsync__verbose: true
  rsync__args: [
    "--verbose", "--archive", "--delete",
    "--chmod=Dgu=rwx,Do=rx,Fgu=rw,Fo=r", # 664 for files, 775 for directories
    "--owner", "--group", # required for the following command
    "--usermap=*:vagrant", "--groupmap=*:www-data,*:apache",
  ]
```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -66,7 +66,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
   # Synced folders.
   for synced_folder in vconfig['vagrant_synced_folders'];
-    config.vm.synced_folder synced_folder['local_path'], synced_folder['destination'],
+    options = {
       type: synced_folder['type'],
       rsync__auto: "true",
       rsync__exclude: synced_folder['excluded_paths'],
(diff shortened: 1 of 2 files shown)


geerlingguy:
I'd be happy to merge this pull request, but could you add your example above (with a little verbiage explaining why it's helpful to be able to override, etc.) to the documentation inside `docs/extras/syncing-folders.md`?I'd be happy to merge this pull request, but could you add your example above (with a little verbiage explaining why it's helpful to be able to override, etc.) to the documentation inside `docs/extras/syncing-folders.md`?

oxyc:
@geerlingguy I added some docs

geerlingguy:
Looks good to me, thank you so much!

PR #331, opened 2015-12-04
Remove PHP 7 RC comments from docs

Chapabu:
Removed "PHP 7 is out soon" comments from PHP 7 documentation page .

PHP 7.0.0 is out now! HURRAH!


geerlingguy:
Looks good! It's still slightly experimental until I get time to make sure all my PHP-based roles work well with PHP7. Currently some extensions that power core functionality don't work with PHP7 :(

PR #343, opened 2015-12-11
Adding Amazon Web Services Server support

Tigatok:
We assume that the user has an Ansible version of at least 1.9.4 (May work with an older version, only used the newest one.).

We also assume that the user has an AWS account already, has at least one created user, created a Key Pair (and has the PEM key associated), and has a security group set up.

### DIFF
--- a/example.config.yml
+++ b/example.config.yml
@@ -34,6 +34,7 @@ aws_key_id: ""
 aws_secret_access_key: ""
 aws_key_pair_name: "KeyPairName_CaSeSeNsItIvE"
 aws_security_group: "Test-security-group-name"
+aws_ami: 'ami-7747d01e'
 aws_tags_name: "Test AWS Vagrant"
 override_ssh_username: "ubuntu"
 override_ssh_private_key_path: 'PATH_TO_YOUR_PEM_FILE'
--- a/example.config.yml
+++ b/example.config.yml
@@ -15,6 +15,29 @@ vagrant_ip: 192.168.88.88
 # See: https://docs.vagrantup.com/v2/networking/public_network.html
 vagrant_public_ip: ""
 
+# If you want to use a Amazon Web Services Server
+# This will remain off unless aws_key_id != ''
+# If you turn this setting on, you must make a 
+# .pem file that points to the path you specify
+# containing the private key of the Amazon Instance.
+# aws_key_id: Retrieved from the users section of AWS
+# aws_secret_access_key: Also retrieved from users sections of AWS
+# aws_key_pair_name: The name of the KeyPair you generated on AWS
+#                    MUST BE EXACT spelling (Caps, Etc).
+# aws_security_group: The group name you want the security to use.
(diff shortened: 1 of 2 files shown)


example.config.yml
```diff
@@ -15,6 +15,30 @@ vagrant_ip: 192.168.88.88
...
+#                    MUST BE EXACT spelling (Caps, Etc).
+# aws_security_group: The group name you want the security to use.
+# aws_tags_name: What you want the AWS tag name to show up as.
+# override_ssh_username: The SSH username. Should be ubuntun 90% of the time.
+# override_ssh_private_key_path: The path to the private key for the keypair
+#                                you created.
+# The ssh username is default to ubuntu
+aws_key_id: ""
+aws_secret_access_key: ""
+aws_key_pair_name: "KeyPairName_CaSeSeNsItIvE"
```
chx: I am not sure whether Ansible would let you do something better but having the secret key in a git tracked file leads to key leak.

Tigatok: Hmm, good point.
I guess because the user renames the file from example.config.yml to config.yml we could add that to the gitignore?

chx: I will let you and @geerlingguy figure out something good , I just pointed out a potential problem :)

geerlingguy: @chx - Some would use something like Ansible Vault if it's a secret that _does_ need to be tracked. However, in this case, it's likely best to do something like pass in environment variables [like Boto uses](https://github.com/boto/boto#getting-started-with-boto) so users can just set them as they would other secret vars.

oxyc:
Due to the amount of providers out there, maybe it's worth considering "sourcing" a local Vagrantfile so users can profit from Drupal-VMs default but still add their on flavours on top?

Eg.
```rb
if File.exist?('Vagrantfile.local')
  eval File.read 'Vagrantfile.local'
end
```

And then your `Vagrantfile.local` could look like:
```rb
config.vm.provider :aws do |aws, override|
  override.nfs.functional = false
  aws.access_key_id = vconfig['aws_key_id']
  aws.secret_access_key = vconfig['aws_secret_access_key']
  aws.keypair_name = vconfig['aws_key_pair_name']
  aws.security_groups = vconfig['aws_security_group']
  aws.ami = vconfig['aws_ami']
  aws.tags['Name'] = vconfig['aws_tags_name']
  override.ssh.username = vconfig['override_ssh_username']
  override.ssh.private_key_path = vconfig['override_ssh_private_key_path']
end
```

geerlingguy:
@oxyc - If that works, I'd be game for that... would also allow other Vagrantfile settings that I'm currently not including just due to the sheer monstrosity of the current state of the Vagrantfile.

oxyc:
The snippet I pasted works when I tested it briefly. I can test it a bit more and send over a PR with docs later (maybe in a few days).

geerlingguy:
Closing in favor of the approach in #401

PR #363, opened 2016-01-02
Update xdebug.md for NetBeans Configuration

ccjjmartin:
Thanks for making Drupal VM!  It is awesome and I was hoping to contribute back.


geerlingguy:
@ccjjmartin - Thanks! The default port is 9000 according to the php-xdebug role, so that doesn't need to be explicitly set in config.yml (I slightly modified the directions to reflect this).

PR #401, opened 2016-01-29
Add support for sourcing a local Vagrantfile

oxyc:
See https://github.com/geerlingguy/drupal-vm/pull/343

@Tigatok I credited you on the docs as this was your research and instructions. I simplified it a bit as users can add whatever configurations they want now, not just what Drupal VM choses to expose. Does it look okay to you?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 vagrant_ansible_inventory_default
 config.yml
 drupal.make.yml
+Vagrantfile.local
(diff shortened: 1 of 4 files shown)


oxyc:
Also used environment variables in the example as @geerlingguy suggested.

geerlingguy:
This looks great, will merge asap after a couple more tests.

Issue #320, opened 2015-11-19
Request: Info/home page like in VDD


robdecker:
VDD (https://www.drupal.org/project/vdd) has a really useful info/home page that gets displayed if you go to the VM's IP address. I like this idea much better than just displaying the default site.

See https://vimeo.com/95936124 at about 3:01 for a demo of VDD's info page.

geerlingguy:
I like this idea. Could be a pleasant landing page with all inline styling that just directs people to the documentation site and maybe also links to installed sections of the VM, e.g. Adminer, PimpMyLog, etc.

geerlingguy:
Related: #143

robdecker:
Additional ideas:
- a list of sites (with links) installed in the VM, along with their local_path on the host OS
- a generated list of entries to add to the hosts file
- db connection info for external tools like Sequel Pro

Derimagia:
I used Dashbrew for a while and it had a pretty decent dashboard. Take a look here (There are screenshots):

https://github.com/mdkholy/dashbrew

oxyc:
I started up a PR for this. Feel free to comment or make suggestions about #404

oxyc:
It's merged this task can now be closed :) There's a follow up here https://github.com/geerlingguy/drupal-vm/issues/451

PR #404, opened 2016-01-29
[WIP] Issue #320: add a dashboard

oxyc:
First take at a dashboard mentioned in #320. I'm not sure how we would want to add this to the webserver actually? At the moment it doesn't work :smile: It works if you set `apache_remove_default_vhost: false` hehe

Also I've only tested minimally (only apache for example).

I used bootstrap so this wouldn't work if the user is offline. I feel that's still okay (the stylesheet is loaded at the end of the document not to pause the rendering), maybe add some minimal styling for such cases.

One minor hack is the `Adminer` link for each database. This links to `adminer.drupalvm.dev/?username=root&db=drupal`, which is actually an auto-login feature, but it fails due to not being configured. Instead it redirects the user back to login screen with an error message _AND_ also pre-fills the fields.

screen shot 2016-01-30 at 12 32 29

### DIFF
--- a/provisioning/tasks/dashboard.yml
+++ b/provisioning/tasks/dashboard.yml
@@ -1,0 +1,11 @@
+- name: Ensure the dashboard directory exists
+  file:
+    path: "{{ dashboard_install_dir }}"
+    state: directory
+    mode: 0755
+
+- name: Copy dashboard page into place.
+  template:
+    src: ../templates/dashboard.html.j2
+    dest: "{{ dashboard_install_dir }}/index.html"
+    mode: 0644
--- a/provisioning/templates/dashboard.html.j2
+++ b/provisioning/templates/dashboard.html.j2
@@ -1,0 +1,317 @@
+<!DOCTYPE html>
+  <head>
+    <title>Drupal VM</title>
+    <meta charset="utf-8">
+    <style>
+      .section-host {
+        font-family: monospace;
+        list-style-type: none;
+        padding: 0;
+        margin: 0;
+      }
+      /* minor fallback styling for when loaded without an internet connection */
+      .panel, .jumbotron, .well {
+        margin-bottom: 20px;
+      }
+      td, th {
+        padding: 8px;
+      }
+    </style>
+  </head>
+  <body>
+    {%
+      set devtool_docroots = [
+        adminer_install_dir if 'adminer' in installed_extras else '',
+        php_xhprof_html_dir if 'xhprof' in installed_extras else '',
+        pimpmylog_install_dir if 'pimpmylog' in installed_extras else '',
+        dashboard_install_dir
+      ]
+    %}
+
+    {# Returns the hosts server name based on the document root #}
+    {%- macro getServernameFromDocroot(path) -%}
+      {%- if drupalvm_webserver == 'apache' -%}
+        {%- for host in apache_vhosts -%}
+          {%- if host.documentroot == path -%}
+            {{ host.servername }}
+          {%- endif -%}
+        {%- endfor -%}
+      {%- elif drupalvm_webserver == 'nginx' -%}
+        {%- for host in nginx_hosts -%}
+          {%- if host.root == path -%}
+            {{ host.server_name }}
+          {%- endif -%}
+        {%- endfor -%}
+      {%- endif -%}
+    {%- endmacro -%}
+
+    {%- macro printSite(servername, docroot) -%}
+      {%- if docroot not in devtool_docroots -%}
+        <tr>
+          <td><a href="http://{{ servername }}">{{ servername }}</a></td>
+          <td><code>{{ docroot }}</code></td>
+        {% if configure_local_drush_aliases %}
+          <td><code>drush @{{ vagrant_machine_name }}.{{ servername }}</code></td>
+        {% endif %}
+        </tr>
+      {% endif %}
+    {%- endmacro %}
+
+    {%- macro sectionHost() -%}
+      <ul class="section-host">
+        {% if drupalvm_webserver == 'apache' -%}
+          {%- for host in apache_vhosts -%}
+            {%- if host.documentroot != dashboard_install_dir -%}
+              <li>{{ host.servername }} {{ vagrant_ip }}</li>
+            {%- endif -%}
+            {% if host.serveralias is defined -%}
+              {%- for alias in host.serveralias.split() -%}
+                <li>{{ alias }} {{ vagrant_ip }}</li>
+              {%- endfor -%}
+            {%- endif %}
+          {%- endfor -%}
+        {%- elif drupalvm_webserver == 'nginx' -%}
+          {%- for host in nginx_hosts -%}
+            {%- if host.root != dashboard_install_dir -%}
+              <li>{{ host.server_name }} {{ vagrant_ip }}</li>
+            {%- endif -%}
+          {%- endfor -%}
+        {%- endif %}
+      </ul>
+    {%- endmacro -%}
+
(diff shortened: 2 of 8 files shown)


provisioning/templates/dashboard.html.j2
```diff
@@ -0,0 +1,294 @@
...
+    {%- endmacro -%}
+
+    <div class="container">
+
+      <div class="row">
+        <div class="col-md-12">
+          <section class="jumbotron">
+            <div class="row">
+              <div class="col-md-7">
+                <h1>Welcome!</h1>
```
oxyc: Should probably be replaced with the Drupal VM logo instead

geerlingguy:
I think @oxyc is trying to win contributor of the millenium award. I think I owe you something like six beers now. This is already awesome, and looks exactly like what I would like.

Since it's using the default path (at least for Debian, have to verify for RedHat) for the default server, the IP address of the domain should just open up to this, correct? We might need to set the apache/nginx roles to not delete the default vhost file though.

oxyc:
That's that I thought, that it would only take over the IP, but then it took over all the hostnames too. Not that familiar with configuring apache though.

Haha, I'm running low on tasks at work so I'm trying to find other things to spend the day on.

provisioning/templates/dashboard.html.j2
```diff
@@ -0,0 +1,294 @@
...
+                <p><a href="https://github.com/geerlingguy/drupal-vm">Drupal VM</a> is A VM for local Drupal development, built with Vagrant + Ansible.</p>
```
oxyc: Typo: `A VM`. Appears in https://github.com/geerlingguy/drupal-vm/blob/master/docs/index.md as well.

oxyc:
I got it working as expected if I changed the filenaming so that `000-default.conf` loads last, and then added a `ServerAlias *` to it.

But I feel like it's supposed to work out of the box. Is there some hostname variable somewhere that says the IP maps to the hostname locally?

oxyc:
So some more research for apache under ubuntu.

> If you omit the ServerName directive from any name-based virtual host, the server will default to a fully qualified domain name (FQDN) derived from the system hostname. This implicitly set server name can lead to counter-intuitive virtual host matching and is discouraged.
https://httpd.apache.org/docs/trunk/vhosts/name-based.html

So the default vhost resolves to the hostname due to the following in /etc/hosts:
```
127.0.1.1 drupalvmv.dev drupalvmv
```

Guess we should really be creating a new vhost set to the ip.

oxyc:
So this works for both nginx and apache on Ubuntu at least.

```diff
diff --git a/example.config.yml b/example.config.yml
index 3072d80..c402358 100644
--- a/example.config.yml
+++ b/example.config.yml
@@ -97,6 +97,9 @@ apache_vhosts:
     extra_parameters: |
           ProxyPassMatch ^/(.*\.php(/.*)?)$ "fcgi://127.0.0.1:9000/usr/share/php/pimpmylog"

+  - servername: "{{ vagrant_ip }}"
+    documentroot: "/var/www/dashboard"
+
 apache_remove_default_vhost: true
 apache_mods_enabled:
   - expires.load
@@ -125,6 +128,10 @@ nginx_hosts:
     root: "/usr/share/php/pimpmylog"
     is_php: true

+  - server_name: "{{ vagrant_ip }}"
+    root: "/var/www/dashboard"
+
+
 nginx_remove_default_vhost: true

 # MySQL Databases and users. If build_makefile: is true, first database will
diff --git a/provisioning/tasks/dashboard.yml b/provisioning/tasks/dashboard.yml
index 142f198..bf52847 100644
--- a/provisioning/tasks/dashboard.yml
+++ b/provisioning/tasks/dashboard.yml
@@ -1,6 +1,11 @@
+- name: Ensure the dashboard directory exists
+  file:
+    path: /var/www/dashboard
+    state: directory
+    mode: 0755
+
 - name: Copy dashboard page into place.
   template:
     src: ../templates/dashboard.html.j2
-    dest: /var/www/html/index.html
-    force: yes
+    dest: /var/www/dashboard/index.html
     mode: 0644
```

geerlingguy:
> Easiest would be to continue bloating example.config.yml I guess. Any other ideas?

Let's do it. I have some long term ideas for config.yml... but I'm happy to keep it growing for now as long as little features like this are enabled by that growth!

oxyc:
Still requires a bit of testing I think.

geerlingguy:
The XHProf role has been updated, and 1.1.1 contains `install_dir`.

oxyc:
Needs fixup if https://github.com/geerlingguy/drupal-vm/pull/408 gets merged.

geerlingguy:
> That would be a breaking variable addition though.

I'm planning on bumping the minor release (at least) for the next revision of Drupal VM. Sheesh, with all these little improvements adding up to something huge, it's worthy of 3.x imo :/

I'm okay with a breaking variable change—we have had a couple in the 2.x cycle already, let's keep adding a tiny bit of pain and then I'll try to re-stabilize once we get the rest of these sweeping architectural fixes in place.

oxyc:
I could also add a revertable commit that uses a `default` filter in the spirit of https://github.com/geerlingguy/drupal-vm/issues/425 if you prefer.

Edit: Just realized we have another breaking change already though: `build_makefile` path (https://github.com/geerlingguy/drupal-vm/pull/412#issuecomment-180552019)

oxyc:
I'm having some very odd connection issues where I'm located at the moment so cant provision this locally. It still needs testing

oxyc:
> fatal: [localhost] => {'msg': "AnsibleError: file: /var/www/drupalvm/provisioning/templates/dashboard.html.j2, line number: 110, error: expected token '=', got 'end of statement block'", 'failed': True}

So the block `set` tag wasnt added until [Jinja 2.8](https://github.com/mitsuhiko/jinja2/blob/master/CHANGES#L28) which apparently isnt available on CentOS.

geerlingguy:
Yeah, macros are a little more widely supported in my experience (though I generally avoid using them if possible because templates start getting more logic-y—in this case it's definitely necessary).

oxyc:
Yeah, `set`-block was a lot more readable :/ Unfortunately I have no clue what's going on with my connection, so I cant test this locally until next week.

@geerlingguy any requests or changes to get this merged?

geerlingguy:
Yeah, as long as the template is contained, I'm fine with it being a little complex. Rather that than add complexity to one of the other bits of the system!

provisioning/tasks/dashboard.yml
```diff
@@ -0,0 +1,11 @@
+- name: Ensure the dashboard directory exists
+  file:
+    path: "{{ dashboard_install_dir }}"
+    state: directory
+    mode: 0755
```
geerlingguy: Could you add (for backwards compatibility, and to give people a way to disable it if they want):

```
  when: dashboard_install_dir is defined
```

To both this task and the next? Or if it looks cleaner, inside `provisioning/playbook.yml`.

oxyc: Sure, I would also need to add a check to the drush alias templates. That's no biggy though.

oxyc: So the `default` filter does not work in the `set` assignment. O.o

oxyc:
@geerlingguy I force updated the last commit so it uses a variable file instead. I think it's a lot nicer actually and something we could extend with more options later.

Tested on ubuntu 14.04 and centos 7 both with and without the `dashboard_install_dir`.

geerlingguy:
WIP no more... I'm loving this, and we can tweak things in a follow-up. Everything about this screams AWESOME!

Issue #379, opened 2016-01-15
Invalid command 'ProxyPassMatch'


robdecker:
During a new build, I get the following error:

```
NOTIFIED: [geerlingguy.apache | restart apache] *******************************
failed: [ddvm] => {"failed": true}
msg: Output of config test was:
AH00526: Syntax error on line 14 of /etc/apache2/sites-enabled/vhosts.conf:
Invalid command 'ProxyPassMatch', perhaps misspelled or defined by a module not included in the server configuration
Action 'configtest' failed.
The Apache error log may have more information.


FATAL: all hosts have already failed -- aborting
```

The relevant portion from `config.yml`:
```
apache_vhosts:
  - servername: "thesite.dev"
    serveralias: "www.thesite.dev"
    documentroot: "/var/www/thesite"
    extra_parameters: |
          ProxyPassMatch ^/(.*\.php(/.*)?)$ "fcgi://127.0.0.1:9000/var/www/thesite"
```

geerlingguy:
If you're running from master and have Ansible installed locally, make sure you have force updated all the Ansible Galaxy roles using the command in the README.

robdecker:
I have updated Ansible using `$ sudo ansible-galaxy install -r provisioning/requirements.yml --force` and have updated the box as well. Could the `serveralias` be the issue?

geerlingguy:
It shouldn't be; what OS are you using?

-Jeff Geerling

robdecker:
Host OS is `OS X 10.11.2` and the vagrant box is `geerlingguy/ubuntu1404`

robdecker:
And no, `serveralias` does not affect the error above.

geerlingguy:
Can you do a diff of your config.yml vs `example.config.yml` and see if there might be some other parameter missing? I'm going to re-test everything on a separate OSX Install.

robdecker:
Never mind. Re-downloaded the repo and all is well.

geerlingguy:
Nice! Glad you could get it working.

ptsimard:
I have the exact same problem. 

```
TASK [geerlingguy.apache : Ensure Apache has selected state and enabled on boot.] ***
[...]

PR #405, opened 2016-01-29
Helps with #379: Set drush_version explicitly to a string

oxyc:
This way the user would either leave it as a string, or maybe remember that it used to be a string when it fails. If you set `drush_version: "7"` it would still fail though as this is not the name of the branch (nor a tag).

> TASK [geerlingguy.drush : Clone Drush from GitHub.] ****************************
fatal: [drupalvmv]: FAILED! => {"changed": false, "failed": true, "msg": "Failed to checkout 7"}

Now the error messages would at least help them figure it out. But maybe it's even worth adding a comment about it being a branch, and _not_ a version :)

### DIFF
--- a/example.config.yml
+++ b/example.config.yml
@@ -172,7 +172,7 @@ ruby_install_gems: []
 # You can configure almost anything else on the server in the rest of this file.
 extra_security_enabled: false
 
-drush_version: master
+drush_version: "master"
 drush_keep_updated: true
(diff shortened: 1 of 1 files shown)


geerlingguy:
Note: I've just updated the Drush role itself to document why it should be a string. This is fine for Drupal VM's purposes. It at least makes the error message more helpful.

Issue #300, opened 2015-11-06
Removing apache vhosts from config.yml will fail drush alias creation


wroxbox:
Remove apache_vhosts from config.yml

```
apache_vhosts:
  - servername: "{{ drupal_domain }}"
    documentroot: "{{ drupal_core_path }}"
```
Change to 
```
nginx_hosts:
  - server_name: "{{ drupal_domain }}"
    root: "{{ drupal_core_path }}"
    is_php: true
```
will fail drush alias creation, because the loops need {% for vhost in apache_vhosts %}

This can be fixed by changing two files 
drupalvm-local.aliases.drushrc.php.j2

```
{% for vhost in nginx_hosts %}
$aliases['{{ vhost.server_name }}'] = array(
  'uri' => '{{ vhost.server_name }}',
  'root' => '{{ vhost.root }}',
);
{% endfor %}
```
and drupalvm.aliases.drushrc.php.j2

```
{% for vhost in nginx_hosts %}
$aliases['{{ vhost.server_name }}'] = array(
  'uri' => '{{ vhost.server_name }}',
  'root' => '{{ vhost.root }}',
  'remote-host' => '{{ vhost.server_name }}',
  'remote-user' => 'vagrant',
  'ssh-options' => '-o PasswordAuthentication=no -i ~/.vagrant.d/insecure_private_key',
);
{% endfor %}
```

Also servername should be server_name and documentroot should be root

This could be also done by providing two templates for both servers
```
- name: Configure drush aliases for root user inside VM.
  template:
    src: ../templates/drupalvm-{{ drupalvm_webserver }}-local.aliases.drushrc.php.j2
    dest: "~/.drush/{{ vagrant_machine_name }}.aliases.drushrc.php"
```
when: drupalvm_webserver == 'nginx'

wroxbox:
https://github.com/wroxbox/drupal-vm/commit/6d5a581b3591012d2e780451f4bf1ca560156f4b is what I did

geerlingguy:
D'oh! You're quite correct, I missed fixing this when I added nginx support.

geerlingguy:
Instead of two templates, I'll just do a switch on the `drupalvm_webserver`, but will hopefully get to this soon! I need to provide a little better nginx support out of the box.

oxyc:
This issue is now fixed as of https://github.com/geerlingguy/drupal-vm/issues/300

geerlingguy:
Yep!

PR #408, opened 2016-01-30
Issue #300: Fix drush aliases not scaffolded for nginx hosts

oxyc:


oxyc:
Also blacklisted the dev tools hosts that are defined in the `example.config.yml`

oxyc:
@geerlingguy you know Ansible inside and out. Are the variables being defined an intended and stable behaviour?

provisioning/templates/drupalvm-local.aliases.drushrc.php.j2
```diff
@@ -6,10 +6,26 @@
  * @see example.aliases.drushrc.php.
  */
 
-{% for vhost in apache_vhosts %}
-$aliases['{{ vhost.servername }}'] = array(
-  'uri' => '{{ vhost.servername }}',
-  'root' => '{{ vhost.documentroot }}',
+{% set devtool_hosts = [adminer_install_dir, pimpmylog_install_dir, '/usr/share/php/xhprof_html'] %}
```
geerlingguy: One note; I think the xhprof path differs on CentOS. I think we might need to add a variable upstream for that, similar to the other install_dir vars you're using.

provisioning/templates/drupalvm.aliases.drushrc.php.j2
```diff
@@ -6,13 +6,29 @@
...
 
-{% for vhost in apache_vhosts %}
-$aliases['{{ vhost.servername }}'] = array(
-  'uri' => '{{ vhost.servername }}',
-  'root' => '{{ vhost.documentroot }}',
-  'remote-host' => '{{ vhost.servername }}',
-  'remote-user' => 'vagrant',
+{% set devtool_hosts = [adminer_install_dir, pimpmylog_install_dir, '/usr/share/php/xhprof_html'] %}
+
+{% macro alias(host, root) %}
```
geerlingguy: I like this; didn't think of defining a macro here.

geerlingguy:
@oxyc - Can you fix this in line with #410 which was just merged?

oxyc:
Hups

frob:
Way to do @oxyc ++

Issue #257, opened 2015-09-26
Make ALL THE THINGS use vagrant_machine_name for the domain


robdecker:
Most of the time I'm running more than one drupal-vm machine at a time since I have to bounce around quite a bit. So when I'm configuring the vm, I will alter the `vagrant_machine_name` to the name of my project, and then edit all the `apache_vhosts` domains (adminer, xhprof, pimpmylog) to match this new `vagrant_machine_name`.

I propose that all the `apache_vhosts` domain names and all the `nginx_hosts` domain names use `vagrant_machine_name` that is set on line 10.

geerlingguy:
The main problem this causes (it was done in the past) is that for those using the Vagrant hostsupdater plugin, their hosts file doesn't get updated because there are Jinja2-style variables in the domain names. If someone can come up with a simple substitution system that will fill in the main domain inside the Vagrantfile so the auto hosts still work, that would be a good solution.

robdecker:
There are six hosts plugins at https://github.com/mitchellh/vagrant/wiki/Available-Vagrant-Plugins#etchosts-approach but I do not know if any would work in this scenario.

frob:
We could use gulp or npm scripts to generate the config from a simplified config.

oxyc:
There's an open PR for this issue at https://github.com/geerlingguy/drupal-vm/pull/184.

geerlingguy:
This is fixed in #410 - Thanks for all the great work on this!

PR #184, opened 2015-07-30
Added support for autohosts with ansible variables.

shawnachieve:
This is related to the following issues and Pull Requests:
* https://github.com/geerlingguy/drupal-vm/pull/182
* https://github.com/geerlingguy/drupal-vm/issues/143

This PR updates the Vagrantfile logic for the hostsupdater plugin to support adding host file entries for the following:
* Apache vhost servername with the {{ drupal_domain}} ansible variable in them.
* Apache vhost serveralias with the {{ drupal_domain}} ansible variable.
* Catches an edge case when a user configures their vm host name to be different than the drupal host name.

NOTE: My Ruby/Ansible skills aren't as strong as in some other languages.  So I welcome any feedback or suggestions on better ways of implementing this update. :-)

PR #410, opened 2016-01-31
PR #184, Issue #257: Add support for jinja variables in `config.yml` replacing hardcoded hosts and paths

oxyc:
Related PR/Issues:
- https://github.com/geerlingguy/drupal-vm/pull/184
- https://github.com/geerlingguy/drupal-vm/issues/257

This is really a second variation of #184 that's more generalised and supports jinja variables in any config value (as well as variables pointing to other variables: `servername: "{{ drupal_domain }}"` -> `drupal_domain: "{{ vagrant_machine_name }}"`).

[Beetbox](https://github.com/drupalmel/beetbox/blob/master/Vagrantfile#L38:L43) does something similar, but I added a recursive walker so we can support nested hashes/lists etc.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -72,6 +72,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
         unless blacklist.include?(host['servername'])
           aliases.push(host['servername'])
         end
+        aliases.concat(host['serveralias'].split()) if host['serveralias']
       end
     else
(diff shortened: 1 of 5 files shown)


oxyc:
Added a commit using jinja variables in docroots as well. Added a `fact` as per discussion in https://github.com/geerlingguy/ansible-role-php-xhprof/pull/10.

PR #411, opened 2016-02-01
Follow-up #410: Use vagrant hostname instead of machine name for servernames

oxyc:
Hups! I did my last batch of testing with another `config.yml` so didn't catch this! `vagrant_machine_name` is by default set to _drupalvm_ while the `vagrant_hostname` is _drupalvm.dev_


geerlingguy:
...I should've caught that too... was flying through a few too many PRs today :D

PR #378, opened 2016-01-15
Add support for storing drupal vm in a subdirectory

oxyc:
This is a work in progress and currently (probably) doesn't work with:
- `ansible_local`
- correct `post_provisioning_scripts` paths

I'm adding my initial proof of concept described in https://github.com/geerlingguy/drupal-vm/issues/305#issuecomment-156702888 as well as some docs on how to set it up.

Would be great to get some feedback and suggestions, so I'm opening this PR for more discussion. Personally I really like this setup as it's fairly simple, doesn't add too much code and allows me to store Drupal VM as a git submodule.

In the discussion I linked to @frob discovered one drawback though. You need to provision the machine in the project root, as the `.vagrant` directory needs to be created in the same location as the delegating `Vagrantfile` or else multiple `.vagrant` folders will be created. Actually now that I think about it, it's probably only an issue if you go through the delegat**ed** Vagrantfile from Drupal VM.

I honestly don't think that scenario is worth fixing... But it could probably be prevented by doing something similar to http://stackoverflow.com/a/25904132/319855

Issue #84, opened 2015-04-25
Make Drupal VM prod-ready (on a DigitalOcean 1GB Droplet)


geerlingguy:
This VM would do nicely on a 1 GB (or greater) Linode of DigitalOcean VM, or really anywhere (server or VM) where you would want to host a typical Drupal site.

While it doesn't (yet) include Nginx or Varnish as a reverse proxy (which would make for huge gains in performance for a typical Drupal site), it has the entire stack in place for a decent and secure production environment. Plus, those comfortable developing with the Drupal VM could use the exact same stack/configuration for production (or other environments), making this a more flexible environment.

I think I can make a couple tweaks to the playbook (to do things like _not_ disable AppArmor on production, based on a variable, and to add the `geerlingguy.security` role for extra security hardening), and then add a guide for 'setting up a Drupal VM production server on DigitalOcean... or Linode... etc."

It's a similar setup to how I deploy and locally test www.midwesternmac.com, www.hostedapachesolr.com, www.servercheck.in, etc.

While I've had the idea in my head a bit, it was really @wizonesolutions, @attiks and @stevepurkiss who kicked off this issue—see: https://twitter.com/wizonesolutions/status/591958086648274944

Steps to success:

- [x] Add new 'extra_security_enabled' variable (to be used as toggle for security-related features)
- [x] Don't disable AppArmor MySQL config when secure variable is true
- [x] Use `geerlingguy.security` role when secure variable is true
- [x] Add `example-prod.config.yml` file with example configuration for 1GB VM baseline (targeting DigitalOcean or Linode, for starters... but it will be generic enough to run anywhere).
- [x] <del>Determine whether it's easier to [specify a command-line-passed vars filename](https://groups.google.com/d/msg/ansible-project/553v9wTGbk4/cBjKM5egENYJ) for the override for prod config, or specify `extra-vars` via CLI.</del>
- [ ] Add documentation page "Drupal VM prod on DigitalOcean" with a tutorial
- [x] Copy the configured make file up to the server into /tmp directory (instead of using shared `/vagrant/*` path)—possibly create new issue for this.

geerlingguy:
Hopefully I'll have this ready by the Ansible for Drupal Infrastructure BoF on Wednesday at DrupalCon LA! https://events.drupal.org/losangeles2015/bofs/ansible-drupal-infrastructure-and-deployments

wizonesolutions:
I suppose I'll have to come to that BoF :o

[...]

Issue #346, opened 2015-12-14
Switch Apache to use FPM instead of mod_php


geerlingguy:
By default, Apache 2.4 will use mod_php instead of FPM. I'd like to have Apache use FPM instead, to ensure extra consistency with Nginx and better overall performance.

geerlingguy:
Steps to manually switch (at least on Ubuntu 14.04):

  1. `sudo apt-get install -y libapache2-mod-fastcgi`
  2. `sudo nano /etc/apache2/mods-enabled/fastcgi.conf`, and add the following inside:

      <IfModule mod_fastcgi.c>
        AddType application/x-httpd-fastphp5 .php
        Action application/x-httpd-fastphp5 /php5-fcgi
        Alias /php5-fcgi /usr/lib/cgi-bin/php5-fcgi
        FastCgiExternalServer /usr/lib/cgi-bin/php5-fcgi -socket /var/run/php5-fpm.sock -pass-header Authorization
        <Directory /usr/lib/cgi-bin>
          Require all granted
        </Directory>
      </IfModule>

  3. `sudo a2enmod actions`
  4. `sudo service apache2 restart`

If it were just Ubuntu 14.04, I'd be adding the support immediately.

geerlingguy:
It's actually much simpler, it seems... adding the line below will proxy requests through PHP-FPM in each Drupal site virtualhost:

```yaml
apache_vhosts:
  - servername: "{{ drupal_domain }}"
    documentroot: "{{ drupal_core_path }}"
    extra_parameters: |
          ProxyPassMatch ^/(.*\.php(/.*)?)$ "unix:/var/run/php5-fpm.sock|fcgi://localhost{{ drupal_core_path }}"
```

And as long as we have Apache 2.4 (e.g. use the below task to set it to 2.4 on Ubuntu 12.04), we can just enable mod_proxy_fcgi, and the above proxypass works wonderfully:

```yaml
- name: Add repository for Apache 2.4 on Ubuntu 12.04.
  apt_repository: repo='ppa:ondrej/apache2'
  when: ansible_distribution_version == "12.04"
```

I'm updating the freshly-renamed `geerlingguy.apache-php-fpm` role now, and will rename it on galaxy or delete/re-add it...

geerlingguy:
[...]

PR #412, opened 2016-02-01
Use Docker-based Travis CI tests for multi-platform support

oxyc:
I realised that we probably can't integrate these for a while due to stuff like https://github.com/geerlingguy/ansible-role-php-xhprof/issues/8, but I thought I'd send it over for some brainstorming.

1. I came up with at least a temporary solution for moving the `drush_makefile` to the VM before running it, instead of relying on it being a _vagrant_ environment. If https://github.com/geerlingguy/drupal-vm/pull/378 gets merged, we should pass the `host_project_root` var to ansible and use that instead. This issue was identified earlier in #84.

2. Also re-arranged the order of how drush files are generated so if run locally at least the local alias will be kept instead of overridden by an invalid remote alias.

3. Besides that I added some basic sanity checks to see that some of the stuff in the `example.config.yml` works out of the box. If we can get travis provisioned correctly, maybe we should create a `tests/` directory and move the tests there instead (so you could also run them locally). One rather nice test framework that could fit is [bats](https://github.com/sstephenson/bats). I've used it on a couple of projects before.

If you don't think we can run the entire playbook on `travis` (maybe just the test duration is too annoying already), I can cherry pick the other commits to separate PRs if you want. I guess even without the added tests, this would be a good step forward to solving #84.

### DIFF
--- a/provisioning/tasks/build-makefile.yml
+++ b/provisioning/tasks/build-makefile.yml
@@ -1,4 +1,10 @@
 ---
+- name: Copy drush makefile into place.
+  copy:
+    src: "{{ drush_makefile_path }}"
+    dest: /tmp/drupal.make.yml
+  when: drupal_site.stat.exists == false
+
 - name: Ensure drupal_core_path directory exists.
   file:
     path: "{{ drupal_core_path }}"
@@ -9,7 +15,7 @@
 
 - name: Generate Drupal site with drush makefile.
   command: >
-    {{ drush_path }} make -y {{ drush_makefile_path }} --no-gitinfofile
+    {{ drush_path }} make -y /tmp/drupal.make.yml --no-gitinfofile
     chdir={{ drupal_core_path }}
   when: drupal_site.stat.exists == false
   become: no
--- a/tests/centos-6-vars.yml
+++ b/tests/centos-6-vars.yml
@@ -1,0 +1,1 @@
+drupalvm_webserver: nginx
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
       version: 6
       init: /sbin/init
       run_opts: "--privileged"
+      additional_vars: "centos-6-vars.yml"
     - distribution: centos
       version: 7
       init: /usr/lib/systemd/systemd
@@ -54,7 +55,7 @@ script:
   - 'sudo docker exec "$(cat ${container_id})" cp /var/www/drupalvm/$MAKEFILE /var/www/drupalvm/drupal.make.yml'
 
   # Append additional variables if set.
-  - '[[ $additional_vars ]] && sudo docker exec "$(cat ${container_id})" bash -c "cat /var/www/drupalvm/tests/${additional_vars} >> /var/www/drupalvm/config.yml"'
+  - '[[ $additional_vars ]] && sudo docker exec "$(cat ${container_id})" bash -c "cat /var/www/drupalvm/tests/${additional_vars} >> /var/www/drupalvm/config.yml" || true'
 
   # Ansible syntax check.
   - 'sudo docker exec --tty "$(cat ${container_id})" env TERM=xterm ansible-playbook /var/www/drupalvm/provisioning/playbook.yml --syntax-check'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,27 +1,103 @@
----
-language: python
-python: "2.7"
+sudo: required
 
-install:
-  # Install Ansible.
-  - pip install ansible
+env:
+  global:
+    - CONFIG: example.config.yml
+      MAKEFILE: example.drupal.make.yml
+      HOSTNAME: drupalvm.dev
+      MACHINE_NAME: drupalvm
+  matrix:
+    - distribution: centos
+      version: 6
+      init: /sbin/init
+      run_opts: "--privileged"
+    - distribution: centos
+      version: 7
+      init: /usr/lib/systemd/systemd
+      run_opts: "--privileged --volume=/sys/fs/cgroup:/sys/fs/cgroup:ro"
+      additional_vars: "centos-7-vars.yml"
+  # - distribution: ubuntu
+  #   version: 14.04
+  #   init: /sbin/init
+  #   run_opts: "--privileged"
+  # - distribution: ubuntu
+  #   version: 12.04
+  #   init: /sbin/init
+  #   run_opts: "--privileged"
 
-  # Add ansible.cfg to set local roles path.
-  - "{ echo '[defaults]'; echo 'roles_path = ../'; } >> ansible.cfg"
+services:
+  - docker
+
+before_install:
+  # Pull container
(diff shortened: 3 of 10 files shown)


geerlingguy:
@oxyc - I'm currently running every individual role in Travis (it's an Ubuntu 12.04 environment by default), and that works well enough.

I definitely would like to add a 'tests' directory in the root, and put in whatever extra testing stuff we need there.

My super-amazing-moonshot-endgoal would be to set up tests like: https://github.com/geerlingguy/ansible-role-apache/blob/master/.travis.yml

It would bootstrap one Docker environment for each of the supported OSes, and inside those Docker environments, run `playbook.yml`. We can't test Vagrant inside Travis' infrastructure, but we can test all the Ansible bits, and treat Docker as an additional destination besides Vagrant VMs.

oxyc:
The docker setup you have for the apache role is pretty impressive for sure. You said you were ironing out some kinks still?

Even if the individual roles have tests, I feel the project could benefit from some integration tests to catch regressions. The tests added in this commit would have caught https://github.com/geerlingguy/drupal-vm/pull/411 for example. Would also verify that `drush` aliases work. Probably a few other slightly fragile pieces that we could test.

oxyc:
Adminer test passes on travis already!

Any idea what causes the failing test, something travis specific?
```
TASK: [Ensure MySQL AppArmor profile is disabled (for slow query log).] ******* 
failed: [localhost] => {"failed": true, "path": "/etc/apparmor.d/disable/usr.sbin.mysqld", "state": "absent"}
msg: Error while linking: [Errno 2] No such file or directory
```

geerlingguy:
@oxyc - This is the main reason I'd like to isolate inside Docker—these kinds of errors pop up all the time because Travis does some _insane_ stuff to their environments. Not to mention they have about a jillion different things installed that break every preconceived notion you can make about a default Ubuntu 12.04 install :)

And yes, I have some kinks to work out for my Ansible roles, but if we can lift what's there for Drupal VM, I have a feeling it'll be very helpful here, and since this use case is different than my individual roles, I wouldn't wait until I work out the rest of the kinks to use the same setup.

> Even if the individual roles have tests, I feel the project could still benefit from some integration tests to catch regressions.

Yes, 100%—I think I was a little unclear in my post. I would really like some Drupal VM integration tests! Can't get perfect, since I don't have a spare Mac and Windows laptop laying around to do end-to-end testing with Vagrant... but we can get pretty close by testing all the Ansible bits inside Docker containers.

oxyc:
Ah makes sense. Okay will _mostly_ hold off on this then. Now I'm just curious to see how far it can go before bailing so I might push a few changes.

geerlingguy:
@oxyc - Sounds good—it's Travis' infrastructure, so feel free to hit it hard with more tests :)

oxyc:
Well that was surprising, just ignoring one error and it passes.

oxyc:
Some subtasks related to this:

- To get CentOS 7 working we either need to A) provide a secondary playbook (difficult to maintain) B) Fix upstream `ansible-role-mysql` to set the default required packages for mariadb https://github.com/geerlingguy/ansible-role-mysql/issues/55.
- Figure out how to get `upstart` or even `systemd` working on Ubuntu 12.04 and 14.04 https://github.com/geerlingguy/ansible-role-mysql/pull/84#issuecomment-179952919

oxyc:
Aha https://github.com/geerlingguy/drupal-vm/issues/346#issuecomment-171869866! That explains why my CentOS6 build kept failing!

So I modified the test so `nginx` is tested on `CentOS 6` as `apache` doesn't work out of the box anyways.

So with this we're testing:
- `CentOS 7` with apache
- `CentOS 6` with nginx

Let's hope the tests pass :) @geerlingguy what do you think about my `build_makefile` change? Will probably cause a lot of invalid bug reports but it would also help solve #84.

Making sure the AppArmor `disabled/` directory exist is something worth checking right? It failed on the ubuntu containers and I noticed I usually verified conf-directory existence in other roles.

provisioning/tasks/build-makefile.yml
```diff
@@ -1,4 +1,10 @@
 ---
+- name: Copy drush makefile into place.
+  copy:
+    src: "../../{{ drush_makefile_path }}"
```
geerlingguy: Can we just include the `../../` in the path and that way people can specify absolute or relative paths to use?

tests/centos-6-vars.yml
```diff
@@ -0,0 +1 @@
+drupalvm_webserver: nginx
```
geerlingguy: Such a nice, simple change :)

geerlingguy:
If this passes, I think the only change I'd request for now is to update the path for the makefile to include the `../../`

oxyc:
Fixed the path, it's definitely better that way.

An earlier build already passed :) https://travis-ci.org/geerlingguy/drupal-vm/jobs/107331207

tests/centos-6-vars.yml
oxyc: I was a bit worried of how to override things at first. Then I just tried to see what happens if you re-declare a YAML variable with a different value :smile: No errors!

geerlingguy:
@oxyc - So is the only holdup to getting ubuntu12/14 tests going the mysql tests? I'd like to focus on getting those fixed asap so we can unblock a whole chain of different tests and make this the most automatedly-tested awesome project ever :)

oxyc:
Currently that's what's failing yes. And any other role that relies on `upstart`

oxyc:
Builds are pretty slow though hehe. Wonder why it's specifically CentOS 6 every time. And more specifically: `Ensure PHP packages are installed.`

geerlingguy:
I wonder if the 6.x PHP repos are just really slow or something :/

geerlingguy:
@oxyc - I'm going to merge this now, just because it gives a heck of a lot better coverage than we're currently providing. We will continue to work on the Ubuntu support for MySQL and here, thanks so much for the help on the `geerlingguy.security` role and for this. This is awesome. Mind-blowingly awesome!

oxyc:
Hmmm https://travis-ci.org/geerlingguy/drupal-vm/builds/108702338

oxyc:
So seems it's _just_ unstable. The tests for https://github.com/geerlingguy/drupal-vm/pull/404 passed.

geerlingguy:
@oxyc - That earlier build was due to a mailhog test gone bad; I updated a path in the upstream daemonize role, but forgot to update the default in mailhog as well.

okapies:
@oxyc @geerlingguy 

Hi. I also get trapped in this problem on Travis this week and found that `docker run -d --privileged /sbin/init` makes CPU 100% and lead to death. You should use such like `docker run -d -it /bin/bash` instead.

https://bugzilla.redhat.com/show_bug.cgi?id=1046469

geerlingguy:
> lead to death

Welcome back from the dead then! 😄 

Thanks for posting this tidbit.

Issue #425, opened 2016-02-10
Possible to version roles?


MichaelMallett:
We use drupalvm in a git submodule. Today I ran the install requirements command, and then went to provision, and ever since I now find that I get an 502 and php-fpm won't create the socket file. I can only assume there is now a change in the roles. I've since pulled master for the drupalvm and now I get

`file or module does not exist: /Users/Mike/Sites/rawpixel/rawpixel-test/drupalvm/provisioning/post_provision_scripts`

Would be very useful to be able to roll back to a previous commit and return back to the working roles, similar to a Gemfile.lock file.

geerlingguy:
The roles are version end, but there is currently no locking, unfortunately; it's not a feature yet supported by Galaxy (though to be be pedantic, I could specify particular versions in requirements.yml, but I am not willing to do that at this time, for the maintenance burden). See upstream issue https://github.com/ansible/ansible/issues/13496 for some other ideas.

geerlingguy:
Also, for the above error, it is simply caused by a missing folder, and not by one of the roles, as it's a new feature that's built into a new version of the Drupal VM playbook. Locking role versions wouldn't help resolve these delta-version feature additions.

At this point in the project, every time there's a new release, the release notes point to specific changes, and if you encounter any issues, you should diff your config.yml with the latest example.config.yml and see what new variables may have been added.

I am not able/willing to maintain the project with 100% backwards compatibility in major or minor point releases, only in patch releases (I try extremely hard to follow semver—for the VM and for all the roles used).

MichaelMallett:
I appreciate the effort in trying to maintain this project, however I would argue that avoiding introducing a breaking change in a minor release is not a case of maintaining backwards compatibility. If this was a drupal module it would be expected to be automatically updatable without introducing a critical error - especially when introducing an optional feature as is the case here.

In this case, adding a check for the folder, or variable, would allow this feature to be introduced without causing an issue for automated builds. I have created a pull request for this.

PR #426, opened 2016-02-10
Check for post_provisioning_scripts being defined before attempting t…

MichaelMallett:
…o run them

Was getting this error before this check
file or module does not exist: /Users/Mike/drupalvm/provisioning/post_provision_scripts

### DIFF
--- a/provisioning/playbook.yml
+++ b/provisioning/playbook.yml
@@ -79,3 +79,4 @@
     - name: Run configured post-provision shell scripts.
       script: "{{ item }}"
       with_items: post_provision_scripts
+      when: post_provision_scripts is defined


geerlingguy:
If you have `post_provision_scripts: []` inside `config.yml`, this task doesn't fail, it just skips due to the list being empty (see: https://github.com/geerlingguy/drupal-vm/blob/master/example.config.yml#L215-L218)

In this case, if upgrading from an older version of Drupal VM, you should run a diff of your older config.yml with the example.config.yml to make sure any new vars are added in. It's not ideal, and I might find a better way to deal with this problem (adding new variables) soon, but it is what it is for now and if I add a conditional for one new variable I'd have to do it for all of them :(

geerlingguy:
I'll keep this open for now, per discussion in #425 — as "MINOR version when you add functionality in a backwards-compatible manner, and" would mean a 2.X.x release shouldn't break 2.X-1.x code... :(

PR #427, opened 2016-02-10
Use php_fpm_listen

opdavies:
Use the nginx vhost template to use `php_fpm_listen` rather than hard-coding it into the template.

I've also opened a PR upstream to update this variable for Debian/Ubuntu servers. https://github.com/geerlingguy/ansible-role-php/pull/78


geerlingguy:
Makes sense! For the upstream issue, I left a comment there. I'd like to keep the default using the TCP/IP port for easier translatability with HHVM and other systems.

Issue #434, opened 2016-02-12
Either get CentOS 6 test build faster, or ditch it


geerlingguy:
In #412 a CentOS 6 Docker build was added to test everything in Drupal VM... and it works, but it's really really slow. Like 4-6x slower than the other environment builds.

I wonder if something in that Docker base image is screwy (a network setting?), or if it's just something crazy-slow that we won't be able to fix. So that we can get CI feedback in < 10 minutes, I vote to either disable it or fix it, and if it takes more than a couple hours' effort and we're getting nowhere, we just disable it and give everyone the caveat that only the latest OS releases have full automated test coverage.

geerlingguy:
cc @oxyc who originally mentioned the issue.

oxyc:
Wondering if this is what's going on:

- https://www.reddit.com/r/devops/comments/3omkvm/anyone_else_have_slow_docker_builds_on_centos/
- https://github.com/docker/docker/issues/10248#issuecomment-71690895

oxyc:
I also feel like the build started quicker when we defined `python` as our language. Now there's a long delay before the tests even get started. Do they have a set amount of ruby boxes (we always get a ruby box now), and a set amount of python boxes? Or maybe it's just some travis free-account limit...

geerlingguy:
Not sure... it's hard sometimes to gauge, because free accounts for OSS are always in a pool with no particular guarantees, so sometimes I've waited up to an hour for a particular instance to get going.

oxyc:
I vote for ditching it and merging #456. Testing it on the latest versions is good enough imo. If we decide to ditch centos 6, we need to change Ubuntu 14.04 to test the nginx scenario though.

geerlingguy:
Sounds good to me... maybe have ubuntu 14.04 (and moving forward, whatever the default is inside `example.config.yml`) run both the `nginx` and `apache`. Otherwise we could do both for both of the latest supported versions, but it's more important to me to make sure the defaults run across CentOS and Ubuntu.

geerlingguy:
Working on this now.

PR #435, opened 2016-02-12
Use OverlayFS instead of loopback

oxyc:
Just trying something out #434

Based on http://www.projectatomic.io/blog/2015/06/notes-on-fedora-centos-and-docker-storage-drivers/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,14 @@ services:
   - docker
 
 before_install:
+  - >
+    [[ "${distribution}-${version}" == "centos-6" ]]
+    && sudo service docker stop
+    && sudo rm -rf /var/lib/docker
+    && sudo sh -c 'echo "DOCKER_STORAGE_OPTIONS= -s overlay" >> /etc/sysconfig/docker-storage'
+    && sudo service docker start
+    || true
+
   # Pull container
   - 'sudo docker pull ${distribution}:${version}'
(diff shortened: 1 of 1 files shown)


oxyc:
Seems like this wont work anyway:

> RPMs and Yum. OverlayFS only implements a subset of the POSIX standards. This can result in certain OverlayFS operations breaking POSIX standards. One such operation is the copy-up operation. Therefore, using yum inside of a container on a Docker host using the overlay storage driver is unlikely to work without implementing workarounds.
https://docs.docker.com/engine/userguide/storagedriver/overlayfs-driver/

geerlingguy:
:(

oxyc:
Seems it was fixed for CentOS 7 (probably what it's using and why it's fast) but doesnt look like it was backported to CentOS 6. https://bugzilla.redhat.com/show_bug.cgi?id=1213602#c51

The fix seems to be in [`yum-utils-1.1.31-34.el7.src.rpm`](https://rhn.redhat.com/errata/RHBA-2015-2129.html) avaiable for CentOS 7. Latest version for CentOS 6 is `1.1.30`.

Also https://github.com/docker/docker/issues/10180

oxyc:
Okay so this is a dead-end. Overlay isnt even available.

oxyc:
For reference it uses `aufs`.

Issue #349, opened 2015-12-16
Add documentation for SSL support


kevinquillen:
Did a quick search of the docs and couldn't find anything on https, is it possible to setup a VM with http/https support? I want to be able to use a self signed cert and something like securepages to test a local Drupal 7 installation with.

geerlingguy:
See https://github.com/geerlingguy/drupal-vm/pull/162 for a quick example—I really need to add more info to the docs! Also, check out the documentation in the `geerlingguy.apache` role for more info on SSL config.

tripper54:
Following #162 I have SSL enabled, but Apache is just serving php files as text rather than passing them to FPM. Can you give me any pointers as to what I'm doing wrong? Thanks!

oxyc:
@tripper54 that's because Drupal VM recently made a switch to PHP-FPM which requires some changes in your configuration files. You can find instructions in the [release notes](https://github.com/geerlingguy/drupal-vm/releases/tag/2.2.1). If you prefer not switching you can also find instructions on how to continue using `mod_php` [here](https://github.com/geerlingguy/drupal-vm/blob/4da53e4a5794062c9aabe1cef3724c0ab3a80e3c/example.config.yml#L204:L207). Also #394, #393

oxyc:
This can now be closed https://github.com/geerlingguy/drupal-vm/pull/438.

PR #438, opened 2016-02-16
Issue #349, #162: Add SSL docs

oxyc:
Is there any other way to transfer the certificate?
```sh
vagrant up
^Ctrl-C
vagrant ssh -c 'sudo mv /vagrant/example.cert ...'
vagrant provision
```

Just worried about suggesting something potentially dangerous if people use it for other things than simply testing SSL.

### DIFF
--- a/docs/extras/ssl.md
+++ b/docs/extras/ssl.md
@@ -1,0 +1,11 @@
+If you are using Ubuntu as your base OS and you want to get started quickly with a local development environment you can use the snakeoil certificate that is already generated.
+
+```yaml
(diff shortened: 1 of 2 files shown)


geerlingguy:
I think the docs have sufficient warning in them right now... but yeah... we might want to add an extra disclaimer across the top of the new docs page that security is _your_ responsibility, so be educated. Something to that extent. Kind of like I did here: https://github.com/geerlingguy/ansible-role-security#ansible-role-security-basics

Issue #417, opened 2016-02-05
Trying to install nodejs 4.x or 5.x results in error


kevinquillen:
```
TASK: [geerlingguy.nodejs | Ensure Node.js and npm are installed.] ************ 
failed: [local.sandbox] => {"failed": true}
msg: Could not fetch updated apt files
```

I have tried using:
* 5.5.0
* 5.x
* 4.2.6
* 4.x

When I ssh into the machine and do node --version, it reports 0.12.9.

How can I get the latest stable version running inside of the VM?

geerlingguy:
The `geerlingguy.nodejs` role uses the Nodesource binary sources, so you should be able to set `4.x`, `5.x`, `0.12`, or `0.10` without issue. However it only installs once, and after that if you want to change versions inside the VM, you can either do so manually by installing `nvm` and switching on the fly, or you have to reset things by hand.

Is this on a _fresh_ VM provision, or are you reprovisioning an existing VM that was built with 0.12?

kevinquillen:
SSHing in and removing it, then specifying the version as 4.3.0 worked. Rpeating that and trying 5.6.0 did not work.

If I removed it, then did sudo apt-get install in the VM instead of the config.yml, 5.6.0 installs successfully.

I also noticed that on provisioning, npm isn't always installed... and I get this error:

```
TASK: [geerlingguy.nodejs | Add NodeSource repositories for Node.js.] ********* 
ok: [local.d8contrib] => (item=deb https://deb.nodesource.com/node_4.x trusty main)
ok: [local.d8contrib] => (item=deb-src https://deb.nodesource.com/node_4.x trusty main)

TASK: [geerlingguy.nodejs | Ensure Node.js and npm are installed.] ************ 
failed: [local.d8contrib] => {"failed": true}
msg: Could not fetch updated apt files
```

When I SSH into the VM, it reported npm as not being installed... and then reported node as not being installed, even though previously it had installed it. 

Is there something funky in the nodejs role? My success rate on getting it installed is low, whereas from command line the VM will get and install NodeJS 5.6.0 without an issue.

geerlingguy:
See also: https://github.com/geerlingguy/ansible-role-nodejs/pull/16

kevinquillen:
Whats the solution here? Is this an upstream issue? I get the task failed on every run now.

frob:
There might be an upstream task, but for here we need to at least document it better.

geerlingguy:
[...]

PR #447, opened 2016-02-17
Update example.config.yml

frob:
Added version numbers to the nodejs version description. To stem confusion from #417

### DIFF
--- a/example.config.yml
+++ b/example.config.yml
@@ -161,7 +161,11 @@ installed_extras:
 extra_packages:
   - unzip
 
-# `nodejs` must be in installed_extras for this to work.
+# `nodejs` must be in installed_extras for this to work. This currently only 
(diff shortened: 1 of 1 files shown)


example.config.yml
```diff
@@ -161,7 +161,11 @@ installed_extras:
 extra_packages:
   - unzip
 
-# `nodejs` must be in installed_extras for this to work.
+# `nodejs` must be in installed_extras for this to work. This currently only 
+# runs on the initial provision. If the version needs to chanage 'nvm' will need
```
geerlingguy: I don't want to add this amount of detail here. I'd be willing to accept:

```
# `nodejs` must be in installed_extras for this to work. Valid examples: "0.10",
# "0.12", "4.x", "5.x".
```

Issue #444, opened 2016-02-17
Vagrant up fails due to hostsupdater


jvdschraaf:
The vagrant hostsupdater plugin fails with on (some versions of) OS X. See the issue here: https://github.com/cogitatio/vagrant-hostsupdater/issues/92

This issue was reported in December last year and still has not been fixed. Maybe it would be an idea to support multiple hosts manager plugins? This way people have an alternative if one option fails. 

I have tested locally with https://github.com/smdahlen/vagrant-hostmanager and it works perfectly for me. If you are interested I could submit a pull request for supporting this plugin next to hostsupdater.

geerlingguy:
@jvdschraaf - Does `sudo chmod +a "$USER allow write,append" /etc/hosts` work for you?

As an alternative, you can set your sudoers file to allow you write access to /etc/hosts as well... http://www.jeffgeerling.com/blogs/jeff-geerling/better-vagrant-development-workflow

I wouldn't be opposed to supporting both, but it seems if one has trouble with permissions the other might have the same kind of trouble show up at different times in different ways. I haven't seen the issue in the issue linked above before, FWIW.

jvdschraaf:
@geerlingguy expanding permissions on /etc/hosts works, but its not a solution I like using as anything other than a temporary workaround.

This issue is not purely a permissions problem on my machine, as hostsupdater fails to even ask for my sudo password. Looking at the issues referenced in https://github.com/cogitatio/vagrant-hostsupdater/issues/92 multiple projects have opted to either switch completely to a different hosts manager or support multiple managers.

geerlingguy:
@jvdschraaf - The alternative allows only Vagrant to manage the hosts file (which seems a good/convenient compromise, as long as you trust the Vagrant configs you use...), but I'm definitely happy to support both projects if it's a simple addition.

oxyc:
I believe I'm having this issue too. It happened on two Macbook 12" shipped Yosemite and upgraded to El Capitan, while it worked fine on a Macbook 12" that came with El Capitan already installed (please don't ask why i've on my third laptop haha).

Looks like a very minor change to just gather the aliases and then add them to either `config.hostsupdater.aliases` or `config.hostmanager.aliases` depending on which one is installed.

geerlingguy:
Fixed in #453 - Thanks @jvdschraaf !

PR #453, opened 2016-02-18
Add support for vagrant-hostmanager plugin.

jvdschraaf:
This enables users to choose between using vagrant-hostsupdater and vagrant-hostmanager. See #444


geerlingguy:
Failure is due to an upstream role but that I've fixed but for some reason hasn't been imported to Galaxy. I'll try to test this today so I can merge.

Can you also update the README and/or docs to mention install of hostsupdater or host manager?

geerlingguy:
Also referencing #444

PR #456, opened 2016-02-21
Add custom initctl script for fixing docker tests on ubuntu 14.04

oxyc:
Let's see what happens

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,10 +18,10 @@ env:
       init: /usr/lib/systemd/systemd
       run_opts: "--privileged --volume=/sys/fs/cgroup:/sys/fs/cgroup:ro"
       additional_vars: "centos-7-vars.yml"
-  # - distribution: ubuntu
-  #   version: 14.04
-  #   init: /sbin/init
-  #   run_opts: "--privileged"
+    - distribution: ubuntu
+      version: 14.04
+      init: /sbin/init
+      run_opts: "--privileged"
   # - distribution: ubuntu
   #   version: 12.04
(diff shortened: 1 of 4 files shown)


oxyc:
AppArmor is being _really_ annoying. In some cases `state=restarted` ansible triggers a stop followed by a start. If you run `service apparmor stop`, the following warning comes

> All profile caches have been cleared, but no profiles have been unloaded. Unloading profiles will leave already running processes permanently unconfined, which can lead to unexpected situations. To set a process to complain mode, use the command line tool 'aa-complain'. To really tear down all profiles, run the init script with the 'teardown' option.
https://travis-ci.org/geerlingguy/drupal-vm/jobs/110755901#L2082

If I switch to reload instead of restart it fails with `AppArmor not available as kernel LSM`. I'm assuming this is a Docker issue and _if_ it occurred on another system, it's not something we could fix anyways, therefore I'm ignoring it. I found a similar bug report for [edx provisioner](https://groups.google.com/forum/#!topic/openedx-ops/KgDklKGaqzk).

If I avoid restarting AppArmor, and doing what is documented in the Ubuntu docs about disabling a service, https://help.ubuntu.com/community/AppArmor#Disable_one_profile, it fails with
> Warning: unable to find a suitable fs in /proc/mounts, is it mounted?\
https://travis-ci.org/geerlingguy/drupal-vm/jobs/110773476#L2082

To avoid all this headache, and add support systems without AppArmor installed, I added a conditional to ignore the task if AppArmor is not installed.

PR #472, opened 2016-02-26
Rebase of #292 Ruby coding style fixes via rubocop in Vagrantfile

oxyc:
A rebase of #292. Also added rubocop check to travis.

According to https://docs.travis-ci.com/user/ci-environment/#Runtimes we do not need to specify the language as ruby because it's always installed.


oxyc:
I was considering adding a dedicated lint build on travis but the additional gem installation only adds 3sec so not needed yet at least.

PR #473, opened 2016-02-26
Adding a note about the drush aliases NOT being valid for Windows by …

justinlevi:
Hey guys, 

Love the new dashboard! Nice work on the latest couple of releases. 

One thing I noticed is that the drush aliases that come up are nice if you're not on Windows. Might be a good idea to add a little note about that. There is probably a better way to handle this as well but though I'd give it a shot. 

drush-aliases-note


geerlingguy:
Makes sense to me; technically there are situations where the aliases won't be created on the host either... And actually, I may adjust the wording to be a little more accurate in terms of the aliases existing inside the VM only on Windows (so you can still use them, but only if logged in).

Issue #479, opened 2016-02-27
Dashboard host line is backwards


ctrladel:
On my dashboard the host line for the dashboard is backwards but all the rest are fine.

0.0.0.0 guser.local
0.0.0.0 adminer.guser.local
0.0.0.0 xhprof.guser.local
0.0.0.0 pimpmylog.guser.local
dashboard.guser.local 0.0.0.0

PR #480, opened 2016-02-27
Fix various alias issues in dashboard, hostsupdater and drush aliases

oxyc:
**Needs manual testing** I'm on an even slower connection today, so wont be able to test it until some other day.

- This fixes: #479
- Simplifies the alias logic in `Vagrantfile`
- Support nginx aliases as well as redirects for hostsupdater/hostsmanager
- Fixes dashboard hosts entry logic never to ignore hosts, only ignore hosts when the entry becomes `192.168.88.88 192.168.88.88` (dashboard).
- Fixes broken drush aliases when nginx aliases are used.

I do not create drush aliases for every host alias intentionally. Only the first one will be used.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -64,28 +64,25 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
   # If a hostsfile manager plugin is installed, add all server names as aliases.
   aliases = []
-  blacklist = [config.vm.hostname, vconfig['vagrant_ip']]
   if vconfig['drupalvm_webserver'] == 'apache'
     vconfig['apache_vhosts'].each do |host|
-      unless blacklist.include?(host['servername'])
-        aliases.push(host['servername'])
-      end
+      aliases.push(host['servername'])
       aliases.concat(host['serveralias'].split) if host['serveralias']
     end
   else
     vconfig['nginx_hosts'].each do |host|
-      unless blacklist.include?(host['server_name'])
-        aliases.push(host['server_name'])
-      end
+      aliases.concat(host['server_name'].split)
+      aliases.concat(host['server_name_redirect'].split) if host['server_name_redirect']
     end
   end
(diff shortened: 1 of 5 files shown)


geerlingguy:
I will work on some testing later today. Home projects this weekend, which is why I've been a bit under the radar :)

oxyc:
Thanks, otherwise I can test it on Monday, I'll have time then.

- [ ] Need to verify drush alias file (both local and remote) when using nginx
- [ ] Need to check the hostsupdater aliases are gathered correctly for multiple `server_name` and multiple `server_name_redirect` on nginx
- [ ] Check dashboard

oxyc:
@geerlingguy I tested it and it works in my scenarios with multiple `server_name` as well as multiple `server_name_redirect`. Also fixed a bug where redirects didnt work for multiple hosts in `server_name`.

oxyc:
Just realized we should probably create drush aliases for webserver aliases as well. For multisite. I'll push another commit once I've tested it and verified that it works.

geerlingguy:
@oxyc - Sounds good, home projects ran a bit late today, so I'll just wait for your final changes then merge. This looks great!

oxyc:
Thanks. This should be mergable if travis passes.

geerlingguy:
Aww, now getting:

```
TASK [Configure host machine drush aliases.] ***********************************
fatal: [localhost -> 127.0.0.1]: FAILED! => {"changed": false, "failed": true, "msg": "AnsibleUndefinedVariable: 'dict object' has no attribute 'serveralias'"}
```

Happens on non-nginx configs.

oxyc:
I'm too used to twig!

geerlingguy:
Yeah, Twig and Jinja2 are soooo close in most cases, it's those little things :)

geerlingguy:
Passed. Merged!

Issue #146, opened 2015-07-02
Add PostgreSQL support


geerlingguy:
With https://www.drupal.org/node/2157455 finally having been completed, it would be nice if users could switch from MySQL/MariaDB/Percona to PostgreSQL (at least for Drupal 8) optionally.

I don't currently have a Postgres role, but could add one or find a good one in Galaxy. Besides also installing `php-mysql`, there's not much else that needs to be conditionally changed besides swapping out Posgres for MySQL.

gitoutthere:
This actually what I need at the moment, being able to add geerlingguy.pgsl to playbook along side with the PHP connector

gitoutthere:
I used ANXS.postgresql seems to be working ok

geerlingguy:
@gitoutthere - Could you leave some instructions for what changes you made in this issue, so I can point others interested in trying out Postgres with Drupal here? I'm still planning on adding 'official' support sometime, but would be glad to 'unofficially' support it as best I can for now.

gitoutthere:
@geerlingguy cool I am still trying to wrestle and get my Envs setup. But here is the steps I took to get Postgres working, in your vagrant folder:
1) Add 'ANXS.postgresql' to '/provisioning/requirements.txt'
2) Run `sudo ansible-galaxy install -r provisioning/requirements.txt --force` to install the new plugin.
3) Modify your '/provisioning/playbook.yml' and add `- ANXS.postgresql` to the roles.
4) Override settings on '/config.yml'. Config can be found on https://github.com/ANXS/postgresql

JorisVanEijden:
**ANXS.postgresql** seems to be Ubuntu only.
I've been using **patrik.uytterhoeven.PostgreSQL-For-RHEL6x** for CentOS 5,6 and 7

Would love to see these combined into a **geerlingguy.postgresql** role :+1:

geerlingguy:
@JorisVanEijden - heh, always the optimist! I'll see what I can do, now that everything in Drupal 8 is running and tested with Postgres.

JorisVanEijden:
Well, 13 of the 14 galaxy roles in use here at oneshoe are geerlingguy roles. I'd say our infra has been pretty much geerlinged since buying your book half a year ago :wink:

gitoutthere:
Any of you guys are here at DrupalCon? I would love to catch up with you.

geerlingguy:
[...]

PR #490, opened 2016-03-02
[WIP] Support PostgreSQL

oxyc:
Initial stab att PostgreSQL support #146. I've never used it before so I'll need some help here.

There's quite a bit of todos here
- [ ] @geerlingguy we need a php-pgsql role.
- [x] Figure out exactly why the php role always gets skipped when `php-mysql` is skipped.
- [x] Dashboard needs a lot of work. Might need to re-structure database connections a bit as someone could use both pgsql and mysql.
- [ ] What's the deal with admin/root login (mysql root login). No pass?
- [x] Docs
- [ ] Figure out if there are any other settings we need to configure (and document)
- [x] Get tests to pass
- [ ] Remove travis tests once this is RTBC


oxyc:
- In the dashboard I set the root postgresql password to _Passwordless_ but I guess this can be changed in the `postgresql` role. Anyone with experience who knows how we should do this? Maybe this is good enough for now.
- @geerlingguy do you want to separate the `php-pgsql` installation into a separate role?
- Anyone with PostgreSQL experience who can review this. I have no clue what I'm doing :) It works at least
- @geerlingguy how do you feel about the `backwards_compatibility.yml` task I added? I prefer this over cluttering all variable references with a default filter at least.

provisioning/tasks/init-debian.yml
```diff
@@ -29,3 +29,7 @@
   set_fact:
     php_xhprof_html_dir: "/usr/share/php/xhprof_html"
   when: php_xhprof_html_dir is not defined
+
+- name: Install PHP PostgreSQL dependencies.
```
geerlingguy: Yeah, I'd like to follow the pattern of setting this up in a separate role. Will try to get to that soon.

provisioning/tasks/install-site.yml
```diff
@@ -14,7 +14,7 @@
     --site-name="{{ drupal_site_name }}"
     --account-name={{ drupal_account_name }}
     --account-pass={{ drupal_account_pass }}
-    --db-url=mysql://{{ drupal_mysql_user }}:{{ drupal_mysql_password }}@localhost/{{ mysql_databases[0].name }}
+    --db-url={{ drupalvm_database }}://{{ drupal_mysql_user }}:{{ drupal_mysql_password }}@localhost/{{ drupal_mysql_database }}
```
geerlingguy: Wouldn't the latter variable need to work with either mysql or pgsql? Or I guess that variable could just change full stop to `drupal_database_name`, perhaps?

provisioning/playbook.yml
```diff
@@ -26,10 +28,11 @@
     - { role: geerlingguy.apache, when: drupalvm_webserver == 'apache' }
     - { role: geerlingguy.apache-php-fpm, when: drupalvm_webserver == 'apache' }
     - { role: geerlingguy.nginx, when: drupalvm_webserver == 'nginx' }
-    - geerlingguy.mysql
     - geerlingguy.php
     - geerlingguy.php-pecl
-    - geerlingguy.php-mysql
+    - { role: geerlingguy.mysql, when: drupalvm_database == 'mysql' }
+    - { role: geerlingguy.php-mysql, when: drupalvm_database == 'mysql' }
+    - { role: ANXS.postgresql, when: drupalvm_database == 'pgsql' }
```
geerlingguy: I don't know if this makes it simpler or not... but we could also allow installation of _both_ pg and mysql. Though I can't think of many people who would ever need it.

provisioning/tasks/install-site.yml
oxyc: `mysql_databases[0]` references the `drupal_mysql_database` so feels we should use that variable instead. Currently this is a bug where drush installs it into the first database defined regardless of what the `drupal_mysql_database` is set to.

Yes renaming these variables would be better imo, but if we just want to add support for PostgreSQL with minimal changes we can keep using these.

provisioning/playbook.yml
oxyc: I considered that too but then opted for the same setup we use with `drupalvm_webserver`. Maybe there are a few more people interested in using multiple database engines than multiple webservers though.

oxyc: Guess we could move both mysql and pgsql to `installed_extras`.

provisioning/templates/dashboard.html.j2
```diff
@@ -215,15 +219,18 @@
...
             <th>Database name</th>
+            <th>Engine</th>
             {% if 'adminer' in installed_extras -%}
               <th></th>
             {%- endif %}
           </tr>
         </thead>
         <tbody>
-          {% for database in mysql_databases -%}
+          {% set database_list = (drupalvm_database == 'pgsql') | ternary(postgresql_databases, mysql_databases) %}
```
oxyc: `ternary` filters were added in Ansible v1.9. Does the project have a minimum version requirement set at the moment? At the moment it's v2 because of the galaxy bug I guess.

oxyc:
@geerlingguy here's a quick patch you can use on top of the php-mysql role (I'm not a reliable maintainer) https://github.com/geerlingguy/ansible-role-php-mysql/compare/master...oxyc:master

Maybe better to name everything `postgresql` instead of `pgsql`... Reason I'm keeping to `pgsql` is because of the database uri schema for drush. Maybe I'm just making it even more confusing though

oxyc:
I also adressed your 2 suggestions. We also need to either remove `backwards_compitability.yml` or add the variable name changes for `drupal_mysql_*`. Feels like it's better to just do a major version bump and exclude it.

geerlingguy:
Version bump++

I'll do one more minor release before merging in these changes, then we can work on a couple other changes for a 3.0

oxyc:
I removed the `backwards_compitabilty.yml` and will do one final test on travis here before removing those builds too. Or do you want me to keep one in to verify we dont break it? Could minimise the build time by removing a few extra installs.

geerlingguy:
I'd keep one build - remove all the extra installs so we're just testing the one thing.

geerlingguy:
I removed the 3.0.0 milestone; don't want this to hold up that release.

geerlingguy:
@oxyc - As with all branches that sit too long... I couldn't pick up from here easily, so I started a separate branch. Thanks for your awesome work on this, and I hope you're having a great time traveling right now :)

Moving further discussion over to the original issue, #146

PR #497, opened 2016-03-04
Fixes #495: Set composer owner to vagrant user

### LINKED ISSUE #495: Composer is owned by root, provisioning scripts that overwrite package conf (like Coder) fail
I am trying to provision a VM with the drupal/coder package installed as a global package, so I added it to the config.yml. That worked just fine. 

The second step that I wanted to do was add a provisioning script to auto set the coder profile for me. So I created a basic script:

```
#!/usr/bin/env bash
phpcs --config-set installed_paths ~/.composer/vendor/drupal/coder/coder_sniffer
```

It fails with "phpcs command not found" - even though if I ssh into the box and type phpcs, the command runs.

So, I changed the script to this:

```
#!/usr/bin/env bash
cd /home/vagrant/.composer/vendor/bin
./phpcs --config-set installed_paths ~/.composer/vendor/drupal/coder/coder_sniffer
```

No error, no warning, but PHPStorm reported a config error with the tool when trying to select a profile.

I SSH'ed into the box again and navigated to /home/vagrant. I noticed that .composer is owned by root, so the command has no permission to overwrite the default conf file in the squizlabs codesniffer package.

```
vagrant@local:~$ phpcs --config-set installed_paths ~/.composer/vendor/drupal/coder/coder_sniffer
Config file /home/vagrant/.composer/vendor/squizlabs/php_codesniffer/CodeSniffer.conf is not writable
```

At this point the only thing I knew to do to get around it was ```sudo chown -Rf vagrant:vagrant .composer```, and running the phpcs command correctly installed the Drupal coder profile configuration.

Should the .composer folder be owned by root, or be owned by vagrant?

oxyc:
https://github.com/geerlingguy/ansible-role-composer/issues/17

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,7 +54,8 @@ script:
   - 'sudo docker exec "$(cat ${container_id})" cp /var/www/drupalvm/$CONFIG /var/www/drupalvm/config.yml'
   - 'sudo docker exec "$(cat ${container_id})" cp /var/www/drupalvm/$MAKEFILE /var/www/drupalvm/drupal.make.yml'
 
-  # Append additional variables if set.
+  # Append additional variables.
+  - 'sudo docker exec "$(cat ${container_id})" bash -c "cat /var/www/drupalvm/tests/test-vars.yml >> /var/www/drupalvm/config.yml" || true'
   - '[[ $additional_vars ]] && sudo docker exec "$(cat ${container_id})" bash -c "cat /var/www/drupalvm/tests/${additional_vars} >> /var/www/drupalvm/config.yml" || true'
 
   # Vagrantfile syntax check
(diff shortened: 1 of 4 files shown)


geerlingguy:
I like it... especially since it plays nice with use on DigitalOcean/other targets.

geerlingguy:
Hmm, tests failed, but probably just because it's inside the Docker containers; maybe we can set the `ansible_ssh_user` variable explicitly for the tests?

oxyc:
Ah right. Is it the same if provisioned locally with `ansible_provisioner` or JJG-Ansible-Windows?

oxyc:
This should work everywhere, also docker and `ansible_local`. For whatever reason ansible didn't have `ansible_env.USER` populated in docker (while it is on a VM) so I had to add a final default to `ansible_user_id` which does exist.

One quirk is that still you need to correct the home path if you're installing with root.

Issue #550, opened 2016-04-07
Zip & unzip not installed


heddn:
Zip & unzip are not installed as part of the base box. If I add them to extra_packages, that doesn't seem to run until after a bunch of composer stuffs. On my slow internet connection, that means vagrant provision always fails. Why should the absence of zip/unzip matter? Composer cannot use prefer-dist if those aren't available, forcing it to do a git clone. That pushes me over the 300 second limit and the composer install fails, killing the provision.

Solution: add zip/unzip as part of the base box so they are just there.

oxyc:
I don't think @geerlingguy has added anything extra to the boxes. Everything is done through ansible.

For a quick hacky solution to get this fixed right away you could do:
```
php_packages:
  - php5
  - php5-mcrypt
  - php5-cli
  - php5-common
  - php5-curl
  - php5-dev
  - php5-fpm
  - php5-gd
  - php-pear
  - php-apc
  - libpcre3-dev
  - unzip
```

This would install it before the composer task runs.

As for a more stable solution. We can't really bump the extra package install task further up as it's common to use this for "addon" packages that depend on the roles that run before. Maybe we just add it as a dependency in the init pre_tasks https://github.com/geerlingguy/drupal-vm/blob/master/provisioning/tasks/init-debian.yml.

geerlingguy:
Yeah, I'm thinking we could just do it in pre_tasks for debian (and RHEL too, if it's not available there).

geerlingguy:
Fixed by #551.

PR #551, opened 2016-04-07
Issue #550: Make unzip a dependency

oxyc:
The package is small and I don't really see a use case for someone not wanting it. What do you think?

### DIFF
--- a/example.config.yml
+++ b/example.config.yml
@@ -166,8 +166,7 @@ installed_extras:
   - xhprof
 
 # Add any extra apt or yum packages you would like installed.
-extra_packages:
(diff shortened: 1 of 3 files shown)


geerlingguy:
Looks good to me.

heddn:
The composer error specifically mentioned zip and unzip. I'm not sure if both are required. I'd have to test.

heddn:
And thanks for such fast turn around time!

heddn:
Yeah, seems unzip is all that is needed for cloning. Now no errors when doing a `composer install --prefer-dist`

PR #552, opened 2016-04-07
Fix drush core-rsync not expanding ~ to $HOME

oxyc:
`~` expansion is a shell feature that doesnt work with rsync -e flag.


oxyc:
Not saying this will solve all `drush` issues, but it's at least one bug.

You can test that `~` does not expand with:
```bash
 rsync -e 'ssh -o PasswordAuthentication=no -i ~/.vagrant.d/insecure_private_key' vagrant@drupalvm-debug.dev:~/.bashrc foobar
```

This is what drush issues under the hood using a core-rsync. Eg:

```
drush core-rsync @drupalvm.drupalvm.dev:index.php testing.php
```

oxyc:
Update, apparently `$_SERVER['HOME']` doesnt exist on windows, so I use drush helper function instead.

geerlingguy:
Nice; never knew about that shortcut.

PR #557, opened 2016-04-09
Fixes #530: Add docs for arch linux where python 3 is the default

### LINKED ISSUE #530: [Arch] SyntaxError during task 'Check if local known_hosts file is present.'
Hello,

I've been running Drupal VM (also) on a Arch Linux host for the past 8 months without any problems. To be able to make use of all the shiny new features in the latest releases, I'm currently upgrading Drupal VM from a master branch version from September ' 15 to the current master branch version.

Before starting the upgrade, I took all changed variables mentioned in the various release notes (such as updating the Apache Ansible Galaxy role & updating Apache vhost configs) into account.

After that I re-ran the provisioning of my Drupal VM box (default Drupal VM Ubuntu 14.04 box). Everything went well except for one of the last tasks, 'Check if local known_hosts file is present.' , which generates into the following error:

```yaml
fatal: [drupalvm]: FAILED! => {
    "changed": false, 
    "failed": true, 
    "module_stderr": "  File \"/home/sidneygijzen/.ansible/tmp/ansible-tmp-1459155009.84-185598221565320/stat\", line 314\n    except OSError, e:\n                  ^\nSyntaxError: invalid syntax\n", 
    "module_stdout": "",
    "msg": "MODULE FAILURE",
    "parsed": false}
```

In order to resolve this I tried the following:
- if my known_hosts file is present -> it is.
- start with a new build of the box (vagrant destroy / vagrant up) -> generates the same error
- Ansible doesn't support Python3 officially, and since Arch has Python 3 by default, I setup an virtualenv with Python 2 and ran it from there -> generates the same error

I have the feeling my Arch host, an interplay between Arch & Ansible is the cause (or maybe the format of the known_hosts file?), since I did the same upgrade of Drupal VM on a Ubuntu 15.10 host yesterday, and that went flawlessly. 
I thought to post my question here to see if someone else has stumbled into this error? Are there any issues known with Drupal VM or Ansible on a Arch host? 

Thank you in advance!

oxyc:


geerlingguy:
Nice and simple; thanks!

Issue #445, opened 2016-02-17
Switch back to JJG-Ansible-Windows to stop the bleeding


geerlingguy:
Vagrant 1.8.2 might never happen :P /s

Basically, Vagrant 1.8.1 is broken for anyone without Ansible installed for Drupal VM usage. And I have no idea when 1.8.2 will be released. So to stop the bleeding for now, I can switch back to JJG-Ansible-Windows so people can use the latest version of Drupal VM successfully on Windows or in other environments where they don't have Ansible installed locally.

I have been holding off doing this for so long because I hate intentionally regressing and stopping the use of awesome new features like `ansible_local`, but I think the bleeding has gone on long enough.

geerlingguy:
See:

  - https://github.com/geerlingguy/drupal-vm/commit/283614aca52342d654d5dd72f0a11f01e928be2b
  - http://www.jeffgeerling.com/blog/use-vagrant-18s-new-ansiblelocal-provisioner

geerlingguy:
Testing this now.

bkosborne:
FYI - Vagrant 1.8.4 is out now. I was testing with ansible-local provisioner and it seems to work fine. There's some stuff in the changelog related to ansible: https://github.com/mitchellh/vagrant/blob/master/CHANGELOG.md

geerlingguy:
@bkosborne - Nice, I had forgotten about this! See: #450

PR #559, opened 2016-04-09
[WIP] Add support for local.config.yml

oxyc:
This is a proof of concept that doesn't require Ansible 2.

New config loading logic:

1. `config.yml`
2. Include any files referenced in `extra_configs` list.
3. If available, and running through Vagrant, include `local.config.yml`
4. If available, and running directly through ansible, include `prod.config.yml`

If someone uses a vagrant plugin for deploying, such as for aws, they can add `env: prod` variable to their `config.yml` to include the prod config instead of the local config. Maybe this isn't a necessary feature?

I'm also assuming this would make #378 more difficult.

Edit: I simplified the PR to only support `local.config.yml` and `prod.config.yml`, both of which are included if available. Include order goes: `config.yml`, `prod.config.yml`, `local.config.yml`.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -24,18 +24,16 @@ def walk(obj, &fn)
   end
 end
 
-# Use config.yml, prod.config.yml and local.config.yml for VM configuration.
+# Use config.yml and local.config.yml for VM configuration.
 require 'yaml'
 dir = File.dirname(File.expand_path(__FILE__))
 unless File.exist?("#{dir}/config.yml")
   raise 'Configuration file not found! Please copy example.config.yml to config.yml and try again.'
 end
 vconfig = YAML.load_file("#{dir}/config.yml")
-# Include environment config files if available.
-['prod.config.yml', 'local.config.yml'].each do |file|
-  if File.exist?("#{dir}/#{file}")
-    vconfig.merge!(YAML.load_file("#{dir}/#{file}"))
-  end
+# Include a local.config.yml file if available.
+if File.exist?("#{dir}/local.config.yml")
+  vconfig.merge!(YAML.load_file("#{dir}/local.config.yml"))
 end
 
 # Replace jinja variables in config.
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -26,12 +26,13 @@ end
 
 # Use config.yml, prod.config.yml and local.config.yml for VM configuration.
 require 'yaml'
-vconfig = {}
 dir = File.dirname(File.expand_path(__FILE__))
 unless File.exist?("#{dir}/config.yml")
   raise 'Configuration file not found! Please copy example.config.yml to config.yml and try again.'
 end
-['config.yml', 'prod.config.yml', 'local.config.yml'].each do |file|
+vconfig = YAML.load_file("#{dir}/config.yml")
+# Include environment config files if available.
+['prod.config.yml', 'local.config.yml'].each do |file|
   if File.exist?("#{dir}/#{file}")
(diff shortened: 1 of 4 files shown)


oxyc:
Hmm just realized, this would probably break https://github.com/geerlingguy/drupal-vm/pull/557

geerlingguy:
One note—since a couple of my roles are now requiring Ansible 2.x+, I'm fine with dropping Ansible <2 support, if that makes anything easier. I just haven't had time to take a look at this in a while.

oxyc:
Actually I'm not aware of how Ansible 2.x would make it easier. You mentioned dynamic includes in #445.  I'm just using an empty fallback config together with `with_first_found` which should work pre 2.x.

What I'm not sure about is when we should include the `prod.config.yml`. I now set it to be included any time the `env` variable isn't set, and vagrant sets it.

Vagrantfile
```diff
@@ -32,6 +32,20 @@ unless File.exist?("#{dir}/config.yml")
...
 
+# Unless it's specifically set, eg. for a vagrant deployment plugin, use local as
+# the environment.
+env = vconfig.include?('env') ? vconfig['env'] : 'local'
+vconfig['extra_configs'] ||= []
+# Try including an environment specific config file.
+vconfig['extra_configs'].push("#{env}.config.yml")
+# Load all custom config files.
+vconfig['extra_configs'].each do |config_file|
+  if File.exist?(config_file)
```
oxyc: todo: this check should not be present for the manually defined `extra_configs` as ansible would fail anyways.

oxyc:
Another option would be to use environment variables
```sh
ENV=prod ansible-playbook provisioning/playbook.yml
```

Or just

```
INCLUDE=prod.config.yml ansible-playbook provisioning/playbook.yml
```

oxyc:
Simplified things to only support local and prod configs for now.

geerlingguy:
So, I like this. A lot. Where do you think we should document this feature? Maybe we could morph the current 'local vagranthost' documentation page to talk about "overriding configuration" or something more generic like that, and talk about:

  - `Vagrantfile.local`
  - `local.config.yml`
  - `prod.config.yml`

oxyc:
I'm not entirely clear on `prod.config.yml` to be honest. I thought they would be conditionally loaded somehow.

Is this the idea
- `prod.config.yml` should be committed to the repo. But what exactly would you put here and not in the regular `config.yml`?
- `local.config.yml` should be ignored and added on a user-to-user basis? Eg increasing the VM memory.

geerlingguy:
@oxyc - I'm thinking:

  - <s>`config.yml`: All defaults for all environments and all developers</s>
  - <s>`prod.config.yml`: Any particular configuration that's specific to your production environment (e.g. more PHP processes, different PHP memory limits, etc.).</s>
  - <s>`local.config.yml`: ...</s>

<s>And then we can add a note like "Currently, the way these override configurations are loaded means that any values in `prod.config.yml` will also be applied in your local environment—so if you want to have your local environment different than `config.yml`</s>

Rethinking this... how about we _only_ add `local.config.yml` for now. And we can punt a prod override setup until a later time, since that feature is still 'experimental'.

oxyc:
Err or did you mean only adding `local.config.yml` to docs but keeping the code for `prod.config.yml` in?

geerlingguy:
I mean remove all code and docs for prod.config.yml (so we only support local.config.yml overrides for now). So exactly like you've done here :)

geerlingguy:
I'm going to try to get to some manual testing tomorrow.

PR #560, opened 2016-04-10
Update php-7.md

iainhouston:
The main change is to document the `php_conf_paths` and `php_extension_conf_paths` you need to get `php.ini` settings etc into the right directories. Also added a note to say that Xdebug 2.4.0 works with PHP7.0 ... but this would depend upon another pull request I made for `ansible-role-php-xdebug`

### DIFF
--- a/docs/other/php-7.md
+++ b/docs/other/php-7.md
@@ -34,7 +34,7 @@ php_extension_conf_paths:
 php_fpm_pool_conf_path: "/etc/php/7.0/fpm/pool.d/www.conf"
 ```
 
-Also, comment out `xhprof`, `xdebug`, `redis` and `memcached` from the `installed_extras` list, as these extensions are not yet supported for PHP 7 (as of late 2015).
+Also, comment out `xhprof`, `redis` and `memcached` from the `installed_extras` list, as these extensions are not yet supported for PHP 7 (as of late 2015).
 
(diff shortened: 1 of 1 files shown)


docs/other/php-7.md
```diff
@@ -23,11 +23,21 @@ php_packages:
...
+php_extension_conf_paths:
+  - /etc/php/7.0/fpm/conf.d
+  - /etc/php/7.0/apache2/conf.d
+  - /etc/php/7.0/cli/conf.d
 php_fpm_pool_conf_path: "/etc/php/7.0/fpm/pool.d/www.conf"
 ```
 
 Also, comment out `xhprof`, `xdebug`, `redis` and `memcached` from the `installed_extras` list, as these extensions are not yet supported for PHP 7 (as of late 2015).
 
+Xdebug 2.4.0 works with PHP 7.0. Sources available at https://xdebug.org/files/xdebug-2.4.0.tgz but you'll need to update `ansible-role-php-xdebug` to pick the source from there.
```
geerlingguy: I've upped the default in that role to 2.4.0, which should work fine with PHP 5.4-7.x, so no need for this extra line anymore. Can you remove it?

iainhouston:
I removed the comment. Not quite sure how this works on GitHub. Assume you will see the change automatically on the same Pull Request?

geerlingguy:
Thanks!

PR #567, opened 2016-04-19
Add support for reverse proxying sites and apps

derimagia:
I've been playing around with syncthing as a replacement for file changes and wanted to throw it under a reverse proxy for my needs. Also can use it for mailhog.

### DIFF
--- a/example.config.yml
+++ b/example.config.yml
@@ -132,6 +132,9 @@ nginx_hosts:
   - server_name: "{{ vagrant_ip }} dashboard.{{ vagrant_hostname }}"
     root: "{{ dashboard_install_dir }}"
 
+  - server_name: "mailhog.{{ vagrant_hostname }}"
+    proxy_pass: "http://127.0.0.1:8025"
+
 nginx_remove_default_vhost: true
 nginx_ppa_use: true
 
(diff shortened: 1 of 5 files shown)


geerlingguy:
Can you make this work under Apache too? I don't want to have to maintain a feature that only works under one of the supported servers.

derimagia:
I was starting on this but after getting it to work it looks like this is already in a pull request here: https://github.com/geerlingguy/ansible-role-apache/pull/84

I committed some changes to make it so that the scripts don't freak out in case you ONLY have a proxy. It also looks like the upstream for the apache role is allowing multiple proxies per vhost.

Is there a reason why this vagrant repo has custom nginx files but not custom apache vhosts? Just wanted to be sure since the nginx work is only in this repo at the moment.

geerlingguy:
Closing for now, as there are some conflicts, and I also want to maybe work on getting better vhost support in the upstream Nginx role instead of adding specific behaviors in here.

For now, it is possible to do some modifications to add in the proxy support manually through [post-provision scripts and/or Ansible task files](http://docs.drupalvm.com/en/latest/extras/scripts/).

PR #580, opened 2016-04-29
Add CTwig PHP extension

oxyc:


oxyc:
Hmm just noticed this though

> You can also install the C extension via PEAR (note that this method is deprecated and newer versions of Twig are not available on the PEAR channel):

http://twig.sensiolabs.org/doc/installation.html#installing-the-c-extension

oxyc:
Probably not worth the effort as:

1. Doesn't compile for PHP 7 https://github.com/twigphp/Twig/issues/1695
2. Core discussion about making the extension requirement less optimistic https://www.drupal.org/node/2568247
3. PEAR package is deprecated.

PR #587, opened 2016-05-04
New Relic support, conditionally-installed

scottrigby:
Here's an initial stab at New Relic support. Untested. CC @cweagans

### DIFF
--- a/example.config.yml
+++ b/example.config.yml
@@ -169,10 +169,6 @@ installed_extras:
 # Add any extra apt or yum packages you would like installed.
 extra_packages: []
 
-# `newrelic` must be in installed_extras for this to work.
-# See all vars: https://github.com/weareinteractive/ansible-newrelic#variables
-newrelic_license_key: yourkey
-
 # `nodejs` must be in installed_extras for this to work. Valid examples: "0.10",
 # "0.12", "4.x", "5.x".
 nodejs_version: "0.12"
--- a/example.config.yml
+++ b/example.config.yml
@@ -155,6 +155,7 @@ installed_extras:
   - drupalconsole
   - mailhog
   - memcached
+  # - newrelic
   # - nodejs
   - pimpmylog
   # - redis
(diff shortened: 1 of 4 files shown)


oxyc:
You also need to add an entry to https://github.com/geerlingguy/drupal-vm/blob/master/provisioning/requirements.yml

cweagans:
I'd also suggest adding a documentation page for how to get this set up, ideally including some information on adding `php_value newrelic.appname "Whatever"` to the vhost to be able to set the New Relic app name on a per-site basis.

example.config.yml
```diff
@@ -168,6 +169,10 @@ installed_extras:
 # Add any extra apt or yum packages you would like installed.
 extra_packages: []
 
+# `newrelic` must be in installed_extras for this to work.
+# See all vars: https://github.com/weareinteractive/ansible-newrelic#variables
+newrelic_license_key: yourkey
```
geerlingguy: I'd rather leave this out by default, and have a new docs page under 'Extra software and setup' that describes how to enable New Relic support. I have a feeling that the number of people who will use New Relic with Drupal VM is fairly small (percentage-wise), so I'd rather leave these three lines out of the example config.yml

provisioning/playbook.yml
```diff
@@ -43,6 +43,7 @@
     - { role: geerlingguy.pimpmylog, when: '"pimpmylog" in installed_extras' }
     - { role: geerlingguy.daemonize, when: '"mailhog" in installed_extras' }
     - { role: geerlingguy.mailhog, when: '"mailhog" in installed_extras' }
+    - { role: franklinkim.newrelic, when: '"newrelic" in installed_extras' }
```
geerlingguy: As @oxyc mentioned, we need to add the role to the `requirements.yml` file as well.

scottrigby: OK Added in f3a465d.

example.config.yml
scottrigby: @geerlingguy OK, I moved this to docs/extras in d72058f.

scottrigby:
@cweagans  OK, check out `docs/extras/newrelic.md` (in d72058f). Is there anything else you think is needed for documentation?

cweagans:
:+1: from me. Looks like feedback has been taken into account.

cweagans:
(For the record, the app name per vhost thing is a nice-to-have, IMO. I don't think many people will use that, and Pubstack doesn't have it either)

geerlingguy:
Thanks, everyone!

PR #588, opened 2016-05-04
Support an optional playbook override file (example use: adding custom roles)

scottrigby:
Here's a PR for discussion. This mimicking what we did on another project inspired by your Ansible talk at DrupalCon Austin @geerlingguy 😄 

Looking through your [examples](https://github.com/geerlingguy/drupal-vm/tree/master/examples) it wasn't clear to me if there is a better way to add custom roles than this?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@ vagrant_ansible_inventory_default
 config.yml
 drupal.make.yml
 Vagrantfile.local
+provisioning/playbook.local.yml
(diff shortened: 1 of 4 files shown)


scottrigby:
I'm going to close this in favor of recommending a custom `Vagrantfile.local` with custom provisioning logic. If there's an outcry for the added convenience of this idea we can re-open it 😉

PR #593, opened 2016-05-05
Fixes NewRelic errors from the Playbook as the repo at franklinkim.ne…

vml-jbartlett:
Fixes NewRelic errors from the Playbook as the repo at franklinkim.newrelic isn't currently shared to the public.

Causes the following error:

```
ERROR! the role 'franklinkim.newrelic' was not found in ~/Sites/drupal-vm/provisioning/roles:~/Sites/drupal-vm/provisioning:/usr/local/etc/ansible/roles

The error appears to have been in '~/Sites/drupal-vm/provisioning/playbook.yml': line 46, column 7, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

    - { role: geerlingguy.mailhog, when: '"mailhog" in installed_extras' }
    - { role: franklinkim.newrelic, when: '"newrelic" in installed_extras' }
      ^ here

Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.
```

### DIFF
--- a/provisioning/playbook.yml
+++ b/provisioning/playbook.yml
@@ -43,7 +43,7 @@
     - { role: geerlingguy.pimpmylog, when: '"pimpmylog" in installed_extras' }
     - { role: geerlingguy.daemonize, when: '"mailhog" in installed_extras' }
     - { role: geerlingguy.mailhog, when: '"mailhog" in installed_extras' }
-    - { role: franklinkim.newrelic, when: '"newrelic" in installed_extras' }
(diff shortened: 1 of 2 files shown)


oxyc:
It is public: https://galaxy.ansible.com/franklinkim/newrelic/

You probably just need to install it locally:

```sh
sudo ansible-galaxy install -r provisioning/requirements.yml --force
```

vml-jbartlett:
How odd... I know that I've provisioned that a few times since last night. The NR repo came down without issue and provisions fine for me now. Burning this away as unnecessary!

PR #600, opened 2016-05-10
Fixes #598: Add support for pre_provision_scripts.

### LINKED ISSUE #598: Add support for `pre_provision_scripts`
I suppose it should be run after all the roles are installed, but before the custom tasks run. I assume most scenarios where this is useful is after the basics are already installed (git for example).

geerlingguy:
First approach—allow scripts to be run prior to role installation... note that this means things like Git aren't guaranteed to be available, and pre-provision scripts would need to at a minimum install any required dependencies before running.

If a script needs other dependencies present, it may be best to use a post provision script instead...?

### DIFF
--- a/example.config.yml
+++ b/example.config.yml
@@ -226,8 +226,9 @@ composer_home_group: "{{ drupalvm_user }}"
 # composer_global_packages:
 #   - { name: phpunit/phpunit, release: '@stable' }
 
-# Run specified scripts after VM is provisioned. Path is relative to the
-# `provisioning/playbook.yml` file.
+# Run specified scripts before or after VM is provisioned. Path is relative to
+# the `provisioning/playbook.yml` file.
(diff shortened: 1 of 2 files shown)


geerlingguy:
See: https://github.com/geerlingguy/drupal-vm/issues/598

oxyc:
One DX annoyance by having it here is that tools like `git` aren't available.

geerlingguy:
@oxyc - But I wonder, would it be more important to have this happen prior to all roles, or after all roles?

geerlingguy:
Going with it... `git` would be the main hindrance I can think of, but for those who will potentially use this feature, asking them to manually add a line `sudo apt-get install -y git` to their script before using git shouldn't be too much a burden.

I just think it makes the most sense to put it before any roles/major configuration is done, otherwise it's not very 'pre' anything.

PR #601, opened 2016-05-10
Fixes #505: Add ansible.cfg file to set a roles_path in Drupal VM directory

### LINKED ISSUE #505: Consider adding `ansible.cfg` and set roles path to the Drupal VM directory
...this would allow for a few different things:

  - Users would not need to use `sudo` to install roles, since they would be installed inside a `roles` directory inside the Drupal VM directory (if we set `roles_path` in the config file).
  - Users could modify their `requirements.yml` file to lock down a specific version (albeit manually for now, since Galaxy doesn't have any true 'lockfile' like PHP's Composer or Rubygems, see: https://github.com/ansible/ansible/issues/13496).
  - We could set a couple other custom settings if needed.

A few references for further reading:

  - http://docs.ansible.com/ansible/galaxy.html#roles-path
  - http://leucos.github.io/ansible-files-layout/

I'd also like to take a stab at improving the Galaxy documentation page linked above to include better/more comprehensive examples in the [role requirements files documentation](http://docs.ansible.com/ansible/galaxy.html#advanced-control-over-role-requirements-files).

geerlingguy:
See: https://github.com/geerlingguy/drupal-vm/issues/505

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ Notes:
     - Copy `example.config.yml` to `config.yml`.
   3. Create a local directory where Drupal will be installed and configure the path to that directory in `config.yml` (`local_path`, inside `vagrant_synced_folders`).
   4. Open Terminal, `cd` to this directory (containing the `Vagrantfile` and this README file).
-  5. (If you have Ansible installed on Mac/Linux) Run `$ sudo ansible-galaxy install -r provisioning/requirements.yml --force`.
+  5. (If you have Ansible installed on Mac/Linux) Run `$ ansible-galaxy install -r provisioning/requirements.yml --force`.
   6. Type in `vagrant up`, and let Vagrant do its magic.
 
 Note: *If there are any errors during the course of running `vagrant up`, and it drops you back to your command prompt, just run `vagrant provision` to continue building the VM from where you left off. If there are still errors after doing this a few times, post an issue to this project's issue queue on GitHub with the error.*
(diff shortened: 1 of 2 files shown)


geerlingguy:
I'm going to merge this _now_; and for the last stable 2.x release, it will be part of the tag. That way people can lock in the role versions per-VM.

I might or might not start versioning all the roles inside `requirements.yml` as well... but I'm thinking not. If people just get in the habit of only running the `ansible-galaxy` command when they do an upgrade, their versions should match the current latest stable role set anyways.

If Drupal VM is ever not primarily maintained by me, this would need to change... but as I work with both the roles and the VM, it's pretty safe.

geerlingguy:
Also, _yay for no sudo!_

oxyc:
👍 once 1.8.2 is released and we go back to `ansible_local` we should also add the possiblity to define a custom `requirements.yml` path the local installer. Otherwise windows users wont be able to lock down their versions.

geerlingguy:
@oxyc - JJG-Ansible-Windows uses our defined file correctly, doesn't it? So currently, we're in the clear?

oxyc:
I didn't know you were planning to lock in the versions here in Drupal VM so I imagined one had to create their own `requirements.yml` and reference that during `ansible-galaxy install` for locking in role versions. That would not have been possible for people using either JJG-Ansible-Windows or `ansible_local` because the path was hardcoded. Without hacking the provided `requirements.yml` file that is.

But now that we do lock in the versions (yey in my opinion) I don't see a reason for someone wanting to change the versions :) They can just downgrade their Drupal VM version. Soooo... nevermind!

Issue #608, opened 2016-05-12
Lock Ansible role versions to a specific release


geerlingguy:
Since Ansible doesn't support a syntax like `3.*` or `3.x-stable`, the next best option for favoring stability over maintainability is to specify the version of each role in requirements.yml.

While there is a maintenance burden to doing this, I think the stability and formality will help reduce the number of issues opened up in the queue for people who update Drupal VM but don't update roles, because we can always automatically point people to the simple 'if you update drupal vm, run this command' documentation.

I'm testing this now, to make sure that specifying specific versions works like I think it will...

geerlingguy:
Relates to / follows https://github.com/geerlingguy/drupal-vm/issues/505

geerlingguy:
Added a fix for this—the other thing this will empower me to do is to do a major 'break-all-the-things' PHP role revision that will switch the default to PHP 7.0, drop support for PHP < 5.5, drop CentOS 6 and Ubuntu 12.04 support—all while still supporting any users who need those things by letting them continue to use the 2.x Drupal VM branch.

I don't know just how long the 2.x branch will continue to work, but with the roles locked down to these versions, I feel a _lot_ more comfortable literally _breaking all the things_ to trim a lot of the annoying Drupal VM fat that's accumulated!

![1400jf](https://cloud.githubusercontent.com/assets/481677/15202989/f1eecdba-17c4-11e6-85a6-ce88e92d870c.jpg)

oxyc:
How do you feel about running the galaxy install step directly through Vagrant with the `ansible.galaxy_role_file`? That would remove another step from the quick start guide and people would never have missing roles after upgrading. The downside is provisioning will take a bit longer (maybe a way to turn it off?)

geerlingguy:
> How do you feel about running the galaxy install step directly through Vagrant with the `ansible.galaxy_role_file`?

Makes sense to me—that's something that we could do now, without 1.8.2, right? But I might not want to make it default... maybe some way of having users enable it only if Drupal VM is updated... or maybe some sort of a lock file that could be used to indicate when that option needs to be set, then the lock file is deleted as part of the Vagrantfile/provision process... (I'm just thinking of how we could make it automatic but not waste an extra 5-10 min of time for every provisoning run).

oxyc:
I opened a PR with some of my thoughts https://github.com/geerlingguy/drupal-vm/pull/611

PR #611, opened 2016-05-12
Followup #608: Automate ansible-galaxy install step

oxyc:
https://github.com/geerlingguy/drupal-vm/issues/608#issuecomment-218793792

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,8 +72,7 @@ Notes:
     - Copy `example.config.yml` to `config.yml`.
   3. Create a local directory where Drupal will be installed and configure the path to that directory in `config.yml` (`local_path`, inside `vagrant_synced_folders`).
   4. Open Terminal, `cd` to this directory (containing the `Vagrantfile` and this README file).
-  5. (If you have Ansible installed on Mac/Linux) Run `$ ansible-galaxy install -r provisioning/requirements.yml --force`.
-  6. Type in `vagrant up`, and let Vagrant do its magic.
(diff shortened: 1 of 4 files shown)


oxyc:
Figured out a way to disable the galaxy step as well. Create a `Vagrantfile.local` with

```rb
config.vm.provisioners[0].config.galaxy_role_file = nil
```

I doubt any regular user will ever re-provision unless they are upgrading Drupal VM. I think it's mostly developers of Drupal VM who might do it and become annoyed by the added time. But by adding this to our `Vagrantfile.local` it wouldn't be an issue for us either.

geerlingguy:
> I doubt any regular user will ever re-provision unless they are upgrading Drupal VM. In which case they do want to re-install the roles. I think it's mostly developers of Drupal VM who might run re-provisions frequently and become annoyed by the added time for provisioning. I believe this hack solves that minor annoyance.

Well... that's true. Hmm. I'm going to think through it and just make sure it won't cause any other weird frustrations that we haven't thought of yet. I like this a lot though. #MakeItMaintainable

oxyc:
So galaxy support was added in 1.8.0 and did have a few bugs but all of them are filed under `provisioners/ansible_local` in the CHANGELOG. So I think that means we're fine.

geerlingguy:
Your other amazing PR has introduced a merge conflict :)

geerlingguy:
Git just can't handle the awesomeness that will be 3.0.0.

oxyc:
Fixed.

Issue #614, opened 2016-05-12
Cron Jobs Not Appearing


attheshow:
I'm trying to set up the cron jobs to run but I'm not seeing them when I run `vagrant ssh` followed by `crontab -l`. It simply says `no crontab for vagrant`. I hope I'm following the correct syntax on this since this is my first time setting these up in the config.yml file:

```
# Cron jobs are added to the root user's crontab. Keys include name (required),
# minute, hour, day, weekday, month, job (required), and state.
drupalvm_cron_jobs:
  - {
    name: "Richland Cron",
    hour: "*/3",
    minute: "15",
    job: "drush -r /var/www/drupalvm/richland cron"
  }
  - {
    name: "BYL Cron",
    hour: "*/3",
    minute: "20",
    job: "drush -r /var/www/drupalvm/byl cron"
  }
  - {
    name: "LCS Cron",
    hour: "*/3",
    minute: "25",
    job: "drush -r /var/www/drupalvm/lcs cron"
  }
```

oxyc:
These are actually added under the `root` user. But sounds like that's something we should fix.

attheshow:
Duh, I see that in the comment above now. Sorry. So how do you switch to the root user? I tried `su -` with password `root` but received the message `su: Authentication failure`.

oxyc:
You can check using the `sudo` command: `sudo crontab -l`

attheshow:
Awesome. Thanks for your help!

oxyc:
@geerlingguy should we add the crons under the vagrant user instead?

oxyc:
For the VM alone i don't think it's worth changing the functionality. But for people running the provisioning on a production environment I dont think these should be added under the root user.

oxyc:
@attheshow heads up, in the next version of Drupal VM this will be changed to how you first expected it to work. The cron jobs will be added to the `vagrant` user. Thanks for pointing this out 👍

geerlingguy:
Also a quick way to change to the root user, in case you ever need to: `sudo su` :)

attheshow:
Great. Thanks for the consideration and the tips!

PR #615, opened 2016-05-12
Issue #614: Install cron jobs as the SSH user instead of as root

oxyc:
This is mostly a security measurement for users provisioning a production environment.


geerlingguy:
Needs no discussion, makes perfect sense. Good catch!

Issue #616, opened 2016-05-13
Fix all the things!


geerlingguy:
Well, at least the small things that I had to leave slightly-disjointed after #609:

  - [x] Need to re-enable the XHProf integration test in `.travis.yml`
  - [x] Need to pass through docs and make sure everything is updated to reflect 
  - [x] Need to update www.drupalvm.com with correct/latest information
  - [x] Need to switch main installation Travis tests to use Ubuntu 16.04 instead of 14.04 Docker.

oxyc:
Do you have an ETA on the next release? I need to upgrade a bunch of VMs and would rather wait until 3.0 unless it's far off.

geerlingguy:
++ to @oxyc for helping push this through!

geerlingguy:
@oxyc - As far as the next release goes, I was hoping to do it yesterday or today, but my energy level is down to like -5... so it will hopefully be either tomorrow by EOD, or very early in the week.

Thank you SO much for your help with all these last-minute hurdles! The last two things I want to get in are tagged in the 3.0.0 milestone, and then I want to work on one more test run pass on my Macs and PCs to make sure the 3.0.0 release is as smooth as possible.

PR #620, opened 2016-05-13
Issue #616: Re-enable the XHProf integration test.

geerlingguy:
See: https://github.com/geerlingguy/drupal-vm/issues/616

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .bundle/
 *.retry
 vagrant_ansible_inventory_default
+local.config.yml
 config.yml
 drupal.make.yml
 Vagrantfile.local
--- a/.travis.yml
+++ b/.travis.yml
@@ -84,11 +84,11 @@ script:
     && (echo 'Pimp my Log install pass' && exit 0)
     || (echo 'Pimp my Log install fail' && exit 1)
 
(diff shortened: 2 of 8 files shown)


oxyc:
The diff I linked to earlier was to the `installed_extras` list where xhprof is disabled :)

geerlingguy:
@oxyc - Umm... yeah, that would do it, wouldn't it :P

oxyc:
This will conflict with https://github.com/geerlingguy/drupal-vm/pull/624/commits/42148c327cd0725b4e9a0d5bb666aa34b0d732e0

oxyc:
Oh I changed the travis logic here. Instead of appending the contents of appending to the default `local.config.yml` I just use a different file if the `local_config` variable is set. So you would have to add the `installed_extras` list to the `ubuntu-16-nginx.config.yml` as well. Felt a bit easier to understand but maybe not.

oxyc:
I did that mostly because I remembered this commit https://github.com/geerlingguy/drupal-vm/commit/3046b6bdaf35d1ca7f6a9214e0e6a2f0b051b76f :)

geerlingguy:
@oxyc - No worries!

PR #623, opened 2016-05-14
Add support for custom pre/post ansible tasks.

thom8:
Allow for custom playbook tasks without modifying the playbook.

You can define `custom_pre_tasks_directory` and/or `custom_post_tasks_directory` in your config to a path which will contain custom tasks. 

These are run in alphabetical order.

### DIFF
--- a/provisioning/playbook.yml
+++ b/provisioning/playbook.yml
@@ -21,6 +21,11 @@
       with_items: "{{ pre_provision_scripts }}"
       when: pre_provision_scripts is defined
 
+    - include: "{{ item }}"
+      with_fileglob:
+        - "{{ custom_pre_tasks_directory }}"
+      when: custom_pre_tasks_directory is defined
+
     - name: Set the PHP webserver daemon correctly when nginx is in use.
       set_fact:
         php_webserver_daemon: nginx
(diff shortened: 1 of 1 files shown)


oxyc:
I'd definitely prefer this over shell scripts 👍

opdavies:
👍

geerlingguy:
@thom8 - Do you think you could also add a new docs page under 'Other', titled 'Running custom scripts and playbooks' (or something along those lines), with instructions for how someone could do a simple playbook (at least), and pointing to the `examples` directory.

I might also add in a playbook that does the same thing as the existing example Search API Solr configuration...

geerlingguy:
er... just realized we already have `docs/extras/scripts.md` — let's change that page to something that encompasses both shell scripts and playbooks, and get it in.

oxyc:
@thom8 if you're busy I don't mind writing the docs.

But I'm wondering, maybe we should rename these variables to align with `pre_provision_scripts`. Maybe  `pre_provision_tasks_dir` instead (pretty much all role varibles use _dir_). Or maybe just `pre_provision_tasks` as it doesn't necessarily have to be a directory.

oxyc:
Also we need to either define this variable in `example.config.yml` or add `default(omit)`.

thom8:
@oxyc that would be great, if you like we can close this PR and create another with the docs.

We named it `beet_custom_*_tasks` and provide defaults `.beetbox/tasks/pre` & `.beetbox/tasks/post` but not sure if it makes sense in the DVM context.

https://github.com/beetboxvm/beetbox/blob/master/provisioning/ansible/playbook-provision.yml#L16

oxyc:
#633 but I might just be too tired to write docs at the moment :D

thom8:
Closing for #633

PR #624, opened 2016-05-14
Issue #616: Switch travis tests from Ubuntu 14.04 to 16.04

oxyc:


oxyc:
More use cases for `local.config.yml` :)

oxyc:
Regarding AppArmor being annoying https://github.com/geerlingguy/drupal-vm/pull/456

oxyc:
Just hacking it to see if travis passes. I think this should still pass on a real system, but not sure. If I recall `update.rc-d` doesn't work in docker containers.

oxyc:
Hmpf didnt work to hack it either. Any ideas? I can have another look later today as well.

oxyc:
Yey

oxyc:
Verified that the apparmor task do run when installed locally.

geerlingguy:
Awesome. Thank you SO much for fixing this up!

PR #625, opened 2016-05-14
Issue #616: Bring docs up to date [ci skip]

oxyc:


oxyc:
From reading through once, this is what I found. Is this section correct though http://docs.drupalvm.com/en/latest/other/base-os/?

geerlingguy:
@oxyc - Yes, at this point ALL the OSes besides CentOS 6 work out of the box :)

Issue #618, opened 2016-05-13
Templates for various Drupal VM use cases


nerdstein:
DrupalVM can be used for various use cases.

The current make file and configuration YML appear to be focused on a core-contributing use case.

One other prominent use case is doing local sandboxes for a Drupal-based project. This has different needs, which include pinning of stable versions of Drupal, Drush, and dependencies (non bleeding edge so that projects can focus on their development without being burdened with potential other issues).

I'm proposing we provide at least one more example.make and example.config.yml optimized for projects. There are considerations like "working copy" of core that projects wouldn't benefit from because updating core becomes harder.

I would like to discuss the potential use cases and identify improvements to meet a broader set of needs.

oxyc:
Personally I always keep my development modules in the dev branch and then disable them on production.

As for `drush`, from a maintenance perspective it makes sense to keep it at master so we don't have to update it every time they release a new version. I agree that this can be an issue though. Nowadays I use [drupal-composer/drupal-project](https://github.com/drupal-composer/drupal-project/blob/8.x/composer.json#L23) as the base for my sites and there I pin down the latest stable version with composer. This way the the globally installed drush delegates to the project specific drush version.

geerlingguy:
> Edit: I hadn't even noticed Drupal VM uses working copy of core. Maybe we should remove this for the common use case? I bet most people don't know what this is.

Yeah, that's one of the main things. We could just comment it out and let people uncomment if they're doing core dev work or using Drupal VM at a workshop.

nerdstein:
The working copy is one thing. But, there are more settings ideal for initializing a Drupal project and having DrupalVM facilitate sandboxes, e.g.: pulling from a stable release of Drush, pulling stable release for Core, etc. I'm sure there is more and we need an approach that separates building a Drupal site from doing core development. This is less about the actual modules installed - I fully expect people to know and understand how to use a make file or Drush for getting what they want.

My gut tells me we should consider separate config.yml and/or make files for various use cases. At a minimum, a different set of configuration for core development and developing a website in Drupal.

fubarhouse:
@nerdstein Let me assure you that with the use of things you can code yourself, DrupalVM is far more powerful than it might look on its shell. I'm now developing node, python and bash as well as drupal and php comfortably without ever worrying about it dying or being restricted.

The latest release actually allow for different configuration files - though make files are rather set and are in fact inevitably going to become redundant...

[...]

PR #630, opened 2016-05-15
Issue #618: Disable working-copy of drupal core by default

oxyc:


oxyc:
Travis gets a bit faster too

Disable Working Copy | Without git | Without git verification
----------|-------|--------
9m 12s   | 8m 41s | 8m 15s
10m 17s | 8m 35s | 9m
10m 40s | 9m 33s | 9m 35s

PR #633, opened 2016-05-16
PR #623: Add support for custom pre/post ansible tasks.

oxyc:
The first sentence is probably the longest in history but not enough sleep to think clearly at the moment.

### DIFF
--- a/docs/extras/scripts.md
+++ b/docs/extras/scripts.md
@@ -1,4 +1,6 @@
-Drupal VM allows you to run extra shell scripts in the beginning and at the end of the provisioning process, in case you need to do extra setup, further configure the VM, or install extra software outside the purview of Drupal VM.
+Drupal VM allows you to run extra shell scripts and ansible task files in the beginning and at the end of the provisioning process, in case you need to do extra setup, further configure the VM, or install extra software outside the purview of Drupal VM.
+
+## Shell scripts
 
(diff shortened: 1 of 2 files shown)


oxyc:
Not sure if omit works here, but let's see. Ref https://github.com/ansible/ansible/issues/13296#issuecomment-159768329

oxyc:
@thom8 heads up over at beetbox: https://travis-ci.org/geerlingguy/drupal-vm/jobs/130178132#L1110

thom8:
@oxyc don't think you need to set a default as it skips if the folder is empty or doesn't exist.

oxyc:
Ah just noticed you don't have the `when` clause, and rely on the variable existing. That's what caused the test failure here.

thom8:
The variable need to be set but it can it can point to an empty or non existing directory.
Otherwise, it could be a list of files similar to provisioning scripts.

geerlingguy:
@thom8 / @oxyc - Do we need to add a var to `example.config.yml` for this, or will `default(omit)` work okay? Tests seem to be passing...

oxyc:
Yes this work. I think we still have a few invalid `when` clauses for `with_item` though. Haven't looked properly but at least the cron tasks `when` clause doesnt do anything.

geerlingguy:
Ha, yeah, we can clean those up later. I often overapply the `when`s.

PR #638, opened 2016-05-17
Error during Drupal installation using Drush

subhojit777:


geerlingguy:
This is run the first time a site is built from a make file automatically: https://github.com/geerlingguy/drupal-vm/blob/master/provisioning/tasks/build-makefile.yml#L29-L34

If your site is built outside of Drupal VM, then you need to run the `composer install` step outside of Drupal VM as well. In the future, we'll support composer-only site build workflows, so this will become less of an issue. But for now, you can either have Drupal VM fully manage the initialization of the codebase, or do it outside of Drupal VM.

Issue #641, opened 2016-05-17
Pre-provisioning error


ccjjmartin:
## Background:
- I have backed up to a stable release 2.5.1 and everything is fine but on the latest commit in master (57fb88d) pre-provisioning fails with this error message:

```
TASK [Run configured pre-provision shell scripts.] *****************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: IndexError: list index out of range
fatal: [gw]: FAILED! => {"failed": true, "msg": "Unexpected failure during module execution.", "stdout": ""}
```

## Additional Information:
- My config.yml might have something out of date in it.  I can verify that it works with 2.5.1.  I am willing to try a different commit if that would help diagnose this problem.

oxyc:
Would you mind posting your `config.yml` and checking which ansible version you are using? I'd like to see if I can reproduce it.

oxyc:
Also to provide some more ansible debugging information could you create a `Vagrantfile.local` in the same directory as the regular `Vagrantfile` with this:

```rb
config.vm.provisioners[0].config.raw_arguments = ["-vvv"]
```

joelpittet:
I ran into something similar.  Add this
`pre_provision_scripts: []`
next to
`post_provision_scripts: []` 

in your config.yml

oxyc:
@joelpittet could you check which ansible version you have installed?

oxyc:
Nevermind, just realized you fixed it. I was able to reproduce it by removing the variable. Thanks!

I'll try to get a proper fix for it if the variable isn't defined.

ccjjmartin:
@joelpittet I can confirm that your fix works.

@oxyc I can confirm that your patch works. 👍

geerlingguy:
Fixed via merged PR above #643

PR #643, opened 2016-05-18
Issue #641: Fix so that optional variables are in fact optional

oxyc:
So turns omit `with_items` did not like `omit`. Unlike `with_fileglob` which had no issue. I guess it's because `with_fileglob` doesnt fail with 0 matches.

Falling back to an empty list works for `with_items` though.


ccjjmartin:
I can confirm this fixed my issue in #641 👍

PR #648, opened 2016-05-18
Make config.yml optional by always loading default configurations

oxyc:
Pushing to see what travis says. I'll start going through docs in the meantime.


oxyc:
I guess we should keep the `example.config.yml` and move some vagrant specific configurations there.

oxyc:
`vagrant_synced_folders` and consequently `drupal_core_path` come to mind as variables people should be aware of at least.

Additionally some variables quite often overridden are:
- `install_site`
- `build_makefile`

Anyone using Drupal VM for more than one site would also need to pay attention to:
- `vagrant_hostname` (note that this is used in the vhost definitions)
- `vagrant_machine_name`
- `vagrant_ip`

How should we do this? My thinking is that `example.config.yml` could still be useful to attract attention to some variables that users should consider.

oxyc:
In my opinion, these are the most frequently edited configurations and at minimum the ones I'd suggest users still going through before setting up their first site.

```yaml
---
# If you need to run multiple instances of Drupal VM, set a unique hostname,
# machine name, and IP address for each instance.
vagrant_hostname: drupalvm.dev
vagrant_machine_name: drupalvm
vagrant_ip: 192.168.88.88

# A list of synced folders, with the keys 'local_path', 'destination', and
# a 'type' of [nfs|rsync|smb] (leave empty for slow native shares). See
# http://docs.drupalvm.com/en/latest/extras/syncing-folders/ for more info.
vagrant_synced_folders:
  # The first synced folder will be used for the default Drupal installation, if
  # build_makefile: is 'true'.
  - local_path: ~/Sites/drupalvm
    type: nfs
    create: true

# Set this to false if you are using a different site deployment strategy and
# would like to configure 'vagrant_synced_folders' and 'apache_vhosts' manually.
build_makefile: true
drush_makefile_path: "{{ config_dir }}/drupal.make.yml"

# Set this to false if you don't need to install drupal (using the drupal_*
# settings below), but instead copy down a database (e.g. using drush sql-sync).
install_site: true

# Settings for building a Drupal site from a makefile (if 'build_makefile:'
# is 'true').
drupal_major_version: 8
drupal_core_path: "/var/www/drupalvm/drupal"
drupal_domain: "{{ vagrant_hostname }}"
drupal_site_name: "Drupal"
drupal_install_profile: standard
drupal_enable_modules: [ 'devel' ]
drupal_account_name: admin
drupal_account_pass: admin
```

oxyc:
Ready for review.

docs/other/overriding-configurations.md
```diff
@@ -1,4 +1,14 @@
-## Overriding variables in `config.yml` with a `local.config.yml`
+## Overriding Drupal VM's `default.config.yml` with `config.yml`
+
+If available, Drupal VM will load a `config.yml` where you can override any of the defaults set in `default.config.yml`. Commonly this is a copy of `default.config.yml` with the values tweaked to your own project. For an easier upgrade path you would only set the values you are actually overriding.
+
+```yaml
+vagrant_box: geerlingguy/centos7
+vagrant_hostname: my-custom-site.dev
```
oxyc: I verified that this particular example works. Ansible will dynamically evaluate the variable when it reaches the apache role and the correct hostname will be used in all places.

oxyc:
For reference, we decided to somewhat _hide_ the docs for this feature so users don't get confused and forget which variables are available.

danbohea:
+1 on this. Coming from Vlad config.yml feels overly verbose.

geerlingguy:
So... ready to merge? I like it so far, and it also fixes #682.

PR #652, opened 2016-05-20
Set roles install path to provisioning/roles. [ci skip]

joestewart:
The ansible.cfg file sets "roles_path = ./roles".

The vagrant provisioner installs the roles in provisioning/roles.


joestewart:
It appears that roles_path should be set to provisioning/roles in the ansible.cfg instead of a doc change.
Will close this PR and open one for ansible.cfg.

PR #653, opened 2016-05-20
Docs: Install php5-apcu. [ci skip]

joestewart:
Adding php5-apcu would be consistent with the default php role "php_enable_apc: true" value.


geerlingguy:
Thanks!

PR #654, opened 2016-05-21
Set roles install path to provisioning/roles.

joestewart:
Vagrant provisioner installs the roles in provisioning/roles.

### DIFF
--- a/ansible.cfg
+++ b/ansible.cfg
@@ -1,2 +1,2 @@
 [defaults]
-roles_path = ./roles
+roles_path = ./provisioning/roles


oxyc:
As of Drupal VM 3.0.0 there's actually no need to run the `ansible-galaxy install` command anymore. It runs automatically with `vagrant provision`. The docs section in your previous PR should actually be removed.

Was there a bug where it installed the roles in the wrong directory?

geerlingguy:
@oxyc - Yeah... do we even need the `ansible.cfg` file anymore, at all?

joestewart:
Removing ansible.cfg would be another option.  The current ./roles is wrong though.
IDK what's up with that test failure.

PR #663, opened 2016-05-25
Setting configure_local_drush_aliases to false now behaves as expected.

### LINKED ISSUE #662: Local drush aliases still created if configure_local_drush_aliases is set to false
PR incoming...

danbohea:
- Fixes #662 
- Looks like people are submitting PRs against `master`. Can someone confirm if that's what you're after?


oxyc:
The `configure_local_drush_aliases` actually refers to local as your host machine only. I personally wan't neither the local or the remote alias whenever I toggle this off so maybe it's worth either adding

```yaml
configure_remote_drush_aliases: "{{ configure_local_drush_aliases }}"
```

or just renaming it `configure_drush_aliases`.

geerlingguy:
`configure_remote_drush_aliases: "{{ configure_local_drush_aliases }}"` seems odd too...

The reason I called it `local` was because I was referring originally to the idea that all the aliases were being generated in my `local` environment (vs. a dev server, a prod server, etc.).

So local == `host` and `local` Drupal VM.

So to keep things less confusing, it's probably best to just rename the entire var to `configure_drush_aliases` instead. It is a breaking change, but maybe we could add a shim for it for now?

oxyc:
Sounds good. @danbohea I think you can use `set_fact` (used elsewhere in the playbook as well) to add a shim. Maybe with a comment stating it will eventually be removed.

danbohea:
Sure, will take a look when I get a spare minute.

danbohea:
PR updated. How's that?

Issue #661, opened 2016-05-25
Drush aliases not read from sites/all/drush/site-aliases/whatever.aliases.drushrc.php


danbohea:
Drush aliases stored in the project at sites/all/drush/site-aliases/whatever.aliases.drushrc.php should get automatically picked up by Drush but they don't. It doesn't appear to be an issue with Drush itself as this is working fine in Vlad with Drush 8.1.0.

Tested with Drupal VM 3.0.0, drush `master`, `8.x`, & `8.1.0`. Setting `configure_local_drush_aliases` to `true` or `false` makes no difference.

danbohea:
Just tested storing an aliases file at drush/site-aliases/whatever.aliases.drushrc.php too and that also doesn't work. From the sound of [this comment](https://github.com/geerlingguy/drupal-vm/issues/610#issuecomment-218811420) that did work very recently.

oxyc:
What directory are you running the drush command from? Are you using the default structure of:

```
.
├── Vagrantfile
├── config.yml
├── docroot
│   ├── index.php
│   ├── sites/all/drush/site-aliases/whatever.aliases.drushrc.php
```

danbohea:
Currently, the structure I'm using looks more like:

```
.
├── docroot
│   ├── index.php
│   ├── sites/all/drush/site-aliases/whatever.aliases.drushrc.php
│   └── other D7 project stuff...
└── drupal-vm
    ├── Vagrantfile
    ├── config.yml
    ├── provisioning
    └── other Drupal VM repo stuff...
```

- `local_path` is configured with the full path to that docroot location.
- `destination` has been left as the default value `/var/www/drupalvm`.
- I'm testing this by running `drush sa` once I've SSHd into the guest so the path would be /home/vagrant.
- I've just noticed that if I try to `cd` to /var/www/drupalvm the shell just hangs - not sure if related but doesn't seem right.

oxyc:
If you use site specific aliases they are only available within that project dir. So you can only see them if you `cd /var/www/drupalvm/docroot/`

danbohea:
That would explain it, I wasn't aware of that restriction and I'm likely too used to Vlad cd-ing me into the web root as soon as I ssh into the guest.

Now I just need to figure out why `cd /var/www/drupalvm/docroot/` hangs...

danbohea:
[...]

PR #665, opened 2016-05-25
Set ssh_home as default pwd for SSH.

thom8:
cd to `drupal_core_path` when SSHing to the VM. #661 

*note: this will be ignored if not set and fail gracefully if the path doesn't exist.

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -292,3 +292,4 @@ selenium_version: 2.46.0
 # Other configuration.
 dashboard_install_dir: /var/www/dashboard
 known_hosts_path: ~/.ssh/known_hosts
+#ssh_home: "{{ drupal_core_path }}"
--- a/provisioning/tasks/sshd.yml
+++ b/provisioning/tasks/sshd.yml
@@ -19,3 +19,13 @@
     mode: 0644
   become: no
   when: known_hosts_file.stat.exists
+
+- name: Set SSH home directory.
+  lineinfile:
+    dest: "/home/{{ drupalvm_user }}/.bashrc"
+    state: present
+    create: yes
+    regexp: "^SSH_HOME="
+    line: "SSH_HOME={{ ssh_home }} && [ -e $SSH_HOME ] && cd $SSH_HOME"
+  become: no
+  when: ssh_home is defined


example.config.yml
```diff
@@ -271,3 +271,4 @@ selenium_version: 2.46.0
 # Other configuration.
 dashboard_install_dir: /var/www/dashboard
 known_hosts_path: ~/.ssh/known_hosts
+ssh_home: "{{ drupal_core_path }}"
```
danbohea: Might the word "home" be confused with the path `/home/`? Perhaps this should be `ssh_dir` or `ssh_path` or `ssh_dest`?

oxyc:
Cool, I wasn't aware of that env variable! I'll add that our teams VM dotfiles 👍

example.config.yml
thom8: @danbohea - haha - I struggled with the naming of this too! + it's not a convention.

Couldn't find a standard name so since by default SSH connections drop you in the users home dir, this config for that SSH home dir.

opdavies:
+1 for this.

I just tested it out, and it works great.

geerlingguy:
I like it too. Need to poke and prod it a tiny bit more, but one thing—I think it would be best to _default_ to the normal .ssh dir (`~`, which is `~/.vagrant`), and allow override. For a guy like me, it would throw me off to be used to dropping into my home folder, and I'd rather not change that particular default—especially since there may be many synced drupal docroots (as is the case with many of my VMs).

thom8:
@geerlingguy how about we just don't set a default and comment it out in `example.config.yml`?

example.config.yml
```diff
@@ -271,3 +271,4 @@ selenium_version: 2.46.0
 # Other configuration.
 dashboard_install_dir: /var/www/dashboard
 known_hosts_path: ~/.ssh/known_hosts
+#ssh_home: "{{ drupal_core_path }}"
```
thom8: Commented out var in `example.config.yml`

geerlingguy: That works for me

geerlingguy:
D'oh, just noticed this PR has a conflict. Would you mind rebasing?

thom8:
@geerlingguy all done :)

geerlingguy:
Changed milestone. Small/non-breaking enough that it can go into a patch release instead of full version bump.

PR #667, opened 2016-05-26
Separate PPAs for PHP 5.5, 5.6 are now deprecated. 

thom8:
The PPA maintainer has now deprecated `php5` & `php5-5.6` -- https://launchpad.net/~ondrej

`***** The main PPA for PHP (5.5, 5.6, 7.0) with many PECL extensions *****`
https://launchpad.net/~ondrej/+archive/ubuntu/php

### DIFF
--- a/provisioning/tasks/init-debian.yml
+++ b/provisioning/tasks/init-debian.yml
@@ -20,17 +20,9 @@
   apt_repository: repo='ppa:ondrej/apache2'
   when: ansible_distribution_release == "precise" and ansible_distribution == "Ubuntu"
 
-- name: Add repository for PHP 5.5.
-  apt_repository: repo='ppa:ondrej/php5'
-  when: php_version == "5.5" and ansible_distribution == "Ubuntu" and ansible_distribution_version == "12.04"
-
-- name: Add repository for PHP 5.6.
-  apt_repository: repo='ppa:ondrej/php5-5.6'
-  when: php_version == "5.6" and ansible_distribution == "Ubuntu" and ansible_distribution_version != "16.04"
-
-- name: Add repository for PHP 7.0.
+- name: Add repository for PHP.
   apt_repository: repo='ppa:ondrej/php'
-  when: php_version == "7.0" and ansible_distribution == "Ubuntu" and ansible_distribution_version != "16.04"
+  when: ansible_distribution == "Ubuntu" and ansible_distribution_version != "16.04"
 
 - name: Define php_xhprof_html_dir.
   set_fact:


geerlingguy:
@thom8 hmm... so, how would a user be able to install 5.6 (instead of, say, 7.0) on Ubuntu 14.04?

geerlingguy:
D'oh... it seems the package names are flip flopping again—php5.6 would be the one for 5.6, php5.5, php7.0, and then I guess you have to use something like `php-apcu` (note the lack of a version number in there) to install whatever respective version is required?

This seems like it's going to have to be thoroughly documented, since package lists for PHP will have to change based on whether you want to use system PHP (5.4.x in 14.04, other versions in 12.04), or a PHP version from Ondrej's repo (without a number on some of the packages).

thom8:
There's been a lot of work done to tidy this up so you can install `php5.5` or `php5.6` or `php7.0` and package names are shared between versions.

https://launchpad.net/~ondrej/+archive/ubuntu/php/+index?field.series_filter=trusty

thom8:
Importantly, as these old PPAs have been deprecated they will unlikely receive security patches going forward.

geerlingguy:
@thom8 - True, but one thing that we'll have to do, maybe, is change all the defaults for Drupal VM and/or the `geerlingguy.php` role if not on Ubuntu 16.04. I generally default to using the system packages (as a principle, since I trust them slightly more than a community/individual-contributed package), then use other PPAs for newer versions if needed.

The annoying thing with this new layout is that system packages (for some silly historical reason :P) on Debian-based systems have the version number in them, so using Ondrej's packages now means that users have to have entirely different package sets depending on if they want system PHP (e.g. 5.4 on Ubuntu 14.04) or 5.5/5.6/7.0 on 14.04, or 7.0 on 16.04.

thom8:
@geerlingguy I guess it was never intended to have a single config support multiple versions but config management has quickly evolved.

> The annoying thing with this new layout is that system packages (for some silly historical reason :P) on Debian-based systems have the version number in them

This is the tidy up so as of 12.04/PHP 5.5 all package names should be the same apart from a couple of edge cases that have version specific dependencies. Packages with the version number should still work for BC reasons.

So while PHP 5.5 support ends soon, Ubuntu will support default versions of their releases (backporting security patches) for the life of the LTS therefore PHP 5.3 (12.04) and 5.5 (14.04) still have a bit of life left. 

http://php.net/supported-versions.php
https://wiki.ubuntu.com/LTS

thom8:
Ended up requiring a separate list of packages for each PHP version -- https://github.com/beetboxvm/beetbox/blob/master/provisioning/ansible/config/beetbox.config.yml#L276-L302

Which is set depending on the `php_version` -- https://github.com/beetboxvm/beetbox/blob/master/provisioning/ansible/roles/beetbox-php-ppa/tasks/main.yml#L21-L34

We also need to purge packages to support switching versions without rebuilding.

oxyc:
Loving that this is now possible

```yaml
php_conf_paths:
  - "/etc/php/{{ php_version }}/fpm"
  - "/etc/php/{{ php_version }}/apache2"
  - "/etc/php/{{ php_version }}/cli"
php_extension_conf_paths:
  - "/etc/php/{{ php_version }}/fpm/conf.d"
  - "/etc/php/{{ php_version }}/apache2/conf.d"
  - "/etc/php/{{ php_version }}/cli/conf.d"
php_fpm_daemon: "php{{ php_version }}-fpm"
php_fpm_conf_path: "/etc/php/{{ php_version }}/fpm"
php_mysql_package: "php{{ php_version }}-mysql"
php_memcached_package: "php{{ php_version }}-memcached"
```

geerlingguy:
This looks good to me, but we will need to add docs or automatic package switching (and remove the php-56 docs) before I can merge it in.

@oxyc - Can we just add that into Drupal VM's config if the OS is Ubuntu?

geerlingguy:
(Because we also need to support CentOS)

geerlingguy:
@thom8 - @oxyc took the progress from this PR and is building on it in #714, so I'm going to close this PR and focus on getting everything working there.

thom8:
Continued in #714

geerlingguy:
@thom8 Thanks for doing what I said I'd do but forgot to do 👍

PR #678, opened 2016-05-30
Update drush.md

richardrobinsonx:
Apparently Ansible is required to make this happen and it doesn't mention it in the documentation.


richardrobinsonx:
I'm not sure, but it looks like my intended comment was the commit message. Sorry about that :\

geerlingguy:
Quite true!

PR #685, opened 2016-06-01
Fixes incompatibility of dashboard and vagrant_ip set to 0.0.0.0

shumushin:
When using vagrant-auto_network plugin.

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -97,8 +97,7 @@ apache_vhosts:
     extra_parameters: |
           ProxyPassMatch ^/(.*\.php(/.*)?)$ "fcgi://127.0.0.1:9000{{ pimpmylog_install_dir }}"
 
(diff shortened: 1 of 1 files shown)


geerlingguy:
Unfortunately, this won't work with our automated tests, or in the main use case where someone brings up a VM and the hosts file isn't already updated (e.g. if the user doesn't have the autohosts/hostsmanager plugin installed).

Is there some other way you can get it to work? I haven't used auto_network much, so I don't know if there's a simpler/different way to make sure the generated IP address is used in the configuration.

shumushin:
Hmm, you are right, i'll look into it. For now i prefer not having auto generated IP address and having dashboard than the opposite.

geerlingguy:
This was hopefully fixed by #730.

Issue #668, opened 2016-05-26
Add support for Solr 5.x


mgp-novicell:
It would be really helpful if support for Solr was updated to include version 5

geerlingguy:
Upstream issue: https://github.com/geerlingguy/ansible-role-solr/issues/10

geerlingguy:
Did one better and also got Solr 6.x support added!

I will update the role requirements file for `3.0.0` soon, or I might finish adding automatic core provisioning and tag a new release of the upstream role before adding (therefore I'll leave this issue open until that's complete).

The alternative (if I use the current version of the role) is to just have the post-install script example also create a custom core to be used by the Search API Solr module.

geerlingguy:
3.1.0 includes automatic core setup/generation, as well as memory configuration, so I'll be moving to that version for Drupal VM 3.1.0, and also updating the default version in `default.config.yml` to the latest 5.x Solr release.

It looks like Solr 6 support is already in 8.x (https://www.drupal.org/node/2704383), but not yet in 7.x (https://www.drupal.org/node/2711685).

Once this is in, you could change solr versions by destroying and rebuilding the VM, switching the `solr_version` between `solr_version: "5.5.1"` and `solr_version: "6.0.1"`—or really any version between 3.5.x and the latest 6.x!

PR #693, opened 2016-06-03
Issue #668: Default to Solr 5.x.

geerlingguy:


geerlingguy:
Still needs an update to the post-provision example script so it works with Solr 5.x out of the box... testing changes now.

Issue #506, opened 2016-03-07
Use of Composer (instead of Drush Make) for building Drupal site


minorscience:
Just getting started with Drupal 8 and seeing that quite a few folks are using Composer to build. Any plans to include Composer support as a build option, instead of make?

geerlingguy:
I would like to do this soon, yes; just waiting for things to standardize a little more. I know there is a more-or-less 'official' workflow now (see below references), but there is still differing opinions on best practices for some of the sticky bits, like versioning, patches, extra libraries (non-PHP), etc. so I'm not sure if this will be a feature by Drupal VM 3.0.0 or not.

References:

  - [Using Composer in a Drupal project](https://www.drupal.org/node/2404989)
  - [Composer template for Drupal projects](https://github.com/drupal-composer/drupal-project)

The full workflow is currently supported, but not automated by, Drupal VM.

geerlingguy:
This is also something that I'm considering using as an excuse to switch things to my `geerlingguy.drupal` role instead of custom playbooks inside Drupal VM.

itsgreggmarshall:
I noticed this drupal.org issue for direct Composer support that might be worth following:

[https://www.drupal.org/node/2551607](https://www.drupal.org/node/2551607)

minorscience:
Good news, Jeff. Thanks for such a great resource. Slightly off-topic: wondering whether the addition of Drupal Console is also something worth exploring in the context of this project?

geerlingguy:
@minorscience - Drupal Console has been part of Drupal VM for some time :) — just make sure it's uncommented in `installed_extras`.

oxyc:
@geerlingguy but then it's only a documentation issue right? Would be cool to get the building blocks in for a 3.0.0 release as I assume switching to `geerlingguy.drupal` would require quite a few breaking changes. Docs for best practice could be added or tuned later in my opinion. I'd like to give `composer` a shot too as we've made the switch on all of our non-Drupal projects already :)

thom8:
:+1: for `geerlingguy.drupal` - would be good to collaborate if possible.

We're looking to push this [role](https://github.com/beetboxvm/beetbox/tree/master/provisioning/ansible/roles/beetbox-drupal) to galaxy when stable, which currently supports composer installations.

geerlingguy:
[...]

PR #694, opened 2016-06-03
Issue #506: Add support for composer based site builds

oxyc:
Issue #506. This is still WIP, just checking travis' opinion.

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -21,7 +21,8 @@ vagrant_public_ip: ""
 # http://docs.drupalvm.com/en/latest/extras/syncing-folders/ for more info.
 vagrant_synced_folders:
   # The first synced folder will be used for the default Drupal installation, if
-  # build_makefile: is 'true'. By default it's set to the drupal-vm folder.
+  # any of the build_* settings are 'true'. By default the folder is set to
+  # the drupal-vm folder.
   - local_path: .
     destination: /var/www/drupalvm
     type: nfs
@@ -34,17 +35,29 @@ vagrant_cpus: 2
 # The web server software to use. Can be either 'apache' or 'nginx'.
 drupalvm_webserver: apache
 
-# Set this to false if you are using a different site deployment strategy and
+# Set this to 'false' if you are using a different site deployment strategy and
 # would like to configure 'vagrant_synced_folders' and 'apache_vhosts' manually.
 build_makefile: true
 drush_makefile_path: "{{ config_dir }}/drupal.make.yml"
 
-# Set this to false if you don't need to install drupal (using the drupal_*
+# Set 'build_makefile' to 'false' and this to 'true' if you are using a
+# composer based site deployment strategy.
+build_composer: false
+drupal_composer_path: "{{ config_dir }}/drupal.composer.json"
+drupal_composer_install_dir: "/var/www/drupalvm/drupal"
+drupal_composer_dependencies: []
+
+# Set this to 'true' and 'build_makefile', 'build_composer' to 'false' if you
+# are using Composer's create-project as a site deployment strategy.
+build_composer_project: false
+drupal_composer_project_package: ""
+drupal_composer_project_options: "--prefer-dist --stability dev --no-interaction"
+
+# Set this to 'false' if you don't need to install drupal (using the drupal_*
 # settings below), but instead copy down a database (e.g. using drush sql-sync).
 install_site: true
 
-# Settings for building a Drupal site from a makefile (if 'build_makefile:'
-# is 'true').
+# Settings for installing a Drupal site (if 'install_site:' is 'true').
 drupal_major_version: 8
 drupal_core_path: "/var/www/drupalvm/drupal"
 drupal_domain: "{{ vagrant_hostname }}"
@@ -138,8 +151,8 @@ nginx_hosts:
 nginx_remove_default_vhost: true
 nginx_ppa_use: true
 
-# MySQL Databases and users. If build_makefile: is true, first database will
-# be used for the makefile-built site.
+# MySQL Databases and users. If 'install_site' is 'true', first database will
+# be used for the site.
 mysql_databases:
   - name: "{{ drupal_mysql_database }}"
     encoding: utf8
--- a/default.config.yml
+++ b/default.config.yml
@@ -240,8 +240,8 @@ composer_path: /usr/bin/composer
 composer_home_path: "/home/{{ drupalvm_user }}/.composer"
 composer_home_owner: "{{ drupalvm_user }}"
 composer_home_group: "{{ drupalvm_user }}"
-# composer_global_packages:
-#   - { name: phpunit/phpunit, release: '@stable' }
+composer_global_packages:
+  - { name: hirak/prestissimo, release: '^0.2' }
 
 # Run specified scripts before or after VM is provisioned. Path is relative to
 # the `provisioning/playbook.yml` file.
--- a/docs/deployment/composer.md
+++ b/docs/deployment/composer.md
@@ -1,0 +1,71 @@
+Drupal VM is configured to use [Drush make](drush-make.md) by default but supports building Drupal from a custom composer.json or creating a project from a composer package (`composer create-project`).
+
+## Using composer.json
+
+1. Copy `example.drupal.composer.json` to `drupal.composer.json` and modify it to your liking.
+2. Switch the build system by setting `build_makefile: false` and `build_composer: true` in your `config.yml`.
+3. Configure `drupal_core_path` to point to the webroot directory: `drupal_core_path: {{ drupal_composer_install_dir }}/docroot`
+
+```yaml
+build_makefile: false
+build_composer: true
+drupal_core_path: "{{ drupal_composer_install_dir }}/docroot"
+```
+
+_The file set in `drupal_composer_path` (which defaults to `drupal.composer.json`) will be copied from your host computer into the VM's `drupal_composer_install_dir` and renamed `composer.json`. If you already have a composer.json within that directory, you can set `drupal_composer_path: false`._
+
+## Using Composer when [Drupal VM is a composer dependency itself](../other/drupalvm-composer-dependency.md)
+
+In the scenario where you already have an existing `composer.json` in the root of your project, follow the usual steps for installing with a composer.json but instead of creating a `drupal.composer.json` file, disable the transfering of the file by setting `drupal_composer_path` to `false`, and change `drupal_composer_install_dir` to point to the the directory where it will be located. If `drupal_composer_path` is not truthy, Drupal VM assumes it already exists.
+
+```yaml
+build_makefile: false
+build_composer: true
+drupal_composer_path: false
+drupal_composer_install_dir: "/var/www/drupalvm"
+drupal_core_path: "{{ drupal_composer_install_dir }}/docroot"
+```
+
+## Creating a project from a composer package: `composer create-project`
+
+There's a couple of things you need to configure in your `config.yml`:
+
+- Switch the build system by setting `build_makefile: false`, `build_composer: false` and `build_composer_project: true`.
+- Configure the composer package in `drupal_composer_project_package`.
+- Additionally you can adjust the create-project CLI options in `drupal_composer_project_options` as well as add additional dependencies in `drupal_composer_dependencies`.
+- Ensure that the webroot configured in the composer package matches the one set in `drupal_core_path`.
+
+With [drupal-composer/drupal-project](https://github.com/drupal-composer/drupal-project) as an example your `config.yml` overrides would be:
+
+```yaml
+build_makefile: false
+build_composer: false
+build_composer_project: true
+
+drupal_composer_project_package: "drupal-composer/drupal-project:8.x-dev"
+# Added `--no-dev` to avoid installing development dependencies.
+drupal_composer_project_options: "--prefer-dist --stability dev --no-interaction --no-dev"
+drupal_composer_dependencies:
+  - "drupal/devel:8.*"
+
+drupal_core_path: "{{ drupal_composer_install_dir }}/web"
+```
+
+## Improving composer build performance
+
+Opting for composer based installs will most likely increase your VM's time to provision considerably.
+
+If you manage multiple VM's own your computer, you can use the [`vagrant-cachier` plugin](http://fgrehm.viewdocs.io/vagrant-cachier/) to share Composer's package cache across all VM's. The first build will be as slow as before but subsequent builds with the same `vagrant_box` (eg `geerlingguy/ubuntu1604`) will be much faster.
+
+1. Install the plugin on your host computer: `vagrant plugin install vagrant-cachier`
+2. Create a `Vagrantfile.local` next to your `config.yml` with the following:
+
+```rb
+config.cache.scope = :box
+config.cache.auto_detect = false
+config.cache.enable :generic, { :cache_dir => "/home/vagrant/.composer/cache"  }
+```
+
+_Note: Out of the box, sharing the Composer cache is not supported as the plugin requires PHP to be installed when the VM first boots up. This is why the generic cache bucket is used instead._
+
+_You can also use this plugin to share other package manager caches. For more information read the [documentation](http://fgrehm.viewdocs.io/vagrant-cachier/usage/)._
--- a/provisioning/tasks/build-composer.yml
+++ b/provisioning/tasks/build-composer.yml
@@ -1,0 +1,29 @@
+---
+- name: Ensure drupal_composer_install_dir directory exists.
+  file:
+    path: "{{ drupal_composer_install_dir }}"
+    state: directory
+    mode: 0775
+  become: no
+  when: drupal_composer_path and not drupal_site_exists
+
+- name: Copy composer.json into place.
+  copy:
+    src: "{{ drupal_composer_path }}"
+    dest: "{{ drupal_composer_install_dir }}/composer.json"
+  when: drupal_composer_path and not drupal_site_exists
+  become: no
+
+- name: Run composer install.
+  command: >
+    composer install
+    chdir={{ drupal_composer_install_dir }}
+  when: not drupal_site_exists
+  become: no
+
+- name: Install dependencies with composer require.
+  command: >
+    {{ composer_path }} require {{ item }}
+    chdir={{ drupal_composer_install_dir }}
+  with_items: "{{ drupal_composer_dependencies|default([]) }}"
+  become: no
--- a/provisioning/tasks/build-makefile.yml
+++ b/provisioning/tasks/build-makefile.yml
@@ -3,7 +3,7 @@
   copy:
     src: "{{ drush_makefile_path }}"
     dest: /tmp/drupal.make.yml
-  when: drupal_site.stat.exists == false
+  when: not drupal_site_exists
 
 - name: Ensure drupal_core_path directory exists.
   file:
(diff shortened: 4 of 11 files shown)


default.config.yml
```diff
@@ -34,19 +35,28 @@ vagrant_cpus: 2
...
+# Set this to 'false' if you don't need to install drupal (using the drupal_*
 # settings below), but instead copy down a database (e.g. using drush sql-sync).
 install_site: true
 
-# Settings for building a Drupal site from a makefile (if 'build_makefile:'
-# is 'true').
+# Settings for installing a Drupal site (if 'install_site:' is 'true').
 drupal_major_version: 8
-drupal_core_path: "/var/www/drupalvm/drupal"
+drupal_core_path: "{{ drupal_composer_path }}/web"
```
oxyc: Out of the box this _should_ still work for makefiles. It's a bit weird having the otherwise empty `drupal/` directory though.

oxyc: As everyone has their `config.yml` configured already, it's technically not a breaking change though.

oxyc:
@geerlingguy if you like this approach and travis passes, tell me and I'll start working on updating the docs. (I also need to manually test that everything works with a custom `composer.json`).

default.config.yml
```diff
@@ -34,19 +35,28 @@ vagrant_cpus: 2
...
 drush_makefile_path: "{{ config_dir }}/drupal.make.yml"
 
-# Set this to false if you don't need to install drupal (using the drupal_*
+# Set this to 'false' if you are using a different site deployment strategy and
+# would like to configure 'vagrant_synced_folders' and 'apache_vhosts' manually.
+# If there's a composer.json file present in 'drupal_composer_path', it will be
+# used instead of the 'drupal_composer_package'.
+build_composer: true
+drupal_composer_path: "/var/www/drupalvm/drupal"
+drupal_composer_package: "drupal-composer/drupal-project:8.x-dev"
```
geerlingguy: I was thinking we could have an `example.composer.json` and build from that, just like using `drupal.make.yml`; this approach isn't necessarily wrong, but I feel like it would be more intuitive to have a composer.json and build from that rather than building the composer install using a CLI command.

default.config.yml
geerlingguy: I generally see projects use `docroot` rather than `web` for the actual drupal directory... could we do that if we use `composer.json` and install it in the proper directory?

default.config.yml
```diff
@@ -138,8 +148,8 @@ nginx_hosts:
 nginx_remove_default_vhost: true
 nginx_ppa_use: true
 
-# MySQL Databases and users. If build_makefile: is true, first database will
-# be used for the makefile-built site.
+# MySQL Databases and users. If 'install_site' is 'true', first database will
```
geerlingguy: Good catch :)

provisioning/tasks/build-composer.yml
```diff
@@ -0,0 +1,27 @@
+---
+- name: Check if composer.json already exists.
+  stat: "path={{ drupal_composer_path }}/composer.json"
+  register: drupal_composer
+
+- name: Generate Drupal site with composer.
+  command: >
+    {{ composer_path }} create-project
```
geerlingguy: See notes from earlier regarding an `example.composer.json` file... what are your thoughts?

default.config.yml
oxyc: You might be correct, `drupal-project` adds some complexity. However, with this approach we don't need to maintain a `composer.json` at all, and it also removed some steps from the quick start guide.

But if you prefer, I can add a basic `composer.json` and maybe add the `create-project` command installation as an option. `build_composer_project: true` or something.

joestewart:
Some context to drupal-project and web/ vs docroot/ - https://github.com/drupal-composer/drupal-project/issues/64

There is a little work being done there to make specifying the docroot easier.

oxyc:
In case we want to support `create-project` based site builds I added it to a separate commit.

All 3 setups are being tested on travis. Note: docs are entirely out of sync.

oxyc:
The CI fail is unrelated, I'm assuming it works but will test it locally as well.

@geerlingguy want me to keep support for `composer create-project` or should we go simple and just support `composer.json`?

oxyc:
Erm just realized it's a bit chaotic with `example.composer.json` next to Drupal VM's own `composer.json` hehe... I'm currently using the same setup as with makefiles (copying it over to the VM with ansible so it works when provisioning production environments as well). I'm also looking for it in `config_dir` to be consistent with the makefile.

I guess we need to commit the `docroot/` directory and place it as well as look for it there.

oxyc:
I don't really like placing the `example.composer.json` in a separate subdir like that, but don't see any other option. 

@geerlingguy thoughts? Personally I prefer the first approach more to be honest. Just default to `drupal-project` (I'm biased as I use that project though), and let the file be generated with optionally required dependencies in `drupal_composer_dependencies`. On top of that add support for skipping the project creationg, and instead specifying a custom `composer.json`, making the user aware of Drupal VM's own `composer.json`.

geerlingguy:
@oxyc - Let's do the first and use drupal-project. We can pivot at some point if needed, and it makes more sense to do that and simplify overall setup, I think. People who are deep enough into the composer workflow should/will use Drupal VM as a dependency anyways, so no need to support only composer.json builds out of the box.

geerlingguy:
Sorry for creating the bit of rework, but I'm coming around to what you originally proposed. Will need to poke and prod locally in a bit.

provisioning/tasks/build-composer.yml
```diff
@@ -0,0 +1,33 @@
+---
+- name: Copy composer.json into place.
+  copy:
+    src: "{{ drupal_composer_json }}"
+    dest: "{{ drupal_composer_install_dir }}/composer.json"
+  when: drupal_composer_json is defined and not drupal_site.stat.exists
```
oxyc: This isn't needed for the usual setups (BLT or projects where Drupal VM is in vendor), but I added it as it could be useful in some scenarios. Eg. provisioning production environments.

oxyc:
I'll get started on rewriting the docs and running some more tests in a bit.

geerlingguy:
Testing locally.

geerlingguy:
@oxyc - CentOS 7 build failed with:

```
TASK [Check if site is already installed.] *************************************
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: OSError: [Errno 2] No such file or directory: '/var/www/drupalvm/drupal/web'
```

geerlingguy:
We'll see how test run 777 goes.

geerlingguy:
@oxyc is the `Generate Drupal project with composer package.` supposed to take forever? My CPU's been spinning for a while at 200%, for I think 10 minutes or so. Is there a way to get that process to speed up without killing my CPU, or is this just the way it is with drupal-project?

provisioning/tasks/build-makefile.yml
```diff
@@ -5,15 +5,6 @@
     dest: /tmp/drupal.make.yml
   when: not drupal_site.stat.exists
 
-- name: Ensure drupal_core_path directory exists.
-  file:
-    path: "{{ drupal_core_path }}"
-    state: directory
-    recurse: yes
```
oxyc: Hmm, what was the reason behind this being recursive?

oxyc:
> @oxyc is the Generate Drupal project with composer package. supposed to take forever? My CPU's been spinning for a while at 200%, for I think 10 minutes or so. Is there a way to get that process to speed up without killing my CPU, or is this just the way it is with drupal-project?

Hmm not sure... I'll see if I can reach them on twitter.

geerlingguy:
See also: https://github.com/drupal-composer/drupal-project/issues/137#issuecomment-221791682 — but yeah, this is kind of a deal breaker imo, as it's been about 20 minutes (still killing my CPU currently), and I'm on a 1.7 Ghz i7 on VMware with an SSD, and a 100 Mbps internet connection.

I can't even begin to imagine how many hours it will take a poor soul stuck on an old PC with an Atom processor and a 2 Mbps internet connection :P

geerlingguy:
Everything installed successfully, so there's that :)

oxyc:
[This run](https://travis-ci.org/geerlingguy/drupal-vm/builds/135106886) had all 3 installs:

- With a bare minimum `composer.json` – 12min
- With the old Makefile – 8min
- With drupal-project – 16min

Yeah that's really bad. I'm on a 1.1ghz Macbook and it only takes a few minutes for me though. Haven't even noticed it's slower than the makefiles.

geerlingguy:
Maybe we could still default to the makefile but support composer.json, then... I'm going to test again on a rebuild and see if it's any quicker.

geerlingguy:
22 minutes on first build for me. Will update and add 2nd build time in a few.

geerlingguy:
@oxyc could we use https://github.com/hirak/prestissimo somehow?

geerlingguy:
11 minutes on second build.

geerlingguy:
Updated my stats—6 min vs 22 minutes is a definite deal-breaker :(

We could default to makefile since it's waaaay faster, and then allow people to use composer if they'd like. Unless we can figure out a way to get it at least less than 2x slower than makefiles. Maybe prestissimo will work.

oxyc:
@geerlingguy what are your stats if you use a bare minimum `composer.json`? Such as the one in `tests/composer.json`.

geerlingguy:
Testing that now.

geerlingguy:
I'm getting:

```
TASK [Check if site is already installed.] *************************************
fatal: [drupalvm]: FAILED! => {"changed": false, "failed": true, "failed_when_result": true, "module_stderr": "", "module_stdout": "Traceback (most recent call last):\r\n  File \"/tmp/ansible_jH9QUN/ansible_module_command.py\", line 250, in <module>\r\n    main()\r\n  File \"/tmp/ansible_jH9QUN/ansible_module_command.py\", line 186, in main\r\n    os.chdir(chdir)\r\nOSError: [Errno 2] No such file or directory: '/var/www/drupalvm/drupal/web'\r\n", "msg": "MODULE FAILURE", "parsed": false}
```

when I try composer.json install.

oxyc:
Was it this https://github.com/geerlingguy/drupal-vm/pull/694/commits/36bff6d0f08206dea5d6bc201ab251aa55bf6d05?

oxyc:
Running `composer create-project drupal-composer/drupal-project:8.x-dev foo --stability dev --no-progress --profile` with various setups locally:

Options                        | Cache | No cache | 
------------------------------|-----------|---------------
Default                         | 142s  | not sure I want to
--no-dev                       | 79s    | 620s (WOAH)
Prestissimo                  | 93s    |
Prestissimo --no-dev    | 67s  |

geerlingguy:
lol @ WOAH

geerlingguy:
`Prestissimo --no-dev` - how is it with cache?

geerlingguy:
Now I'm getting

```
fatal: [drupalvm]: FAILED! => {"changed": true, "cmd": ["/usr/local/bin/drush", "site-install", "standard", "-y", "--site-name=Drupal", "--account-name=admin", "--account-pass=admin", "--db-url=mysql://drupal:drupal@localhost/drupal"], "delta": "0:00:00.502764", "end": "2016-06-03 22:04:41.987479", "failed": true, "rc": 1, "start": "2016-06-03 22:04:41.484715", "stderr": " [error]  Command site-install needs a higher bootstrap level to run - you will need to invoke drush from a more functional Drupal environment to run this command. \n [error]  The drush command 'site-install standard' could not be executed. ", "stdout": "", "stdout_lines": [], "warnings": []}
```

oxyc:
> Prestissimo --no-dev - how is it without cache?

I'll keep updating the table. `Prestissimo --no-dev` with cache was 67s though. Without cache I'm guessing it's going to be close to that WOAH.

> Now I'm getting...

Hm this one was passing in [travis](https://travis-ci.org/geerlingguy/drupal-vm/jobs/135142352#L2142). Maybe I missed something. What `config.yml` are you using? Provisioning VM or remote server?

geerlingguy:
I had the wrong `drupal_core_path` :(

oxyc:
I can't continue more for now. I'll have a look again tonight but like you said, maybe it's best just to default to makefile. I think it'll still be slow as long as we use composer for checking out drupal core.

If we go with makefiles, maybe remove all the defaults regarding `drupal-project` and revert the webroot to `drupalvm/drupal/docroot`? I can add a section in the docs about installing using composers `create-project` and have `drupal-project` as an example.

geerlingguy:
Yeah, let's do that. Unfortunately, I don't want to default to composer until we find a way to make the build time reasonable.

oxyc:
I agree. I'm currently running the basic composer install with prestissimo, without cache and it's been going at it for like 10min already.

geerlingguy:
Grr... this thread has some great data in it though—we can do a follow-up issue to look at how to make things faster. Maybe there's a trick someone has up their sleeve to supercharge the install.

oxyc:
We could look for `composer.json` in the Vagrantfile and issue it there instead :D That way the users cache would be re-used across all projects.

oxyc:
A more maintainable solution would be to add a sync tasks that copies the users composer cache to the vm. As we do with `known_hosts`.

thom8:
@oxyc you could try syncing the host `~/.composer/cache` to the VM but no help when you don't have an existing cache.

```
vagrant_synced_folders:
  - local_path: ~/.composer/cache
    destination: /root/.composer/cache
    type: nfs
    create: true
```

I assume the local cache is why it's not such a major issue for most ppl not to mention pulling projects with such a huge history like drupal..

Shallow clones would be a big help! -- https://github.com/composer/composer/pull/4961

joestewart:
For comparison, try vagrant-cachier.  It supports composer - http://fgrehm.viewdocs.io/vagrant-cachier/buckets/composer/

oxyc:
> @oxyc you could try syncing the host ~/.composer/cache to the VM but no help when you don't have an existing cache.

@thom8 yeah we should probably add a section on how to improve performance. That would be one way, together with `vagrant-cachier` as @joestewart suggested. Maybe that can be a separate docs task though.

---

A lot of back and forth here. Now I'm back to drush makefile being the default, and having `build_composer` and `build_composer_project` as separate tasks. `build_composer_project` is not tested in travis but I'm testing it locally now.

oxyc:
Tested locally:
- [x] default makefile approach
- [x] drupal.composer.json
- [x] drupal-project
- [ ] `drupal_composer_path: false` + `drupal_composer_install_dir: "/var/www/drupalvm"`

default.config.yml
```diff
@@ -34,19 +35,31 @@ vagrant_cpus: 2
...
+# Set this to 'false' if you are using a different site deployment strategy and
 # would like to configure 'vagrant_synced_folders' and 'apache_vhosts' manually.
 build_makefile: true
 drush_makefile_path: "{{ config_dir }}/drupal.make.yml"
 
-# Set this to false if you don't need to install drupal (using the drupal_*
+# Set 'build_makefile' to 'false' and this to 'true' if you are using a
+# composer based site deployment strategy.
+build_composer: false
+drupal_composer_path: "{{ config_dir }}/drupal.composer.json"
```
oxyc: Want me to include a `example.drupal.composer.json` even if we don't fully stand behind composers based installs until the build duration is improved?

default.config.yml
```diff
@@ -34,19 +35,31 @@ vagrant_cpus: 2
...
+# Set this to 'false' if you don't need to install drupal (using the drupal_*
 # settings below), but instead copy down a database (e.g. using drush sql-sync).
 install_site: true
 
-# Settings for building a Drupal site from a makefile (if 'build_makefile:'
-# is 'true').
+# Settings for installing a Drupal site (if 'install_site:' is 'true').
 drupal_major_version: 8
-drupal_core_path: "/var/www/drupalvm/drupal"
+drupal_core_path: "{{ drupal_composer_install_dir }}/docroot"
```
oxyc: I can also keep the old behaviour as makefiles are still the default. Just need to mention that this needs to be changed in the composer docs.

oxyc: I can also keep the old behaviour as makefiles are still the default. Just need to mention that this needs to be changed in the composer docs.

oxyc:
`vagrant-cachier` doesn't work as it [requires PHP to exist](https://github.com/fgrehm/vagrant-cachier/blob/da0eff6a71c8e292ccf5b938402bac7f8c37561c/lib/vagrant-cachier/cap/linux/composer_path.rb#L9) before booting the VM.

oxyc:
Not sure why, but `vagrant-cachier` isn't caching anything for me...

default.config.yml
geerlingguy: Yes; it will help people who might not be familiar with the workflow to test it out if they'd like.

default.config.yml
geerlingguy: Yeah, let's stick to the old for now, since we're not defaulting to composer.

geerlingguy:
I think if we get in the other two changes mentioned above, I'm happy to merge this and get the 3.1.0 release tagged. Hopefully we can find a way to get performance to the point where composer is the default by .2 or .3!

default.config.yml
```diff
@@ -34,19 +35,31 @@ vagrant_cpus: 2
...
 # would like to configure 'vagrant_synced_folders' and 'apache_vhosts' manually.
 build_makefile: true
 drush_makefile_path: "{{ config_dir }}/drupal.make.yml"
 
-# Set this to false if you don't need to install drupal (using the drupal_*
+# Set 'build_makefile' to 'false' and this to 'true' if you are using a
+# composer based site deployment strategy.
+build_composer: false
+drupal_composer_path: "{{ config_dir }}/drupal.composer.json"
+drupal_composer_install_dir: "/var/www/drupalvm/drupal"
```
oxyc: Guess in the docs this could be `/var/www/drupalvm` with the `drupal.composer.json` setup as well. Using `composer create-project` it needs to be in a subdir though.

default.config.yml
```diff
@@ -34,19 +35,31 @@ vagrant_cpus: 2
...
+# Set this to 'false' if you don't need to install drupal (using the drupal_*
 # settings below), but instead copy down a database (e.g. using drush sql-sync).
 install_site: true
 
-# Settings for building a Drupal site from a makefile (if 'build_makefile:'
-# is 'true').
+# Settings for installing a Drupal site (if 'install_site:' is 'true').
 drupal_major_version: 8
-drupal_core_path: "/var/www/drupalvm/drupal"
+drupal_core_path: "{{ drupal_composer_install_dir }}"
```
geerlingguy: Let's leave this as the default `/var/www/drupalvm/drupal` (the `composer` in the var might confuse people who aren't using composer, which at this point is likely the majority of users).

docs/deployment/composer.md
```diff
@@ -0,0 +1,71 @@
...
+drupal_composer_project_package: "drupal-composer/drupal-project:8.x-dev"
+# Added `--no-dev` to avoid installing development dependencies.
+drupal_composer_project_options: "--prefer-dist --stability dev --no-interaction --no-dev"
+drupal_composer_dependencies:
+  - "drupal/devel:8.*"
+
+drupal_core_path: "{{ drupal_composer_install_dir }}/web"
+```
+
+## Improving composer build performance
```
geerlingguy: This is extremely helpful. Hopefully we can get more visibility/testing on this so we can improve build performance across the board. One thing I hate about the current D8 project I'm working on is how, every time I add a module to the `composer.json` and run `composer update`, I have to wait about 5-10 minutes before I can be productive again...

geerlingguy:
I'm ready to merge once that conflict is resolved. A long road to this point, but it's well worth it. This will be an awesome improvement!

oxyc:
Whops I somehow managed to get in the commit from my other PR. Getting tired....

Want me to squash these some more? A lot of back and forth going on...

geerlingguy:
If it's quick, you can... otherwise I can live with the history. It's not all bad to have some of the pivots as part of the history, though it makes my release note creation a little more fun :D

geerlingguy:
Looks good. Once test is green, merge button will be pressed 👍

PR #284, opened 2015-10-25
Solve nfs permissions issues

vladdancer:
Include vagrant-bindfs plugin.
When I try to use drupal-vm on MacOS X El Capitan, I get errors with folder perms using nfs. So I've found nearly a solution in this issue https://www.drupal.org/node/2485869
1 vagrant-ssh-with-wrong-perms
And after I've changed a little bit a Vagrantfile shared folder was mounted with vagrant:vagrant ownership and nginx stopped return 404 error.

Also I've found this thread https://groups.google.com/forum/#!topic/vagrant-up/qXXJ-AQuKQM but haven't yet tested.

PR #698, opened 2016-06-05
Issue #284: Add docs about vagrant-bindfs

oxyc:


oxyc:
Tested it. Before `/var/www/drupalvm` was owned by `501:dialout`. Following the docs they were owned by `vagrant:www-data`.

geerlingguy:
Great, gladly merged in.

cdmo:
Following the instructions in the documentation only works for me after the initial `vagrant up`. Oh and it requires a `vagrant reload` by the way in my experience.

However, every time I try to add it in before the initial `vagrant up` it fails. See below:

```
==> drupalvm: Mounting NFS shared folders...
==> drupalvm: Bindfs seems to not be installed on the virtual machine, installing now
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

wget http://bindfs.org/downloads/bindfs-1.13.0.tar.gz -O bindfs.tar.gz

Stdout from the command:



Stderr from the command:

--2016-07-01 17:41:09--  http://bindfs.org/downloads/bindfs-1.13.0.tar.gz
Resolving bindfs.org (bindfs.org)... 127.0.53.53
Connecting to bindfs.org (bindfs.org)|127.0.53.53|:80... failed: Connection refused.
```

Not sure it makes a difference but I'm working with CentOS 7 on the latest version of all software, including DrupalVM.

cdmo:
Guess it might be a coincidence in terms of when I attempt to load bindfs, during a `reload` or the initial `up`. Right now http://bindfs.org/downloads/bindfs-1.13.0.tar.gz is indeed 404ing, [bindfs.org](bindfs.org) is down. Does this site go down intermittently?

geerlingguy:
@cdmo - See upstream issue https://github.com/gael-ian/vagrant-bindfs/issues/46 — I don't use bindfs myself, so I don't know how the long-term reliability has been with it :)

gael-ian:
I confirm its nothing more than a coincidence. bindfs.org has some DNS misconfiguration since Friday but shouldn't be unavailable for long.

PR #701, opened 2016-06-05
Use composer project based site build as the default

oxyc:
@geerlingguy would you mind testing this to see how the performance has improved since https://github.com/geerlingguy/drupal-vm/pull/694#issuecomment-223688740?

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -37,7 +37,7 @@ drupalvm_webserver: apache
 
 # Set this to 'false' if you are using a different site deployment strategy and
 # would like to configure 'vagrant_synced_folders' and 'apache_vhosts' manually.
-build_makefile: true
+build_makefile: false
 drush_makefile_path: "{{ config_dir }}/drupal.make.yml"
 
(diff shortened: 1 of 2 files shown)


oxyc:
If this turns out to be fast enough, should we use `build_composer_project` as the default so we can remove a bunch of steps from the quick start guide. Or should we just use `build_composer` and change everything to reference the `example.drupal.composer.json` instead of `example.drupal.make.yml`?

geerlingguy:
Yes! If the performance is close, let's use the simplest setup method.

geerlingguy:
I'll try to run tests tonight, it's crepe night, so it'll have to wait a bit.

geerlingguy:
|| Test type || Time |
| ----- | ----- |
| Makefile default config | 00:06:01 |
| Composer (this PR) | TODO |

geerlingguy:
Testing is good, I'm going to include a default cachier config for #702 once I get a little more testing completed.

PR #706, opened 2016-06-06
Docs for 3.1.0

oxyc:


geerlingguy:
Sheesh, I missed quite a bit there!

oxyc:
Almost too much docs nowadays hehe

geerlingguy:
Yeah, it's that situation where if there are too few people complain. Too many they don't read. Main thing for me is they need to be google-able and reference-able.

Also, imo, if we hit the 'there are too many docs' issue too frequently, then that's a sign to me as a project maintainer that I have too many features to support. Maybe time to start cutting off some things that are very rarely used or only cover edge cases (e.g. ancient Ubuntu 12.04/CentOS 6 support :P).

Issue #699, opened 2016-06-05
Default to Composer-based site building instead of Drush Make


geerlingguy:
Background:

  - #506 and PR #694 
  - http://docs.drupalvm.com/en/latest/deployment/composer/

Drupal VM currently supports building codebases using Composer or Drush make files. We were going to default to building with `drupal-project` and/or a `composer.json` file instead of the example `drush.make.yml` file, but some performance testing showed abysmal regressions in terms of the initial site build—and even fully cached site builds using Composer were 2x or more slower than Drush make file builds.

I'd like to move to using Composer as the default, as that's kind of where the puck is going... but before that can happen, we need to continue poking and prodding at Composer and all the default setup to make sure the Composer install is not a horrendously slow user experience, especially the first time (uncached).

@oxyc did some awesome investigative work and found that [there are a few things we can do](http://docs.drupalvm.com/en/latest/deployment/composer/#improving-composer-build-performance) out of the gates to speed things up, like:

  - Configure `vagrant-cachier` correctly to cache everything from Composer outside the VM so it persists across builds.
  - Include `prestissimo` to speed up `composer` commands by parallelizing downloads
  - Use `--prefer-dist`

But I'm hoping there are some other optimizations we can make to get the composer installation time down quite a bit. I know @thom8 suggested https://github.com/composer/composer/pull/4961 might help by using shallow Git clones, but we definitely need hard data (benchmarks) to back up theories.

oxyc:
Also worth checking how much slowdown `"minimum-stability": "dev"` causes. [Ref](https://karsten.dambekalns.de/blog/stability-settings-in-composer.html)

jmolivas:
did you turned off x-debug from the CLI?

fubarhouse:
I'm curious about build times in comparrison to the drupal-console and drush make methods, and the possibility of optionally switching xdebug on/off during this process...

oxyc:
xdebug is by default disabled so the tests from earlier ran without it. I added a few tests in PR #700 to get some basic comparison readings with different configs.

oxyc:
Test                                  | [802][1] | [803][2] | [804][3]  | [805][4]
----------------------------------|------------|-----------|-------------|--------
Makefile                            | 9:12      | 9:16     | 9:11        | 9:03
Composer                         |              |             | 11:59      | 11:56
Composer + XDebug        | 13:19    | 13:11   | 14:17      | 13:11
[...]

PR #707, opened 2016-06-06
Use packages.drupal.org composer repository in example

oxyc:
Followup #699. Also opened a PR at drupal-project. https://github.com/drupal-composer/drupal-project/pull/159


geerlingguy:
Platform.sh [just announced](https://platform.sh/2016/05/drupal-8-starter-kit-now-with-php-7-and-composer/) they're defaulting to the new repository. I think it's time we do the same :)

Issue #709, opened 2016-06-06
The following settings shouldn't exist: galaxy_role_file


frob:
This is a posterity issue.

I downloaded the latest version of drupalvm, configured the config.yml file and ran vagrant up.

```bash
14:38 $ vagrant up
Bringing machine 'drupalvm' up with 'virtualbox' provider...
There are errors in the configuration of this machine. Please fix
the following errors and try again:

ansible provisioner:
* The following settings shouldn't exist: galaxy_role_file

14:58 $ vagrant --version
Vagrant 1.7.4
```

frob:
The fix for this is to install the latest vagrant 1.8.1

PR #710, opened 2016-06-06
Issue #709: Add strict vagrant version requirement for easier debugging

oxyc:
I think we've had a few of these issues as well and this makes it super clear

> This Vagrant environment has specified that it requires the Vagrant
version to satisfy the following version requirements:

> \>= 1.8.1

> You are running Vagrant 1.x.x, which does not satisfy
these requirements. Please change your Vagrant version or update
the Vagrantfile to allow this Vagrant version. However, be warned
that if the Vagrantfile has specified another version, it probably has
good reason to do so, and changing that may cause the environment to
not function properly.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,6 +1,7 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 VAGRANTFILE_API_VERSION = '2'
+Vagrant.require_version '>= 1.8.1'
 
 # Absolute paths on the host machine.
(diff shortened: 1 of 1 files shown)


geerlingguy:
Will help for those using an older version, for sure. Thanks!

joestewart:
This will be nice in the future and vagrant cooperates.  This didn't seem to be the case with "The following settings shouldn't exist: galaxy_role_file".  I got frustrated trying to add this check for that situation.  Thanks.

Issue #449, opened 2016-02-18
Make drush optional


oxyc:
Would you be up for making the drush installation/configuration optional? My reasoning is that at work, we use Drupal VM on projects other than Drupal. Mainly Wordpress, and it fits perfectly. Drush is really the only non-optional installation step that's not useful for general php projects. https://github.com/generoi/bedrock/blob/genero/config/drupal-vm.config.yml

If we end up merging https://github.com/oxyc/drupal-vm/commit/8265f232e457d5ab4282c49777d0542a313d84fa that `vars/main.yml` could be a good spot to place the default variable if we want to "hide" the functionality.

geerlingguy:
I think this is a good idea... but I'd like to fix up my `geerlingguy.drupal` role so it can be used (maybe within a loop even) to do Drupal installs instead of all the custom stuff in task includes in Drupal VM.

I'm thinking if we can define a list of `drupal_sites` (default to just the one), and that loop can be passed to the included role that allows install by makefile or composer, that kind of setup would be by far the nicest/most streamlined.

There are a few other architectural changes that would need to be made, but it could make Drupal VM a lot more flexible—and offer a great solution to other users who just want a good `geerlingguy.drupal` role to use.

itsgreggmarshall:
I like the idea of having multiple Drupal sites, I was going to open an issue to request it (since I support about 40 Acquia Cloud sites, something DD2 does do well).

oxyc:
Related to supporting multiple site installations #181, #33, #10.

oxyc:
One issue that will come up once we do this is that the you would need to use arrays to reference the drupal variables in `apache_vhosts` and `nginx_hosts`.

```yaml
apache_vhosts:
  - servername: "{{ drupal_sites[0].drupal_domain }}"
    documentroot: "{{ drupal_sites[0].drupal_core_path }}"
    extra_parameters: |
          ProxyPassMatch ^/(.*\.php(/.*)?)$ "fcgi://127.0.0.1:9000{{ drupal_sites[0].drupal_core_path }}"
```

This will break https://github.com/geerlingguy/drupal-vm/blob/master/Vagrantfile#L35:L40 as it does not support arrays at the moment.

Maybe we can somehow use `with_items` for the webserver roles as well?

fubarhouse:
Though I've got my memory/cpus bumped, I thought it might be useful to note that currently my VM is running 13 websites (D7), and I've seen it run at ~25 previously.

[...]

PR #711, opened 2016-06-06
Make drush optional

oxyc:
Checking how much faster #449 would make the build. Would obviously need a lot more work. Also makefile tests will fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ env:
       MACHINE_NAME: drupalvm
       IP: 192.168.88.88
       DRUPALVM_DIR: /var/www/drupalvm
+      DRUSH_BIN: "${DRUPALVM_DIR}/drupal/vendor/drush/drush/drush"
       TEST_INSTALLED_EXTRAS: true
   matrix:
     # Defaults - Ubuntu 16.04.
@@ -24,6 +25,7 @@ env:
       run_opts: "--privileged"
       local_config: tests/ubuntu-16-nginx.config.yml
       config_dir: /var/www/drupalvm/config
+      DRUSH_BIN: "/usr/local/bin/drush"
     # PostgreSQL - Ubuntu 16.04.
     - type: postgresql
       distro: ubuntu1604
@@ -131,7 +133,7 @@ script:
     || (echo 'Dashboard install fail' && exit 1)
 
   - >
-    docker exec ${container_id} drush @${MACHINE_NAME}.${HOSTNAME} status
+    docker exec ${container_id} ${DRUSH_BIN} @${MACHINE_NAME}.${HOSTNAME} status
     | grep -q 'Drupal bootstrap.*Successful'
     && (echo 'Drush install pass' && exit 0)
     || (echo 'Drush install fail' && exit 1)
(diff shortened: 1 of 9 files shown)


oxyc:
I most likely won't have time to help polish out `3.2.0` and I'm guessing this will be a PR that I wont have time to finish (going offline for a few months). If anyone feels like working on this feature, feel free!

geerlingguy:
@oxyc - Godspeed! I'll likely start working on this, as I want to keep working on decoupling things inside the VM a bit more.

oxyc:
@geerlingguy is there anything left to do here except add documentation?

default.config.yml
```diff
@@ -194,6 +194,7 @@ installed_extras:
   - drupalconsole
   # - elasticsearch
   # - java
+  # - drush
```
oxyc: We should probably default to the common case where drush is installed globally, right?

.travis.yml
```diff
@@ -32,6 +33,7 @@ env:
       local_config: tests/ubuntu-16-postgresql.config.yml
       config_dir: /var/www/drupalvm/config
       TEST_INSTALLED_EXTRAS: false
+      DRUSH_BIN: "${DRUPALVM_DIR}/drupal/vendor/drush/drush/drush"
```
oxyc: With #991 drush is no longer installed in this scenario.

default.config.yml
```diff
@@ -215,6 +216,7 @@ extra_packages:
 extra_security_enabled: false
 
 drush_version: "master"
+drush_path: "{{ drupal_composer_install_dir }}/vendor/drush/drush/drush"
```
oxyc: This needs documentation for local codebases. If you're not using composer with drush as a dependency the ansible tasks will fail.

geerlingguy: I'd rather add the task to determine the path and use default if it's not installed via Drupal VM.

provisioning/playbook.yml
```diff
@@ -100,6 +100,11 @@
       set_fact:
         drupal_site_exists: "{{ drupal_site.stat.exists|default(false) }}"
 
+    - name: Update drush_path if drush is not available globally.
+      set_fact:
+        drush_path: "{{ drupal_composer_install_dir }}/vendor/drush/drush/drush"
+      when: "'drush' not in installed_extras"
```
oxyc: Due to [variable precedence](http://docs.ansible.com/ansible/playbooks_variables.html#variable-precedence-where-should-i-put-a-variable) this will override the role default but can still be overridden in `config.yml` if set. I think? I'll test it and make sure a bit later.

oxyc: Nevermind, it's not possible to override it.

default.config.yml
oxyc: My task idea didn't work. If I set the variable it's not possible to override it in `config.yml`. Any ideas?

joestewart:
"Always use project specific drush when available" - Thanks!

geerlingguy:
This is looking good to me. Any other things we need to change? Since it has to be placed into `installed_extras` to continue behaving as it is currently, this should be a minor version bump at least.

oxyc:
I think that's it! The only quirk with this is that you cannot override `drush_path` if you have drush as a composer dependency. I don't see a use case for ever wanting to do that though. If drush isn't a composer dependency, you can change the path.

geerlingguy:
I like it a lot!

geerlingguy:
(This will go into a 3.6.0 release).

PR #714, opened 2016-06-07
[WIP] Followup #667: Make it easier to switch between PHP versions

oxyc:
This still needs work. Not really sure how to adress all issues. Also added some temporary tests for Ubuntu 14.04, unfortunately we never figured out how to test 12.04 using docker so that needs some manual testing (together with memcached and redis).
- We dropped support for PHP 5.5 at some point so I kept it that way.
- I tried to use the same packages as in `geerlingguy.php` with the exceptions:
  - php-mcrypt is added (not sure what happens if you re-provision using PHP 7.0, needs to be tested).
  - php-apcu isn't available in the PPA.
- Also using `ppa:ondrej/php5-compat` in addition to what `beetbox` had so there's less confusion with package names...
- I automatically set the correct xhprof zipball as well.
- Unlike `beetbox` I just fallback to `geerlingguy.php` if `php_version == "7.0"`
- I source a `{{ ansible_distribution }}.yml` file before `config.yml` so that users can still override the values. We could also use a bunch of conditional `set_fact`. That's a lot of overridden variables though..

Also needs to test upgrade path and see what happens when an Ubuntu 14.04 box upgrades. Lots of manual testing...


oxyc:
@thom8 would be great if you could give this a quick review (might just wait a day or so until travis passes. On my network it's a lot faster running tests on travis then locally)

thom8:
@oxyc switching from 7.0 to 5.6 isn't working on `16.04` as it not using the PPA.

I also use CI for testing 😄  + Circle CI also allows you to SSH into the environment to "test" stuff out..
you can enable the build without a `circle.yml` file 😉

geerlingguy:
@thom8 - I've heard that Travis will be adding a similar feature soonish... and that some people can get access to it sooner if they just ask (especially if they do a lot of OSS work).

And yeah, `default.config.yml` might be one of my favorite new things too!

oxyc:
@thom8 everything is supposed to work on 16.04? I thought I tried it and something like memcached or redis didn't like the packages available. Not sure, it was in the beginning of my attempts so maybe it was something unrelated–I'll give it a second try as switching to PHP 5.6 on 16.04 as well would be a nice new feature.

provisioning/tasks/init-debian.yml
```diff
@@ -20,19 +20,24 @@
   apt_repository: repo='ppa:ondrej/apache2'
   when: ansible_distribution_release == "precise" and ansible_distribution == "Ubuntu"
 
-- name: Add repository for PHP 5.5.
-  apt_repository: repo='ppa:ondrej/php5'
-  when: php_version == "5.5" and ansible_distribution == "Ubuntu" and ansible_distribution_version == "12.04"
+- name: Add repository for PHP on Ubuntu.
+  apt_repository: repo='ppa:ondrej/php'
+  when: ansible_distribution == "Ubuntu" and ansible_distribution_version != "16.04"
```
thom8: @oxyc the PPA isn't included in 16.04 so it's probably conflicting with `php56_packages`.

I'm still working out some issues when switching from an existing version so will keep you posted.

oxyc: I tested it briefly locally but the build failed on 16.04 while 14.04 worked out of the box. I hope to get a second look at it this weekend.

That would be awesome. Being able to switch versions without destroying would be quite useful!

geerlingguy: @thom8 / @oxyc - FYI, note that as of https://github.com/geerlingguy/ansible-role-php/pull/135 (latest PHP role version), you can set `php_packages_state: "latest"`, and that, along with a different repo (yum) or PPA (apt), will switch PHP versions on the fly as well.

I just used that to finally bump hostedapachesolr.com to a modern PHP version on CentOS 6 (after testing on staging servers of course) :D

thom8:
@oxyc made some progress here, and we can now cleanly switch between PHP 5.5, 5.6 & 7.0.

As we've locked in @ 14.04, we can always assume that the new PPA in being used.

The module purge could be avoided by setting a version specific module path so modules that require compiling like xdebug & xhprof would have a separate module per version. The PR's referenced above would allow us to remove this purge task -- https://github.com/beetboxvm/beetbox/blob/master/provisioning/ansible/roles/beetbox-php-ppa/tasks/main.yml#L18

provisioning/vars/Debian.yml
```diff
@@ -0,0 +1,32 @@
...
+php_extension_conf_paths:
+  - "/etc/php/{{ php_version }}/fpm/conf.d"
+  - "/etc/php/{{ php_version }}/apache2/conf.d"
+  - "/etc/php/{{ php_version }}/cli/conf.d"
+php_fpm_daemon: "php{{ php_version }}-fpm"
+php_fpm_conf_path: "/etc/php/{{ php_version }}/fpm"
+php_fpm_pool_conf_path: "{{ php_fpm_conf_path }}/pool.d/www.conf"
+php_mysql_package: "php{{ php_version }}-mysql"
+php_opcache_conf_filename: "10-opcache.ini"
+php56_packages:
```
geerlingguy: One potential problem here—is this list overridable via `config.yml`? As long as it is, I think I'm okay with this particular setup.

oxyc:
Related #1031

oxyc:
I have some WP projects going on at the moment but I'll try and work on this next weekend (hopefully).

oxyc:
Moving work over to #1043.

PR #715, opened 2016-06-09
Set node global install dir to a location writable by vagrant

thom8:
https://github.com/geerlingguy/ansible-role-nodejs/issues/29


default.config.yml
```diff
@@ -244,6 +244,9 @@ composer_home_group: "{{ drupalvm_user }}"
 composer_global_packages:
   - { name: hirak/prestissimo, release: '^0.3' }
 
+# NodeJS Configuration.
```
geerlingguy: Can you put this after the MySQL configuration?

PR #717, opened 2016-06-09
php5.6 xdebug documentation

justinlevi:


geerlingguy:
XDebug is built from source and should work no matter which version of PHP you're using; you just need to make sure it's in the list of `installed_extras` inside `config.yml`: https://github.com/geerlingguy/ansible-role-php-xdebug

justinlevi:
Ugh... Yup, I totally forgot to uncomment `- xdebug` in `installed_extras:`.

PR #724, opened 2016-06-10
Add www.drupalvm.dev as a default webserver alias

oxyc:
Maybe not such a useful feature but I add this to almost every project so I don't need to worry if the site was configured to redirect to `www` or not. Thought I'd send over a PR in case interested.

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -92,6 +92,7 @@ configure_drush_aliases: true
 # View the geerlingguy.apache Ansible Role README for more options.
 apache_vhosts:
   - servername: "{{ drupal_domain }}"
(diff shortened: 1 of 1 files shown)


joelpittet:
Sounds a bit like the personal preference, as I do the exact opposite.

geerlingguy:
Works for me—I typically default to `www.*` for everything, because bare domains grate on me (and can have weird side-effects as you build out a site with real-world infrastructure)... but it's good to allow either, and for Apache, users can set the redirect in `.htaccess` if they like, or for Nginx, they can add a redirect vhost.

Issue #734, opened 2016-06-13
Drupalvm sync folder not showing on host machine


mooru:
I have done 2 installs of drupalvm from the git clone but the sync folder has not shown up for either installations (ie under /Users/me/Sites/drupalvm/drupal)
vagrant version 1.8.3
Virtualbox version 5.0.20
MacOSX 10.11.5

I also didn't changed the config.yml file

I have done a second vagrant destroy and vagrant up still no changes. when  i do vagrant ssh i can see the drupal files installed.

geerlingguy:
The only difference between my environments (Windows 10 / Mac OS X 10.11) and yours seems to be that I have Vagrant 1.8.1, while you have 1.8.3.

Over in #683 you had shown that the synced dir seems to not be created, so I'm thinking we can maybe test some more troubleshooting to see where exactly that directory is being generated!

Inside the VM, can you run `sudo cat /proc/mounts` and give the output? Maybe the synced dir isn't being mounted...

mooru:
I ran that command and this is the output. unfortunately my linux skills are not there yet

sysfs /sys sysfs rw,nosuid,nodev,noexec,relatime 0 0
proc /proc proc rw,nosuid,nodev,noexec,relatime 0 0
udev /dev devtmpfs rw,nosuid,relatime,size=490136k,nr_inodes=122534,mode=755 0 0
devpts /dev/pts devpts rw,nosuid,noexec,relatime,gid=5,mode=620,ptmxmode=000 0 0
tmpfs /run tmpfs rw,nosuid,noexec,relatime,size=101620k,mode=755 0 0
/dev/mapper/vagrant--vg-root / ext4 rw,relatime,errors=remount-ro,data=ordered 0 0
securityfs /sys/kernel/security securityfs rw,nosuid,nodev,noexec,relatime 0 0
tmpfs /dev/shm tmpfs rw,nosuid,nodev 0 0
tmpfs /run/lock tmpfs rw,nosuid,nodev,noexec,relatime,size=5120k 0 0
tmpfs /sys/fs/cgroup tmpfs ro,nosuid,nodev,noexec,mode=755 0 0
cgroup /sys/fs/cgroup/systemd cgroup rw,nosuid,nodev,noexec,relatime,xattr,release_agent=/lib/systemd/systemd-cgroups-agent,name=systemd,nsroot=/ 0 0
pstore /sys/fs/pstore pstore rw,nosuid,nodev,noexec,relatime 0 0
[...]

PR #737, opened 2016-06-15
Issue #734: Fix Vagrant 1.8.4 failing to mount NFS synced folders with empty mount_options

oxyc:


oxyc:
@geerlingguy could you test this on Vagrant 1.8.1 or have you already upgraded?

oxyc:
Tested and it also works if I set `mount_options: ['vers=3', 'udp']` which is what vagrant does on linux guests.

geerlingguy:
I'm still on 1.8.1 for now, so I'll test there.

geerlingguy:
Worked! I'm going to tag another bugfix release so those using Vagrant 1.8.3+ don't hit this.

Issue #655, opened 2016-05-23
Ansible roles re-downloaded on each provision


opdavies:
Since 3.0.0, the Ansible roles are being re-downloaded each time `vagrant provision` is run.

oxyc:
It's a feature https://github.com/geerlingguy/drupal-vm/pull/611. My main reason for standing behind that commit was that we had suggested the `ansible-galaxy install` command ~29 times here in the issue queue. Also I believe that the common user _does_ want to re-install the roles whenever they re-provision the VM.

There's a fairly easy workaround for _power users_ though. Add a `Vagrantfile.local` with the following content:

```rb
# Disable the galaxy role re-installation during provisions.
config.vm.provisioners[0].config.galaxy_role_file = nil
```

@geerlingguy maybe this should be brought up for more debate though? Or we could add some way to toggle it of. I still feel it should be the default for the common case though.

opdavies:
OK, cool. I was just wondering if it was intentional or not.

I was assuming that it would download the roles once unless told to do so, sort of how `vagrant reload --provision` works, but I can see why it would be useful to automatically download them each time when provisioning to ensure that everyone has the latest specified versions.

thom8:
If only there was an galaxy update option -- https://github.com/ansible/ansible/issues/6466

danbohea:
https://github.com/ansible/galaxy-issues/issues/49

thom8:
Nice! I ended up using ansible to install roles so it could be idempotent --
https://github.com/beetboxvm/beetbox/blob/master/provisioning/ansible/playbook-roles.yml

However, it also clones the roles from github in debug mode to make it easier to push PR's which would also be a nice feature for galaxy...

geerlingguy:
Yeah, it's the lesser of two evils. An extra minute or two each provision by default, but a lot of extra issues not opened up in here. It's what the Vagrant Ansible (built in) provisioner does, and it's also what the `windows.sh` script does (in JJG-Ansible-Windows), so now it's consistent.

Power users (or people like me who provision all day long) can override, so it's not a huge issue.

It also helps boost my download counts :D https://galaxy.ansible.com/explore#/

danbohea:
> It also helps boost my download counts :D

LOL

PR #753, opened 2016-06-24
Issue #655: Add workaround as hint in documentation.

stephang:
When I use `vagrant provision` often, ansible roles are re-downloaded each time `vagrant provision` is run.

A workaround is described in issue [#655](https://github.com/geerlingguy/drupal-vm/issues/655#issuecomment-221106065). I added the workaround to the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ Drupal VM runs on almost any modern computer that can run VirtualBox and Vagrant
   - You can change the installed version of Drupal or drush, or any other configuration options, by editing the variables within `config.yml`.
   - Find out more about local development with Vagrant + VirtualBox + Ansible in this presentation: [Local Development Environments - Vagrant, VirtualBox and Ansible](http://www.slideshare.net/geerlingguy/local-development-on-virtual-machines-vagrant-virtualbox-and-ansible).
   - Learn about how Ansible can accelerate your ability to innovate and manage your infrastructure by reading [Ansible for DevOps](http://www.ansiblefordevops.com/).
+  - Sometimes you want to be faster when provisioning often on the same VM. It can make sense to [skip repeated downloading ansible roles](https://github.com/geerlingguy/drupal-vm/issues/655#issuecomment-221106065) on each `vagrant provision`.
 
 ## License
 


geerlingguy:
Note that this is documented here: http://docs.drupalvm.com/en/latest/other/overriding-configurations/#extending-the-vagrantfile-with-vagrantfilelocal

Maybe we should just add that particular change to it's own heading in that file?

stephang:
> Maybe we should just add that particular change to it's own heading in that file?

That might solve the issue.

However, I'm still sceptic if this would solve the issue. First of all it took me a while to find this issue and its workaround. I asked google for somethink like *drupalvm provision skip galaxy* or *drupalvm faster provisioning*, aware that the terms are quite general. Eventually, I found the issue only when browsing the issue queue.

Second, I'd assume that slow provising is something that many people face. Finding the workaround should be easer, so this should be prominently documented.

So, what's your thoughts on this?

geerlingguy:
@stephang - Here's the problem—since Galaxy can't yet do idempotent updates, and since role versions are updated from time to time (even in minor releases), its safer—even though it adds 2-3 min of time per provision—to default to always doing it.

And I don't want people to start disabling it without knowing full well what they're doing, because then I'll start getting bug reports from people who have disabled it and updated Drupal VM, but run into bugs because now their role versions aren't being kept up to date.

Additionally, if you don't have Ansible installed it's a moot point, because both JJG-Ansible-Windows and the `ansible_local` provisioner built into Vagrant install everything every time anyways.

stephang:
> And I don't want people to start disabling it without knowing full well what they're doing ...

This is a thin line here, You don't recommend to disable galaxy updates. But, we want to know pro users about the workaround.

In this case we'll need to improve documentation. Lets add a heading, the problem description and a little warning ("You should know what you're doing. Uncomment before writing bug reports"). Users would find it via google, but would have a clear warning. Would that work out?

geerlingguy:
Yeah, I think that would work.

geerlingguy:
There's also another cool potential workaround that was mentioned by someone in #ansible IRC. I think @joestewart ? — It would allow for a way to manually install the galaxy updates if desired during a provisioning run.

geerlingguy:
Ah... he just posted a PR, nice!

Check out https://github.com/geerlingguy/drupal-vm/pull/756 as well.

joestewart:
Will work in Vagrantfile.local.

To skip downloading roles run

`vagrant provision  --provision-with ansible-playbook`

https://gist.github.com/joestewart/f22d854591a909461f031f262f35bc6d

Vagrantfile.local
```
# vagrant provision  --provision-with ansible-galaxy
config.vm.provision 'ansible-galaxy', type: 'ansible' do |ansible|
  ansible.playbook = "#{host_drupalvm_dir}/provisioning/playbook.yml"
  ansible.galaxy_role_file = "#{host_drupalvm_dir}/provisioning/requirements.yml"
  ansible.raw_arguments = "--syntax-check"
end

# vagrant provision  --provision-with ansible-playbook
config.vm.provision 'ansible-playbook', type: 'ansible' do |ansible|
  ansible.playbook = "#{host_drupalvm_dir}/provisioning/playbook.yml"
  ansible.extra_vars = {
    config_dir: host_config_dir
  }
end
```

geerlingguy:
Closing in favor of #756 — I think that may be a better long-term plan, as it could make it easier for people to choose to reinstall roles on-the-fly.

PR #756, opened 2016-06-24
Use named provisioners for Ansible Galaxy role install flexibility

joestewart:
Allows the use of:

`vagrant provision` - Runs the default provisioning.
`vagrant provision  --provision-with ansible-galaxy` - Runs only the galaxy role installation
`vagrant provision  --provision-with ansible-playbook` - Skips downloading the roles via galaxy.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -124,7 +124,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     config.vm.provision 'ansible-galaxy', type: 'ansible' do |ansible|
       ansible.playbook = "#{host_drupalvm_dir}/provisioning/playbook.yml"
       ansible.galaxy_role_file = "#{host_drupalvm_dir}/provisioning/requirements.yml"
-      ansible.raw_arguments = "--syntax-check"
+      ansible.raw_arguments = '--syntax-check'
     end
     # vagrant provision  --provision-with ansible-playbook
     config.vm.provision 'ansible-playbook', type: 'ansible' do |ansible|
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -120,9 +120,15 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
 
   # Provisioning. Use ansible if it's installed, JJG-Ansible-Windows if not.
   if which('ansible-playbook')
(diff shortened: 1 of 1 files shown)


geerlingguy:
Adding note from @joestewart in #753:

Will work in Vagrantfile.local.

To skip downloading roles, add the following in `Vagrantfile.local`:

```ruby
# vagrant provision  --provision-with ansible-galaxy
config.vm.provision 'ansible-galaxy', type: 'ansible' do |ansible|
  ansible.playbook = "#{host_drupalvm_dir}/provisioning/playbook.yml"
  ansible.galaxy_role_file = "#{host_drupalvm_dir}/provisioning/requirements.yml"
  ansible.raw_arguments = "--syntax-check"
end

# vagrant provision  --provision-with ansible-playbook
config.vm.provision 'ansible-playbook', type: 'ansible' do |ansible|
  ansible.playbook = "#{host_drupalvm_dir}/provisioning/playbook.yml"
  ansible.extra_vars = {
    config_dir: host_config_dir
  }
end
```

...and run `vagrant provision  --provision-with ansible-playbook`

See more: https://gist.github.com/joestewart/f22d854591a909461f031f262f35bc6d

geerlingguy:
I'm not sure if I'll necessarily merge in the PR as-is... the main thing is I want `vagrant up` and `vagrant provision` to work exactly as they do now. It would be ideal if the new code would add the ability to do something like `vagrant provision --no-galaxy` to skip the galaxy install, or something along those lines.

joestewart:
Hi, the comment is unclear to me.  If it is easier, ping me on IRC.

"the main thing is I want vagrant up and vagrant provision to work exactly as they do now"

What behavior change with this?

`vagrant up` or `vagrant provision` is used normally.

"It would be ideal if the new code would add the ability to do something like vagrant provision --no-galaxy to skip the galaxy install"

I didn't find a way with syntax that simple, so used the named provisioners like:

`vagrant provision  --provision-with ansible-playbook`

I understand if you rather this be used in Vagrantfile.local instead.

geerlingguy:
Since 3.2.0 (I think that was the version), the roles are included in the repo, so this feature is no longer necessary. Thanks for the work on this, and I'll definitely have to refer back to it for some other projects where I _won't_ include roles...

Issue #754, opened 2016-06-24
Using https on nginx


mark-casias:
Hi there,
  We are trying to use nginx locally to properly mirror the Pantheon environment, but don't see a way to configure HTTPS in the config.yml. I looked at the playbook and don't see anything there. Am I just missing it?

geerlingguy:
See: https://github.com/geerlingguy/ansible-role-nginx#role-variables

You can basically configure the server using whatever server block directives you need (setting up ciphers, cert paths, using port 443, http2, etc.).

We should probably add an example in the docs for Nginx, like we have currently for Apache: http://docs.drupalvm.com/en/latest/extras/ssl/

joestewart:
Looked like the provisioning/templates/nginx-vhost.conf.j2 template did not support the extra_parameters as in the template for ansible-role-nginx.  So created #761

geerlingguy:
D'oh... disadvantages of diverging from the upstream for a feature :P

mchelen:
Yeah I think some more documentation here would be nice, or include it commented out in `config.yml`. Setting an Apache server to use SSL is pretty clear from looking at the config, it would be nice to say the same for Nginx.

mattelkins:
Just thought I'd share my findings as I've been having a look in to this today. Here's the host definition I added to `nginx_hosts` in my `config.yml`:

```
- listen: "443 ssl default_server"
  server_name: "{{ drupal_domain }} www.{{ drupal_domain }}"
  ssl_certificate: "/vagrant/ssl/my.crt"
  ssl_certificate_key: "/vagrant/ssl/my.key"
  root: "{{ drupal_core_path }}"
  is_php: true
```

This definition is in addition to the default:

```
- server_name: "{{ drupal_domain }} www.{{ drupal_domain }}"
  root: "{{ drupal_core_path }}"
  is_php: true
```

`vagrant up` completed successfully, but I had to change a few configuration things.

In the server block within `/etc/nginx/conf.d/{{ drupal_domain }}.conf` (where `drupal_domain` is the domain I specified in the `vagrant_hostname` variable in my `config.yml`), only `listen 443 ssl default_server` was specified. The default `listen 80` host configuration was nowhere to be found, so I added it to the server block:

```
server {
    listen 80 default_server;
    listen 443 ssl default_server;

    ...
}
```

I'm guessing I needed to do this because both the default host definition and my SSL host definition share the same `server_name`. Would this have caused the host configuration to be overwritten?

[...]

PR #761, opened 2016-06-27
Nginx extra parameters

joestewart:
From Using https on nginx #754

The provisioning/templates/nginx-vhost.conf.j2 template did not support the extra_parameters as in the template for ansible-role-nginx


joestewart:
deleting.  will rebase with up to date branch.

geerlingguy:
@joestewart - Thanks!

PR #764, opened 2016-06-28
Update README.md

amstaker:
Updated documentation related to the hosts file updates in Windows environments.  This likely will need to be set manually with the current setup, I also created a related ticket to address underlying root cause, but in the meantime this might be a pitfall that folks run into and the documentation gives a clear way out and to a working site.


geerlingguy:
@amstaker - I would recommend you either install the hostsupdater plugin (`vagrant plugin install vagrant-hostsupdater`) or follow the directions on the dashboard (accessed at the configured IP address); I don't think there's need to add extra documentation here, as it's the same situation on Mac and Linux as well.

I just tested everything latest with Windows 10 and Vagrant hostsupdater, and the hosts were automatically configured correctly—see https://youtu.be/mNio_aXMLos?t=241

amstaker:
Ok so we now this works on Win10 and that's fine. But not necessarily on 7, I think the documentation is misleading and needs some clarity.  

Also if you manually do need to setup those addresses setting up the host file should be documented so it's clear and easy to do.  If those addresses are not documented the only way you'll see them is if you actually look in your hosts file.  

The documentation isn't necessarily bad it could just be a bit better here.

BTW thanks for your quick responses and feedback.

PR #768, opened 2016-06-29
Update devel package in composer and config

illepic:
Update devel composer package mentioned in `default.config.yml` and `example.drupal.composer.json` so they

- don't explode Composer
- are consistent
- are updated

### DIFF
--- a/example.drupal.composer.json
+++ b/example.drupal.composer.json
@@ -19,7 +19,7 @@
         "composer/installers": "^1.0.20",
         "drupal-composer/drupal-scaffold": "^2.0.1",
         "drupal/core": "~8.0",
-        "drupal/devel": "1.0.0-alpha1"
+        "drupal/devel": "1.x-dev"
     },
     "replace": {
         "drupal/drupal": "*"
--- a/example.drupal.composer.json
+++ b/example.drupal.composer.json
@@ -19,7 +19,7 @@
         "composer/installers": "^1.0.20",
         "drupal-composer/drupal-scaffold": "^2.0.1",
         "drupal/core": "~8.0",
-        "drupal/devel": "1.0.x-dev"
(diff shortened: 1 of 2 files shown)


example.drupal.composer.json
```diff
@@ -19,7 +19,7 @@
         "composer/installers": "^1.0.20",
         "drupal-composer/drupal-scaffold": "^2.0.1",
         "drupal/core": "~8.0",
-        "drupal/devel": "1.0.x-dev"
+        "drupal/devel": "1.0.0-alpha1"
```
geerlingguy: I'd like to keep it -dev

geerlingguy:
Failure:

```
The requested package drupal/devel 1.0.0-alpha1 exists as drupal/devel[7.1.0, 7.1.0-alpha1, 7.1.0-beta1, 7.1.0-beta2, 7.1.0-rc1, 7.1.1, 7.1.2, 7.1.3, 7.1.4, 7.1.5, 8.1.0-alpha1, dev-7.x-1.x, 7.1.x-dev, dev-8.x-1.x, 8.1.x-dev, dev-8.x-1.x-kint] but these are rejected by your constraint.
```

This is because the non-drupal.org packagist (used for drupal-project currently, until https://github.com/drupal-composer/drupal-project/pull/159 is incorporated) uses a different format for the version numbering.

illepic:
Odd, this builds locally: `1.x-dev`. Does this mean it should be updated to `8.1.x-dev`?

geerlingguy:
Maybe it's Travis' PR cache. I'll empty the cache and re-run the test.

geerlingguy:
I'm going to close this for now—until the drupal composer project updates to using all the drupal.org endpoints, I don't want to make a change that could break the version numbering differences between the two (our composer.json and the drupal project composer config).

PR #775, opened 2016-07-03
Emphasised need to install Virtualbox

jonathanjfshaw:


geerlingguy:
Is this just in reference to making sure the version of VirtualBox is the latest version? Also, the text regarding the VMware fusion plugin is a bit incorrect... it won't automatically install either VirtualBox or VMware Fusion just because the paid plugin is present.

jonathanjfshaw:
The previous version of the test began with "Vagrant will automatically install Virtualbox ..." which encouraged me to think I didn't need to install virtualBox. Which led to unfriendly vagrant error when I did "vagrant up" on drupal vm.

So I was trying to rewrite the text to make it bluntly clear that virtualbox needed to be installed. Assuming I've understood that right ...

geerlingguy:
@jonathanjfshaw - What version of Vagrant do you have installed? I think it was in the 1.7.x range when VirtualBox actually started installing VirtualBox for you if it wasn't installed already. You can find the version with `vagrant --version`.

jonathanjfshaw:
Vagrant 1.84. But I'm on Windows 10, that may be the issue. It may also be that when I "installed" virtualBox I effectively updated it. Either way I'll modify the PR to attempt to clarify that line.

jayusa123:
I uninstalled VirtualBox from my machine (MacOS 10.11.5) to test this premise. Vagrant (v1.8.4) doesn't automatically install VirtualBox when I run `vagrant up`. Instead, it complains:
>No usable default provider could be found for your system.

>Vagrant relies on interactions with 3rd party systems, known as
"providers", to provide Vagrant with resources to run development
environments. Examples are VirtualBox, VMware, Hyper-V.

>The easiest solution to this message is to install VirtualBox, which
is available for free on all major platforms.

>If you believe you already have a provider available, make sure it
is properly installed and configured. You can see more details about
why a particular provider isn't working by forcing usage with
`vagrant up --provider=PROVIDER`, which should give you a more specific
error message for that particular provider.

I also tried specifying a default provider by adding `ENV['VAGRANT_DEFAULT_PROVIDER'] = 'virtual box'` to my Vagrantfile. But that also didn't work:
>The provider 'virtualbox' that was requested to back the machine
'drupalvm' is reporting that it isn't usable on this system. The
reason is shown below:

>Vagrant could not detect VirtualBox! Make sure VirtualBox is properly installed.
Vagrant uses the `VBoxManage` binary that ships with VirtualBox, and requires
this to be available on the PATH. If VirtualBox is installed, please find the
`VBoxManage` binary and add it to the PATH environmental variable. 

The `--install-provider` command line option should be unnecessary, since it is set to true by default (according to [Vagrant's documentation](https://www.vagrantup.com/docs/cli/up.html)) but I tried that anyway and still got the same error.

To me, based on my experience, this documentation is wrong. But it could just be a quirk with my system.

geerlingguy:
@jayusa123 - Interesting... on my Windows laptop, I just tested by uninstalling VirtualBox, and Vagrant automatically installed it when I ran `vagrant up` (albeit an older version, like 5.0.11 or something).

But yeah, we might still want to encourage people to manually install VirtualBox.

geerlingguy:
I'm updating the docs to recommend installing VirtualBox manually again, just because it seems to be updated regularly enough that the Vagrant-installed version will always be quite behind.

Issue #792, opened 2016-07-11
Devel module dependency undocumented in "Add Drupal VM as a Composer dependency"


ctorgalson:
When [adding Drupal VM as a composer dependency](http://docs.drupalvm.com/en/latest/other/drupalvm-composer-dependency/#add-drupal-vm-as-a-composer-dependency), the docs don't mention the [Devel dependency in `default.config.yml`](https://github.com/geerlingguy/drupal-vm/blob/master/default.config.yml#L49).

If you follow the docs as-is, the process hangs at `TASK [Install configured modules with drush.]`.

The docs should mention this dependency and specify that either

1. Devel should be added to `composer.json` file in the user's own project, or
2. `drupal_enable_modules` should be altered in the user's own `config.yml` file.

geerlingguy:
Fixed in #793

PR #793, opened 2016-07-11
Adds drupal_enabled_modules info to composer section

ctorgalson:
If this makes sense in the first place, please let me know if the wording needs to change.

- Without either adding devel as a dependency to composer.json, or
  removing devel from drupal_enabled_modules in config.yml, vagrant up
  hangs when enabling modules.
- gh-792


geerlingguy:
Good catch!

PR #794, opened 2016-07-11
Adds optional defaults for php-fpm user & group.

### LINKED ISSUE #781: Using nginx runs php-fpm under the wrong user
Using nginx, the worker process runs as user `nginx`, but the php-fpm pools run as user `apache`. This causes issues with things like log files, where (for example) pimpmylog expects logs to be owned by `apache`, when they're owned by `nginx`. Other goofy issues happen too, like `/var/log/php-fpm` being owned by `apache`.

config.yml is fairly standard, set up for a local environment (no makes/builds). Relevant:
```
# The web server software to use. Can be either 'apache' or 'nginx'.
drupalvm_webserver: nginx
```
After ssh'ing into the newly built vm:
```
$ ps aux | grep nginx
root      1214  0.0  0.1  48084  1500 ?        Ss   13:50   0:00 nginx: master process /usr/sbin/nginx -c /etc/nginx/nginx.conf
nginx     1215  0.0  0.2  48880  3008 ?        S    13:50   0:00 nginx: worker process
```
and
```
$ ps aux | grep apache
apache    1260  0.0  1.3 652112 14056 ?        S    13:50   0:00 php-fpm: pool www
apache    1261  0.0  1.2 652088 12328 ?        S    13:50   0:00 php-fpm: pool www
apache    1262  0.0  1.1 652152 11412 ?        S    13:50   0:00 php-fpm: pool www
apache    1263  0.0  1.6 652344 16432 ?        S    13:50   0:00 php-fpm: pool www
apache    1264  0.0  1.2 652152 12760 ?        S    13:50   0:00 php-fpm: pool www
```

I expect this might be a downstream problem with the geerlingguy.nginx role, but I haven't dug in enough to find out yet.

entendu:
Fixes #781 which is solved upstream but not yet referenced in requirements. Also adds an optional user/group param for php-fpm.

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -230,6 +230,10 @@ php_max_input_vars: "4000"
 php_enable_php_fpm: true
 php_fpm_listen: "127.0.0.1:9000"
 
+# If you want to run php-fpm as a user other than your OS's default user,
+# uncomment this line and specify. For example, to run as the "nginx" user:
+# php_fpm_pool_user: "nginx"
+
 composer_path: /usr/bin/composer
 composer_home_path: "/home/{{ drupalvm_user }}/.composer"
 composer_home_owner: "{{ drupalvm_user }}"
--- a/default.config.yml
+++ b/default.config.yml
@@ -230,9 +230,10 @@ php_max_input_vars: "4000"
 php_enable_php_fpm: true
 php_fpm_listen: "127.0.0.1:9000"
 
-# If you want to run php-fpm as a user other than your OS's default user,
-# uncomment this line and specify. For example, to run as the "nginx" user:
+# If you want to run php-fpm as a user/group other than your OS's default user,
+# uncomment this line and specify. For example, to run as the "nginx" user & group:
 # php_fpm_pool_user: "nginx"
+# php_fpm_pool_group: "nginx"
 
 composer_path: /usr/bin/composer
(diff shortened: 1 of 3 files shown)


default.config.yml
```diff
@@ -230,6 +230,11 @@ php_max_input_vars: "4000"
 php_enable_php_fpm: true
 php_fpm_listen: "127.0.0.1:9000"
 
+# If you want to run php-fpm as a user/group other than your OS's default user,
```
geerlingguy: Instead of adding in this file, can you just add a note on this docs page (http://docs.drupalvm.com/en/latest/other/webservers/#using-nginx) mentioning these two variables and how to set them if switching to `nginx`? Since it's a smaller use case (at least at this point in time), I'd rather not add the four lines to `default.config.yml`.

provisioning/requirements.yml
```diff
@@ -36,7 +36,7 @@
 - src: geerlingguy.nodejs
   version: 3.2.0
 - src: geerlingguy.php
-  version: 3.2.1
+  version: 3.2.2
```
geerlingguy: Good catch! I thought I had updated the role in requirements... but apparently not.

entendu:
@geerlingguy, good call. Added some explanation to webservers.md (http://docs.drupalvm.com/en/latest/other/webservers/#using-nginx) and reverted changes to default.config.yml.

PR #803, opened 2016-07-13
Add elasticsearch to installed extras

joestewart:
Adds elasticsearch to installed extras.

Note:

For use on the host outside vagrant will depend on an upstream change like in PR's 9 or 17 or equivalent.


geerlingguy:
Looks good—can you also add it to the README as one of the installed/optional extras?

geerlingguy:
Perfect!

geerlingguy:
Also to follow up on first comment in this PR thread, I had merged in https://github.com/geerlingguy/ansible-role-elasticsearch/pull/21, which you're using for the example vars. Thanks!

PR #807, opened 2016-07-16
Add ability to override drush make command line options.

jayusa123:
I needed to be able to customize the drush make command line options for a project. So, I added the ability to completely override the options used during provisioning. I think others may find it useful as well.

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -39,6 +39,7 @@ drupalvm_webserver: apache
 # would like to configure 'vagrant_synced_folders' and 'apache_vhosts' manually.
 build_makefile: false
 drush_makefile_path: "{{ config_dir }}/drupal.make.yml"
+drush_make_options: "--no-gitinfofile"
 
 # Set 'build_makefile' to 'false' and this to 'true' if you are using a
 # composer based site deployment strategy.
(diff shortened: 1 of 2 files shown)


jayusa123:
Thought I'd just provide an example situation where one might want to override the drush make options. I have created a dev DrushVM repo that pulls down core and our custom installation profile. The profile then pulls in a number of custom modules that we maintain. It is very helpful to run drush make with the `--working-copy` option so that we have usable clones of all the custom modules' repos. This way we can just push the code after making changes to any of our modules.

Thanks!
Jay

geerlingguy:
I'm not necessarily opposed to this addition (I'll merge after poking and prodding it a little), but one note: you can also specify working-copy per-project inside your make file, e.g. https://github.com/geerlingguy/drupal-vm/blob/master/example.drupal.make.yml#L21 (in case you hadn't seen that before; I learned of it a year or so ago and was happy to know I could specify that option in the makefile).

jayusa123:
Hmm, no kidding. I haven't seen that before. Do you know if it applies to all the modules in that case or just core?

geerlingguy:
@jayusa123 - I think all modules.

Issue #431, opened 2016-02-11
Provision specific task


justinlevi:
Is there a way to provision a specific task? 

I want to run `vagrant provision --provision-with post_provision_scripts` instead of having to go through the entire provisioning but it's not working for me.

oxyc:
I think not as the `vagrant` provisioner is still the same, `ansible`.

If you're on the master branch (or have commit 8a05b75dcf51eb) you could get it to work by not using Drupal VMs own `post_provisioning_scripts` variable but instead extend vagrant with a shell provisioner. To do that you create a `Vagrantfile.local` with the following contents:
```ruby
config.vm.provision "shell", path: "my-post-provisinoing-script.sh"
```
Then you can run `vagrant provision --provision-with shell` and it should work.

geerlingguy:
Technically you can also do this if you use Ansible outside Vagrant, or if Vagrant had a way to pass through `tags` to Ansible.

I don't use them in Drupal VM, but it's easy enough to add them, and if Vagrant supports them via CLI sometime (not just through the Vagrantfile config), I will add them.

With Ansible, you can just do something like `ansible-playbook -i provisioning/inventory provisioning/playbook.yml --tags=post` or something along those lines.

geerlingguy:
And yes, for now, the workaround would be to add what `oxyc` mentioned. I'll be tagging a new release once I get a few other upstream issues worked out, but for now that feature is available in `master`.

justinlevi:
Thanks for the information guys.

geerlingguy:
See @joestewart's PR above for a nice approach here.

PR #820, opened 2016-07-24
Pass ansible arguments via environment variable if provided.

joestewart:
Opened PR for discussion to address #431 

Possible usage to run only post provisioning tasks and scripts as suggested.

`ANSIBLE_ARGS='--tags=post' vagrant provision`

As mentioned in #431 tags would need to be added for this to be useful.

We could restrict the enviroment variable to tags only like 

"ansible.tags = ENV['ANSIBLE_TAGS']"

But the more general argument would allow other flexibility.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -125,7 +125,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       ansible.extra_vars = {
         config_dir: host_config_dir
       }
-      ansible.raw_arguments = ENV['ANSIBLE_ARGS']
+      ansible.raw_arguments = ENV['DRUPALVM_ANSIBLE_ARGS']
     end
   else
     config.vm.provision 'ansible_local' do |ansible|
@@ -133,7 +133,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       ansible.extra_vars = {
         config_dir: guest_config_dir
       }
-      ansible.raw_arguments = ENV['ANSIBLE_ARGS']
+      ansible.raw_arguments = ENV['DRUPALVM_ANSIBLE_ARGS']
     end
   end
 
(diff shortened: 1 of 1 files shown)


geerlingguy:
I really like this, nice idea! Might think of a different env variable name though, just to avoid any potential namespace collisions. Maybe `DRUPALVM_ANSIBLE_ARGS`?

joestewart:
Updated to use DRUPALVM_ANSIBLE_ARGS.  It's not going to be a short command line anyway so this seems good.

geerlingguy:
If the env variable is not set, does this cause any issues/warnings?

joestewart:
I haven't seen any difference when the variable isn't set.  Including warnings.

oxyc:
Other useful flags:

- `--vault-password-file ~/.vault_pass.txt`
- `--start-at-task=Run configured post-provision ansible task files.` _due to how vagrant passes the arguments you can't use quotes here._
- `--list-tasks` would be useful for debugging

geerlingguy:
@oxyc - Would you like to take a crack at adding one or more of those and build on top of this PR?

oxyc:
With the current PR it already works.

```
DRUPALVM_ANSIBLE_ARGS='--start-at-task=Run configured post-provision ansible task files.' vagrant provision
```

geerlingguy:
@oxyc - Ah... obvious in hindsight.

Could we add a note in the docs about this new option, and some suggested uses?

oxyc:
Sure.

One minor issue with this is that it only supports a single argument. Vagrant requires the arguments to be passed as an array if there are more than one. This is also why I'm not using quotes in my `--start-at-task` example. I guess we can just add a note on that.

geerlingguy:
Sounds good.

geerlingguy:
Closing for #992 — see progress there.

PR #836, opened 2016-07-30
Remove shell-module touch in mysql role, use file-module instead.

rpsu:
This patch solves the WARNINGs during provisioning.


geerlingguy:
Please submit the PR against the upstream role: https://github.com/geerlingguy/ansible-role-mysql/pulls

PR #839, opened 2016-07-31
vagrant halt [id] causes "warning: already initialized constant VAGRANTFILE_API_VERSION"

joestewart:
When using the id argument for vagrant halt to act on a vagrant installation in a different directory a warning is currently generated.  This is due to the VAGRANTFILE_API_VERSION already being define.  This PR checks if it has been defined.

Warnings - 
```
Vagrantfile:3: warning: already initialized constant VAGRANTFILE_API_VERSION
```
```
Vagrantfile:3: warning: previous definition of VAGRANTFILE_API_VERSION was here
```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,6 +1,6 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
-VAGRANTFILE_API_VERSION = '2'
+VAGRANTFILE_API_VERSION = '2' unless defined? VAGRANTFILE_API_VERSION
 Vagrant.require_version '>= 1.8.1'
 
 # Absolute paths on the host machine.


geerlingguy:
When building a `Vagrantfile.local` or a separate 'delegate' `Vagrantfile`, what's the rationale for defining `VAGRANTFILE_API_VERSION` in there as well? If it makes sense to support the constant definition in multiple places, I'm okay with this... I just haven't run into this issue before and am wondering about the use case.

joestewart:
The use case is when multiple DrupalVM installations are running.

vagrant global-status lists the id's of the vm's

From directory of vm1.

vagrant halt with id's of vm 2,3,4 etc. to shut them down without changing directories.

So not really about a delegating Vagranfile or Vagrantfile.local.

Sorry not more clear,  https://www.vagrantup.com/docs/cli/halt.html doesn't describe the id argument.

thom8:
You can use `Vagrant.configure("2") do |config|` as there's currently no logic to switch version compatibility.

https://www.vagrantup.com/docs/installation/backwards-compatibility.html

morsok:
I'm also encountering this warning quite often, the check seems harmless and would help with that thank you !

geerlingguy:
Sounds good then!

PR #848, opened 2016-08-04
Adding support for vagrant-proxyconf. Fixes #847

### LINKED ISSUE #847: Integrate with vagrant-proxyconf
For those of us behing corporate proxy it would be awesome if DrupalVM can integrate with this plugin !

[vagrant-proxyconf github](https://github.com/tmatilai/vagrant-proxyconf)

A PR should follow soon.

morsok:
I think we should add documentation about that ?

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -75,6 +75,17 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # Vagrant box.
   config.vm.box = vconfig['vagrant_box']
 
+  # Configure the proxy.
+  if Vagrant.has_plugin?("vagrant-proxyconf")
+    if vconfig["proxy"]["enable_proxy"]
+      config.proxy.http = vconfig["proxy"]["http_proxy"] || ENV['http_proxy']
+      config.proxy.https = vconfig["proxy"]["https_proxy"] || ENV['https_proxy']
+      config.git_proxy.http = vconfig["proxy"]["http_proxy"] || ENV['http_proxy']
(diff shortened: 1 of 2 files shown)


geerlingguy:
https://github.com/tmatilai/vagrant-proxyconf for reference. Haven't used this plugin, will take a look.

geerlingguy:
@morsok - Can you also provide some more background about your use case? Will be helpful for (a) historical purposes and (b) my review/possible expansion of docs.

morsok:
@geerlingguy The background is simple, i'm behind a corporate proxy so i need to set it up on the vm or else it will be unable to install anything coming from the net.

once set up every request goes to the proxy who decide if it continues or get stopped.

Without it any person infortunate enough to be behind such proxies won't be able to use DrupalVM, so sure you could install the plugin and set it up in a VagrantFile.local but i don't see the harm of having it directly in (except the added strain of documenting it).

geerlingguy:
I like this. I'll adjust a couple of the parameters, and add a little documentation before final merge, but I think it will be helpful to the numerous users who have to run things behind a corporate proxy and don't want to have to log in and manually configure things, or add in a pre-provisioning script.

oxyc:
I vote for using `Vagrantfile.local` and having this in documentation only. If the problem is common it could be it's own section _Using Drupal VM behind a corporate proxy_.

geerlingguy:
@oxyc++ Let's do that — @morsok would you be willing to write up a docs page for this, otherwise @oxyc or I can.

morsok:
Sadly not much time right now, if you can it would be best, sorry.

geerlingguy:
Sounds good, thanks!

PR #853, opened 2016-08-05
added /web to the directory

mndnrdev:
throws a Symfony Exception if you're not in the correct directory


geerlingguy:
Good catch! Going to reset the failed test and make sure it passes, then merge.

PR #859, opened 2016-08-10
Only set mysql_enablerepo when not defined

stevenspasbo:


geerlingguy:
Makes sense, if you want to use a different repo. Thanks!

Issue #66, opened 2015-04-06
Vagrant Virtualbox NFS / mounting issues


enapthine:
Ran into issues when mounting shares (NFS & not), eventually discovered was caused by differing VirtualboxGuestAdditions that wasn't being detected. Not specifically DrupalVM related, but worth documenting. This could be related to earlier bug report regarding NFS shares also, as it stalls at the same location.

Resolved with installing vbguest plugin via:

```vagrant plugin install vagrant-vbguest```

stefanruijsenaars:
I think I may have mounting issues as well, I am running into this issue: https://www.drupal.org/node/2234229

I wonder if replacing NFS with this could help? http://wolfgangziegler.net/auto-rsync-local-changes-to-remote-server#comments

geerlingguy:
@stefanruijsenaars - I've hit that issue as well, but not with Drupal VM: https://www.drupal.org/node/2234229#comment-9744087

You can always switch to rsync if you would rather try that sync method, but with the out-of-the-box configuration, the Drupal install will be written over if you put files into your local folder that's synced inside the VM.

To use rsync, just set the `type` of the synced folder to `rsync`. I'm hoping Vagrant will someday add two-way rsync out of the box, though!

@enapthine - I think I may add a Wiki page with NFS common problems and fixes; enough people have trouble with shared folders of all varieties, and simply keeping my VirtualBox boxes up to date isn't enough all the time (I hate how VirtualBox seems to cause problems every other version upgrade!)

stefanruijsenaars:
@geerlingguy  I think the solution in the link actually bypasses Vagrant synced folders completely, it's a script by fago that watches a directory for changes using inotify-tools and rsyncs the changes to a remote system via SSH. Supposedly it should be faster than NFS, see https://github.com/drunomics/syncd

stefanruijsenaars:
I reckon this is some kind of problem with permissions, now I had a Simpletest fail because it couldn't create a directory :)

stefanruijsenaars:
Which I fixed by adding the following to the synced folders section in the Vagrantfile:

    owner: "vagrant",
    group: "www-data",
    mount_options: ["dmode=775,fmode=664"]

geerlingguy:
I think I may just set up a Wiki page discussing NFS and Synced Folder troubleshooting, and we can work on adding relevant examples/fixes in there. It seems like half or more of the issues reported so far on this project have to do with synced folders.

[...]

PR #872, opened 2016-08-18
If a synced folder's `owner` or `group` is set in `config.yml`, use it.

meitar:
In some cases, [synced folders exhibit permissions issues as described in Drupal VM's documentation](http://docs.drupalvm.com/en/latest/extras/syncing-folders/#permissions-related-errors). The [suggested fix](https://github.com/geerlingguy/drupal-vm/issues/66#issuecomment-90184995) is currently to edit the `Vagrantfile` in such a way as to add `owner` and `group` options to Vagrant's mount operation. This forces end users to maintain a patched Vagrantfile that differs from the one under version control.

This patch addresses the issue by allowing an untracked `config.yml` file to set these synced folder options, instead. If no options for `owner` and `group` are explicitly defined, a default value of `vagrant` is set so as to conform to the [Vagrant default user conventions](https://www.vagrantup.com/docs/boxes/base.html#default-user-settings). As a result, if a user experiences this issue, they can write a `config.yml` section for synced folders that looks like the following while leaving the Drupal VM core `Vagrantfile` untouched:

``` yaml
vagrant_synced_folders:
  - local_path: "."
    destination: "/var/www/drupalvm"
    type: ""
    create: true
    owner: "www-data"
    group: "www-data"
```

With such a stanza in one's `config.yml` file, the executed `Vagrantfile` will be the equivalent of a line such as:

``` ruby
config.vm.synced_folder ".", "/var/www/drupalvm", owner: "www-data", group: "www-data", create: true
```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -107,7 +107,9 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
       rsync__args: ['--verbose', '--archive', '--delete', '-z', '--chmod=ugo=rwX'],
       id: synced_folder['id'],
       create: synced_folder.include?('create') ? synced_folder['create'] : false,
-      mount_options: synced_folder.include?('mount_options') ? synced_folder['mount_options'] : nil
+      mount_options: synced_folder.include?('mount_options') ? synced_folder['mount_options'] : nil,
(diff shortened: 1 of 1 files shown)


oxyc:
Not as clean but you can do this already using the [`options_override` option](http://docs.drupalvm.com/en/latest/extras/syncing-folders/#overriding-defaults)

PR #889, opened 2016-09-02
Document that Parallels requires paid version.

benjifisher:
See http://parallels.github.io/vagrant-parallels/ (bottom of the page).


geerlingguy:
Good idea—I didn't realize Parallels had editions that didn't work like this—it's been a few years since I last used Parallels!

Issue #442, opened 2016-02-17
ERROR! the role 'geerlingguy.repo-remi' was not found in ...


bergstar:
- OS X 10.11.3
- Latest drupal-vm-master (stock config files)
- Installed xcode
- virtualbox 5.0.14 r105127
- vagrant 1.8.1

- sudo easy_install pip
- sudo pip install ansible --quiet
- vagrant up

Got this error:
```
==> drupal1: Running provisioner: ansible...
    drupal1: Running ansible-playbook...
ERROR! the role 'geerlingguy.repo-remi' was not found in /Users/user/Downloads/drupal-vm1/provisioning/roles:/Users/user/Downloads/drupal-vm1/provisioning:/etc/ansible/roles

The error appears to have been in '/Users/user/Downloads/drupal-vm1/provisioning/playbook.yml': line 21, column 7, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

    # Essential roles.
    - { role: geerlingguy.repo-remi, when: ansible_os_family == 'RedHat' }
      ^ here

Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.
```

oxyc:
You missed one step before `vagrant up`:
```sh
sudo ansible-galaxy install -r provisioning/requirements.yml --force
```

mgifford:
The nfsd service does not appear to be running.
Starting the nfsd service
==> drupalvm: Mounting NFS shared folders...
==> drupalvm: Mounting shared folders...
    drupalvm: /vagrant => /Users/mikegifford/drupal-vm-master
==> drupalvm: Running provisioner: ansible...
    drupalvm: Running ansible-playbook...
ERROR! the role 'geerlingguy.repo-remi' was not found in /Users/mikegifford/drupal-vm-master/provisioning/roles:/Users/mikegifford/drupal-vm-master/provisioning:/etc/ansible/roles

The error appears to have been in '/Users/mikegifford/drupal-vm-master/provisioning/playbook.yml': line 22, column 7, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

    # Essential roles.
    - { role: geerlingguy.repo-remi, when: ansible_os_family == 'RedHat' }
      ^ here

Ansible failed to complete successfully. Any error output should be
visible above. Please fix these errors and try again.
[...]

PR #894, opened 2016-09-09
Adding ansible command to README

subhojit777:
- Seems like `ansible-galaxy install -r provisioning/requirements.yml` is mandatory to install the VM. Otherwise, it throws error as mentioned in #442


geerlingguy:
@subhojit777 - That should not be required if you're using Drupal VM 3.2.x (or I think 3.1.x) with Vagrant 1.8.2 or later. Vagrant will install the roles automatically.

PR #906, opened 2016-09-26
Change php-fpm port so that xdebug works out of the box

lucaslegoff:
Currently, the default config set php-fpm to listen on port 9000 and xdebug to listen on the same port.
When you turn on xdebug, it doesn't work out of the box and the variable to change its port is not visible in default.config.yml (php_xdebug_remote_port)

One of the ports, either php-fpm or xdebug, should be changed to avoid conflict in the default configuration.

I chose to change php-fpm because the xdebug port is set in all IDE to 9000 and if changed, it requires an extra step for the developer and needs to be documented. Changing php-fpm seems to give a better DX.

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -99,28 +99,28 @@ apache_vhosts:
     serveralias: "www.{{ drupal_domain }}"
     documentroot: "{{ drupal_core_path }}"
     extra_parameters: |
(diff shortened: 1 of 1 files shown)


geerlingguy:
@lucaslegoff - Since PHP-FPM on port 9000 is basically an industry-wide standard (I've almost never seen it on a different port, except in exotic circumstances, or when multiple instances are on one host), I'd rather not make this change.

It would be better to set a different default port for xdebug and also update the documentation to set things to point to it correctly.

Issue #870, opened 2016-08-17
Message: Psych::SyntaxError: (/<snip>/config.yml): did not find expected key while parsing a block mapping at line 4 column 1


caschbre:
Just downloaded the latest DrupalVM today, updated config.yml like I normally do and `$ vagrant up` gave me the following...

```
There was an error loading a Vagrantfile. The file being loaded
and the error message are shown below. This is usually caused by
a syntax error.

Path: /home/caschbre/virtualmachines/voyant/Vagrantfile
Line number: 0
Message: Psych::SyntaxError: (/<snip>/config.yml): did not find expected key while parsing a block mapping at line 4 column 1
```

caschbre:
Hmm... so it looks like my alteration to the cron setup in config.yml was the culprit.  Here's what I had changed.

from:
`job: "drush -r {{ drupal_core_path }} core-cron"`
to
`job: "drush -r {{ drupal_core_path }} elysia-cron"`

I'm guessing that's just not allowed?

thom8:
@caschbre it's most likely a YAML syntax error, YAML is very fussy about spacing.

geerlingguy:
@caschbre - Are you using tabs at all? YAML doesn't allow tabs to be used anywhere, and will do funny things if you have them (http://www.yaml.org/faq.html)

The job as formatted in your comment looks fine; I'm wondering if you might have your editor set to automatically force tabs. Also, make sure that spacing is correct for each config directive, e.g.

```yaml
drupalvm_cron_jobs: []
  - name: "Drupal Cron",
    minute: "*/30",
    job: "drush -r {{ drupal_core_path }} core-cron"
```

(there should be four spaces before `minute` and `job`, two spaces before the `- name` line).

thom8:
YAML loves them spaces -- https://www.youtube.com/watch?v=SsoOG6ZeyUI

kamkejj:
I'm also getting the same error with the Cron Job section uncommented using the default settings
```
drupalvm_cron_jobs: []
  - name: "Drupal Cron",
    minute: "*/30",
    job: "drush -r {{ drupal_core_path }} core-cron"
```

Haven't figured out what the correct syntax should be.

oxyc:
You need to remove the empty set `[]` declaration

```yaml
drupalvm_cron_jobs:
  - name: "Drupal Cron",
    minute: "*/30",
    job: "drush -r {{ drupal_core_path }} core-cron"
```

[...]

PR #938, opened 2016-10-22
Followup #870: Fix invalid cron example by removing trailing commas

oxyc:


geerlingguy:
Thanks! (And good to have you back!)

oxyc:
Thanks, it's good to be back :)

PR #940, opened 2016-10-25
Mention .gitignore in Quick Start Guide

wturrell:
Thought it worth pointing out for people cloning the project that 
config.yml is in the .gitignore file – as I automatically assumed I'd
need a new branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ Notes:
 ### 2 - Build the Virtual Machine
 
   1. Download this project and put it wherever you want.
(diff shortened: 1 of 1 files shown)


geerlingguy:
I'd rather leave this out, as the two main use cases for Drupal VM—either use it as a Composer dependency (where this doesn't matter really, as Drupal VM will likely be in the vendor directory), or use it as a standalone 'thing'—don't require git knowledge/process for the config.yml file.

For those tracking the project with git, or adding it manually to a project under git version control, it becomes obvious after you start playing around with it (as you have :) ).

Issue #950, opened 2016-10-30
Vagrant cachier uses default sync folder type, which is not available


hussainweb:
When running `vagrant up` with vagrant cachier installed, I get this error:

> Vagrant was unable to mount VirtualBox shared folders. This is usually
> because the filesystem "vboxsf" is not available. This filesystem is
> made available via the VirtualBox Guest Additions and kernel module.
> Please verify that these guest additions are properly installed in the
> guest. This is not a bug in Vagrant and is usually caused by a faulty
> Vagrant box. For context, the command attempted was:
> 
> mount -t vboxsf -o uid=900,gid=900 vagrant-cache /tmp/vagrant-cache
> 
> The error output from the command was:
> 
> /sbin/mount.vboxsf: mounting failed with the error: No such device

I am using geerlingguy/ubuntu1404 box and apparently it does not have VirtualBox guest extensions installed (I get a message saying that earlier as well). I can live with that if cachier can use a different sharing type. Pull Request coming up with changes for that.

geerlingguy:
@hussainweb - I'm guessing you have the 1.0.3 (I think? maybe 1.0.4) version of that box, which has outdated guest additions. You should also make sure to install the `vbguest` plugin too: `vagrant plugin install vagrant-vbguest`, so the guest additions are always up to date.

I just published a new version of the box today, with Guest Additions for 5.1.8.

Closing this issue, will track the PR to still allow the customization to be set for cachier (see my comment there).

hussainweb:
I have the latest version of that box (1.1.3). I have used boxes with outdated guest extensions before and sharing still works. Here, it says the guest extensions are entirely missing. Maybe it was an older version of your box itself because the error just appeared recently, after I updated the boxes.

Thanks for the tip on plugin. I'll check it out.

hussainweb:
I checked the plugin and see that it installs VB Guest additions if not present. I guess you removed Guest additions from your boxes and relied on the plugin to install it. I remember not seeing this error earlier (but that could also be as I was not using cachier earlier, which means everything was on NFS).

Was this plugin required to use DrupalVM? I didn't see any documentation suggesting this.

geerlingguy:
> I guess you removed Guest additions from your boxes and relied on the plugin to install it.

[...]

PR #951, opened 2016-10-30
Issue #950: Allow sync folder type for cachier location to be configured

hussainweb:
This commit allows the sync folder sharing type for the location of vagrant cachier to be
configured. It also sets it to 'nfs' as default.


geerlingguy:
Let's just re-use the `vagrant_cachier_folder_default_type` variable for this (since it's a similar feature and I can't see a use case for having different types for the default vs cachier).

hussainweb:
I thought of that first but then thought you would say the opposite. :)

Anyway, I pushed a commit with the changes.

geerlingguy:
Merged, thanks!

PR #952, opened 2016-10-30
Install drupal/coder by default

hussainweb:
Drupal Coder is almost always required when developing with Drupal. It also implicitly installs codersniffer (as a dependency). It would be helpful if it was installed by default itself.

It is a very simple change that can be overridden in `config.yml`. Composer does the actual installation.

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -257,6 +257,7 @@ composer_home_owner: "{{ drupalvm_user }}"
 composer_home_group: "{{ drupalvm_user }}"
 composer_global_packages:
   - { name: hirak/prestissimo, release: '^0.3' }
+  - { name: drupal/coder }
 
 # Run specified scripts before or after VM is provisioned. Path is relative to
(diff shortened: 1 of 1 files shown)


geerlingguy:
> Drupal Coder is almost always required when developing with Drupal.

Never _required_, but often useful. I'd rather not include it by default though, because that would lead to a rabbit hole of including an entire CI system ala https://github.com/geerlingguy/drupalci-sonar-jenkins — CI is best left separate/up to the project using Drupal VM.

hussainweb:
I think "never" is a stretch. It is very useful to run style checks before sending it off for CI. I always have to remember to install this on a new box and set up links which breaks my flow. I thought having this by default would make it easier.

Thanks for considering the PR though. As long as it can be customized, I am fine. I am just setting it by default with a wrapper script in https://github.com/axelerant/setup-drupalvm which I now use to quickly setup drupalvm on new projects.

PR #955, opened 2016-11-01
Update mariadb.md

BrizzleRocker:
Adding postgresql failure workaround

### DIFF
--- a/docs/extras/mariadb.md
+++ b/docs/extras/mariadb.md
@@ -8,6 +8,8 @@ mysql_packages:
   - mariadb-server
   - python-mysqldb
 ```
+Note that you may also need to add the following line in the event of postgresql failures:
+`postgresql_packages: []`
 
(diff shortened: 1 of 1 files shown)


docs/extras/mariadb.md
```diff
@@ -8,6 +8,8 @@ mysql_packages:
   - mariadb-server
   - python-mysqldb
 ```
+Note that you may also need to add the following line in the event of postgresql failures:
+`postgresql_packages: []`
```
geerlingguy: Is this the same issue/project as mentioned in https://github.com/geerlingguy/drupal-vm/issues/957?

There should really never be a case when `postgresql_packages` isn't defined... unless we're running into the issue in Ansible 2.1.2+ where it tries using certain included tasks' `when` conditions even if variables aren't yet defined...

In either case, this should never be required—let's keep the discussion in #957 and then open an upstream issue in the `postgres` role if required.

Issue #971, opened 2016-11-06
Failed to fetch PPA information (PHP 5.6 / Ubuntu 12.04).


fubarhouse:
Assuming we have no possibility of upgrading Ubuntu or PHP, what's the recommended course of action on the following error?

````
TASK [Add repository for PHP 5.6.] *********************************************
fatal: [20161107093340.loc.gbuild.net]: FAILED! => {"changed": false, "failed": true, "msg": "failed to fetch PPA information, error was: HTTP Error 404: Not Found"}
````

fubarhouse:
So it seems that because the PPA has been consolidated into `https://launchpad.net/~ondrej/+archive/ubuntu/php`, package names have simply changed.

I propose that PHP 5.5 repository changes to this one, and all relevant package names listed on the documentation change to their corresponding package name. Notably php-fpm and php-memcached, but I'm yet to go through the list thoroughly.

I'm yet to do more testing, but it simply updating the documentation should do the trick.

thom8:
The repo referenced here has now been deprecated -- https://github.com/geerlingguy/drupal-vm/blob/master/provisioning/tasks/init-debian.yml#L24 / [ppa:ondrej/php5](https://launchpad.net/~ondrej/+archive/ubuntu/ppa/ubuntu/php5)

[ppa:ondrej/php](https://launchpad.net/~ondrej/+archive/ubuntu/php) supports `5.5`, `5.6` & `7.x`

fubarhouse:
In that case, could we perhaps update the documentation to make the package names or variable references refer to valid packages?

Also the link to PPA php5 you sent, I still received a 404...

Could somebody please verify that php5 works, as I've had zero success with it today.

fubarhouse:
Doing some testing with the packages now, and it seems like the following will ensure PHP 7 gets installed as well as 5.6, but the system defaults to 7.0.
````
php5.6-apcu
````

````
vagrant@vagrant:~$ sudo apt-get install php5.6-apcu
Reading package lists... Done
Building dependency tree
Reading state information... Done
Note, selecting 'php-apcu' instead of 'php5.6-apcu'
php-apcu is already the newest version.
0 upgraded, 0 newly installed, 0 to remove and 30 not upgraded.
````

````
php-pear
````

````
vagrant@20161107194036:~$ sudo apt-get install php-pear
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following extra packages will be installed:
[...]

PR #972, opened 2016-11-07
Remove deprecated PPA ondrej/php5.

thom8:
PR for #971


geerlingguy:
Failure on CentOS 7 was due to:

```
http://cgit.drupalcode.org/drupal/plain/.eslintignore?h=8.2.2\" file could not be downloaded: failed to open stream: Connection timed out
```

Running it again.

PR #991, opened 2016-11-15
Remove duplicate integration tests

oxyc:
There's really no reason to test the extra utilities for the pgsql integration test as everything has already been tested in the default scenario.

I also removed the xhprof integration test so we don't have to define the `installed_extras` variable but can reuse it from `default.config.yml`. I imagine we're mostly interested in knowing that the virtual hosts are created in all scenarios and we already test that with adminer and PML.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,83 +44,84 @@ before_install:
   - 'gem install rubocop'
 
 script:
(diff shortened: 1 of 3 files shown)


geerlingguy:
Looks good, shaves off a couple minutes from the build.

PR #994, opened 2016-11-16
Fixes #847, Followup #848: Add docs on using vagrant-proxyconf

### LINKED ISSUE #847: Integrate with vagrant-proxyconf

oxyc:
The plugin already supports environment variables (`VAGRANT_HTTP_PROXY` etc) so I excluded those from @morsok original PR.

I keep piling docs into that _Overriding configurations_ section as I'm hesitant to create new topics 😄

### DIFF
--- a/docs/other/overriding-configurations.md
+++ b/docs/other/overriding-configurations.md
@@ -80,4 +80,29 @@ Add the `AWS_ACCESS_KEY_ID` and the `AWS_SECRET_ACCESS_KEY` environment variable
 
 Then run `vagrant up --provider=aws` to provision the instance.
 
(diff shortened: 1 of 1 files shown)


geerlingguy:
> I keep piling docs into that Overriding configurations section as I'm hesitant to create new topics 😄

That's okay with me for now... as the docs continue to evolve, we'll have to consider reorganizing again—I'd like to hold off on that until maybe the 4.0 era and also determine at that point if we want to stay on readthedocs.io (since their search is still down).

Issue #455, opened 2016-02-21
Add possibility of prod-specific config.yml file override


geerlingguy:
This issue is an offshoot of #84.

Basically, the instructions right now tell you to use one `config.yml` that's modified for production (with more secure variable settings)... but if you do that, then you can't have the same `config.yml` be useful for your local development environment in the same Drupal VM instance.

It's not the end of the world to have to clone two copies of Drupal VM—one for local, one for prod—but it would be more clean if we could allow some sort of `prod.config.yml`, using the same kind of idea @oxyc mentioned elsewhere about having a `local.config.yml` override.

So the hierarchy of config files would be `prod.config.yml` > `local.config.yml` > `config.yml`. Something like that. I haven't quite worked out the specifics, and maybe there's a better way to do this architecturally, but I wanted to split this issue off of #84 so I could at least get that completed.

oxyc:
How about adding an internal variable to `vars/main.yml`

```yaml
_include_configs:
  - local.config.yml
```

Iterating over it in the Vagrantfile and checking if the files exist before merging them into `vconfig`. This would make it possible to even reference the configs in the `example` directory. To pass these on to ansible it would rely on the `extra_vars` issue being fixed in Vagrant 1.8.2 though.

geerlingguy:
True... but we also need to make sure the configs are set in Ansible itself correctly, and in the prod `ansible-playbook` commands. I don't want to use Vagrant to deploy to DigitalOcean (or other environments), but rather use Ansible directly.

So we would need support inside the Vagrantfile, but we could then use the new dynamic include functionality in Ansible 2.x to see if the file exists, include it if so, then see if the next one exists, include it if so, etc.

Because of that, I'm thinking of waiting until at least 2.0.1 is out, maybe a little further (for more stabilization in 2.x), before making this change.

oxyc:
Ah, true.

oxyc:
Taking into account that we need production specific variables like you say, my approach won't do much good. So more brainstorming.

Another idea that we discussed somewhere before was ansible `tags`. We could have a `production` tag that toggles ansible trying to include either `production.config.yml` or `vm.config.yml` (`local.config.yml` should always be included in my opinion).

[...]

PR #997, opened 2016-11-16
Issue #455: Add docs on environment specific configs

oxyc:
A first draft of the functionality.

### DIFF
--- a/docs/other/production.md
+++ b/docs/other/production.md
@@ -1,0 +1,34 @@
+Drupal VM has _experimental_ support for deploying Drupal VM to a production environment. The security of your servers is _your_ responsibility.
+
+For a configuration example and instructions on how to build a Drupal environment with Drupal VM on DigitalOcean see the [`examples/prod` directory and README](https://github.com/geerlingguy/drupal-vm/tree/master/examples/prod).
+
+## Production specific overrides.
+
+Drupal VM supports loading configuration files depending on the environment variable `DRUPALVM_ENV` and using this feature you can have different configurations between development and production environments.
+
+```sh
+# Loads vagrant.config.yml if available (default).
+vagrant provision
+
+# Loads prod.config.yml if available.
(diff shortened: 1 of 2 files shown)


oxyc:
I avoided giving specific suggestions as I'm not entirely sure what workflow we should be recommending.

Personally I would not entirely trust the environment variable and would instead have the security hardening configurations in `config.yml` while I override them with development tools in `vagrant.config.yml`. This way a production environment will never be provisioned with development tools even if I make a typo in the command.

geerlingguy:
That's... a good point. Do we want to update the docs and recommend cloning `default.config.yml` to `vagrant.config.yml` instead of `config.yml`? Then in the docs we can say "if you consider using Drupal VM for production, you should set the secure overrides in `config.yml` and customize your local environment using `vagrant.config.yml` and production-specific overrides in `prod.config.yml`... something along those lines?

oxyc:
> Do we want to update the docs and recommend cloning default.config.yml to vagrant.config.yml instead of config.yml

You mean in README, all docs, etc?

geerlingguy:
@oxyc - Yeah, maybe... though it might be a bit heavy handed. If Drupal VM becomes used more for multiple environment configuration (e.g. prod, stage, local), then using config this way is a little more correct.

But for the majority use case, currently just mentioning `config.yml` seems to be the simplest/least confusing. So I guess I just convinced myself to _not_ recommend changing things up right now :D

geerlingguy:
@oxyc - Can you also update the README in the prod directory to point to the docs page instead of having duplicated info? Just have the note about 'security is your responsibility', then direct people to the docs URL for more information.

oxyc:
@geerlingguy you want me to move over all the docs from the README?

geerlingguy:
Yeah.

geerlingguy:
This is fine for now... we can iterate on it later :)

PR #998, opened 2016-11-16
Fix ReadTheDocs search feature

oxyc:
I've only been able to test this by pasting it into the console and that does work.

Fix adapted from https://github.com/rtfd/readthedocs.org/issues/1088#issuecomment-224715045


geerlingguy:
Let's try it out!

geerlingguy:
Looks like it worked for Kalabox, too.

oxyc:
Yey!

PR #1004, opened 2016-11-20
Add php_pgsql_package note to PHP 5.6 docs

oxyc:


geerlingguy:
Ah, the wonderful aspects of supporting like 96 different software packages :D

PR #1037, opened 2016-12-06
1036: update acquia.config.yml example with correct PHP 5.6 config.

timcosgrove:


geerlingguy:
@timcosgrove - Good catch; I hadn't updated this in a while...

timcosgrove:
Thanks for DVM! Use it all day long.

PR #1042, opened 2016-12-09
Allow for extra YAML config files to be loaded

oxyc:
The ansible tasks aren't as clean as I would like, but I couldn't come up with anything else.

This would support things like:

```
extra_config_files:
  - local.config.yml
  - configs/*.yml
  - configs/php/php56.yml
  - ../vendor/company/drupalvm-configs/varnish.yml
```

Thoughts?

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -36,15 +36,24 @@ def walk(obj, &fn)
   end
 end
 
+def load_config_file(config_file, force = false)
+  config = YAML.load_file(config_file) if force || File.exist?(config_file)
+  config or Hash.new
+end
+
 require 'yaml'
 # Load default VM configurations.
-vconfig = YAML.load_file("#{host_drupalvm_dir}/default.config.yml")
-# Use optional config.yml and local.config.yml for configuration overrides.
-['config.yml', 'local.config.yml', "#{drupalvm_env}.config.yml"].each do |config_file|
-  if File.exist?("#{host_config_dir}/#{config_file}")
-    vconfig.merge!(YAML.load_file("#{host_config_dir}/#{config_file}"))
+vconfig = load_config_file("#{host_drupalvm_dir}/default.config.yml", true)
+# Use optional config.yml for configuration overrides.
+vconfig.merge!(load_config_file("#{host_config_dir}/config.yml"))
+# Load user defined configuration files.
+vconfig['extra_config_files'].each do |extra_config_file|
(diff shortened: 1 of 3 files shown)


oxyc:
With Ansible 2.2 something like this would also work:

```yaml
- name: Include extra config files.
   include_vars:
     dir: "{{ config_dir }}"
     files_matching: "{{ item }}"
     depth: 1
     with_flattened:
       - "{{ extra_config_files }}"
       - "{{ lookup('env', 'DRUPALVM_ENV')|default(drupalvm_env, true)|default(ansible_env.DRUPALVM_ENV)|default(omit) }}.config.yml"
```

`files_matching` is actually a regex and not a glob. I can't even figure out how to match things... A simpler way that removes globbing possibilty is:

```yaml
- name: Include extra config files.
   include_vars:
     file: "{{ config_dir }}/{{ item }}"
     with_flattened:
       - config.yml
       - "{{ extra_config_files }}"
       - "{{ lookup('env', 'DRUPALVM_ENV')|default(drupalvm_env, true)|default(ansible_env.DRUPALVM_ENV)|default(omit) }}.config.yml"
```

oxyc:
Something worth mentioning with the 2.2 solutions is that you cannot override `extra_config_files` as they are included one by one, not evaluated in advanced. Ie you cant remove `local.config.yml`.

geerlingguy:
Hmm... it just seems like we're starting to get too far into the territory of _too_ much power being exposed. With a few predefined files, it's more obvious that there's a hierarchy, like:

```
default.config.yml
config.yml
local.config.yml
```

But maybe it's just because that's what I'm used to. I'd like to hear from anyone else—does the current structure end up getting in the way often enough that adding in this functionality (and additional code complexity) help? Basically, for others—would this be worth it?

For _myself_, I haven't felt any pain whatsoever after we added the ability for optional config, and then optional local.config on top of that...

oxyc:
Possible. My primary use case is having a company specific composer dev-dependency with shared drupal-vm config files. So that I can just do a `composer update` rather than editing a dozen repos.

oxyc:
The solution isn't as nice as I would have wanted, and there doesn't seem to be anyone requesting these sort of features except me :) I'll make do with a post-update script keeping configs up to date.

geerlingguy:
@oxyc - No problem, it's not something I'm going to put a blanket 'no' on... but it's something I want to see more than one person request before I add it (since it does add to the already-slightly-more-complex-than-I-like level complexity surrounding where variables come from).

PR #1049, opened 2016-12-13
Add vagrant-cachier composer caching for root user

kentr:
### Problem / Motivation

Some component of the composer-based provisioning is running as `root` and using `/root/.composer/...` on the guest machine.

While `Vagrantfile` enables `vagrant-cachier` caching for the `vagrant` user (and the respective composer cache in `/home/vagrant/.composer/cache`), the `root` user doesn't have the same.

I suspect the Drush installation is doing this, but it probably doesn't matter.  In the bigger picture, I think it's likely that some tasks will run `composer install` as `root` in the course of a project based on DrupalVM.

### Proposed solution

Add `/root/.composer/cache` as a `vagrant-cachier` cache.

This PR does that.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -179,6 +179,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     config.cache.enable :apt
     # Cache the composer directory.
     config.cache.enable :generic, cache_dir: '/home/vagrant/.composer/cache'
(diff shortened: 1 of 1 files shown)


kentr:
Unsure why the Travis build is failing.  

Looking at the details, it appears to be stalling at the task `Ensure the firewall is enabled and will start on boot.`, which should be unrelated to this PR.

kentr:
Needs some work.

PR #1061, opened 2016-12-20
Allow user defined post-provision-tasks to use tags while drupal vm does not

oxyc:
At the moment Drupal VM does not use ansible tags out of the box, but I wanted to use it for my own custom post-provision tasks. By default all the pre/post-provision tasks are excluded as the include statements are without tags.

This PR fixes the issue by delegating the decision of tags to the post-provision scripts. Rather than setting the tag to `user-provision` or something similar, I think they should be set it to `always` so that users can take full advantage of tags. http://docs.ansible.com/ansible/playbooks_tags.html#special-tags

```sh
DRUPALVM_ANSIBLE_ARGS='--tags=foobar' vagrant provision
```

Currently I'm only tagging the config includes and the post/pre-provision tasks. But maybe we should also tag _Include OS-specific variables._ and such?

### DIFF
--- a/provisioning/playbook.yml
+++ b/provisioning/playbook.yml
@@ -20,12 +20,14 @@
       set_fact:
         config_dir: "{{ playbook_dir }}/.."
       when: config_dir is not defined
+      tags: ['always']
 
     - include_vars: "{{ item }}"
       with_fileglob:
         - "{{ config_dir }}/config.yml"
         - "{{ config_dir }}/local.config.yml"
         - "{{ config_dir }}/{{ lookup('env', 'DRUPALVM_ENV')|default(drupalvm_env, true)|default(ansible_env.DRUPALVM_ENV)|default(omit) }}.config.yml"
+      tags: ['always']
 
     - include: tasks/init-debian.yml
       when: ansible_os_family == 'Debian'


oxyc:
Nevermind, that didn't work.

geerlingguy:
@oxyc - What's not working? Did you add a tag to a custom task, then set the arg in the vagrant command, and it didn't use just that tag?

geerlingguy:
@oxyc - Also note that tags should always be in an array (I'm 99% sure)—so, like `tags: ['always']`.

oxyc:
If I tag the `include` task with `always`, then all the tasks within that file inherit the tag. Thus I can't limit the tasks that actually run :/ I didn't know that's how tag inheritance worked http://docs.ansible.com/ansible/playbooks_tags.html#tag-inheritance

I was hoping I could somehow read the included file, and only execute the tasks within that file that had a particular tag. Seems that's not possible as `--tags=foo,bar` means the union of `foo` and `bar`, rather than the intersection.

oxyc:
@geerlingguy if you have any ideas on how to allow this, I'm very interested :)

geerlingguy:
Hmm...

geerlingguy:
That `always` behavior sounds more like a bug in Ansible.

oxyc:
I'll keep this open as a reminder to take a look at the ansible issue queue.

oxyc:
It does work, I was just being an idiot. We only need to tag the tasks that allow `config.yml` to be loaded. After that ansible does check the tags for each task within the included files.

Oh, and `tags` does not need to be a list. If you prefer a list, I can fix it.

geerlingguy:
> Oh, and tags does not need to be a list. If you prefer a list, I can fix it.

Ah yes... I just always do a list in case I need to modify it. It just feels _more correct_ to me... I'm one of those insane people that dictates PHP functions should only return one and only one type (e.g. empty array, or array... not null, array, FALSE, etc.). Can you change it?

oxyc:
There you go!

PR #1062, opened 2016-12-20
Allow post-provision tasks to use the item variable

oxyc:
Currently, if you use a loop inside a post/pre-provision taskfile ansible will complain if you use the `item` variable.

 > [WARNING]: The loop variable 'item' is already in use. You should set the
`loop_var` value in the `loop_control` option for the task to something else to
avoid variable collisions and unexpected behavior.

This makes DX a bit nicer by allowing the users to use that variable.

### DIFF
--- a/provisioning/playbook.yml
+++ b/provisioning/playbook.yml
@@ -39,7 +39,9 @@
       with_items: "{{ pre_provision_scripts|default([]) }}"
 
     - name: Run configured pre-provision ansible task files.
-      include: "{{ item }}"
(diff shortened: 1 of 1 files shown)


geerlingguy:
Interesting... I haven't hit this, but haven't done any loops in pre/post task files, so I'll take your word for it. Looks safe to me.

oxyc:
For reference http://docs.ansible.com/ansible/playbooks_loops.html#loop-control

PR #1043, opened 2016-12-09
Followup #667, #714: Make it easier to switch between PHP versions

oxyc:
Doesn't work but I'm pushing to see if I can get travis to help out.

Sorry, I made a duplicate PR for this because rebasing caused a headache.

Issue #1031, opened 2016-12-02
PHP 7.1 - Timeline for switching to default


geerlingguy:
[PHP 7.1.0](https://secure.php.net/releases/7_1_0.php) rolled in just in time for the new year!

It may take a short time before it's available through Remi's repos on RHEL, and Ondrej's PPA on Ubuntu, so this won't happen right away, but I would like to add support for it ASAP, and then decide when to switch the default from 7.0 to 7.1.

Also, we'd probably need to convert the existing [PHP 5.6 docs page](http://docs.drupalvm.com/en/latest/other/php-56/) over to something like "Using older versions of PHP", and list instructions for 5.6 and 7.0.

As part of this process, I'd like Ubuntu to default to using Ondrej's PPA no matter what, so we could technically swap out PHP 5.6, 7.0, and 7.1 on Ubuntu 16.04 (the current setup only allows PHP 7.0 on 16.04).

oxyc:
Tests are passing for #1043 (I'm building 5.6, 7.0 and 7.1)! https://travis-ci.org/geerlingguy/drupal-vm/builds/182736056

Still needs manual testing though.

geerlingguy:
@oxyc - Awesome, glad you could take up the gauntlet on this. I hope to get some time to test (now that I have a retina MBP for work, it has the beef to be able to do things without bogging down everything else on the system), and then maybe we can sneak this into the 3.6.0 release as well (since we're adding a few other slightly-breaking changes).

But what's your opinion of making this the 4.x release? We have enough little things, and this would be a really slick feature to headline a 4.0.

oxyc:
Defaulting to 7.1 does sounds like a nice Drupal VM 4.0 👍 https://github.com/geerlingguy/drupal-vm/releases/tag/3.0.0

oxyc:
XDebug does not work with 7.1 https://github.com/oerdnj/deb.sury.org/issues/444

geerlingguy:
@oxyc - Good point about drush and failed builds. Switching major versions would do a lot to make sure people at least read through the 4.0 release notes!

I think I'll keep the default at 7.0—for various reasons (some Drupal bits and bobs still have edge case issues with 7.1 I'm hearing), but announce that 7.1 support is here. Hopefully within a few months we'll be able to find a way to get xhprof, xdebug, etc. all working splendidly with 7.1 as well!

oxyc:
[...]

PR #1066, opened 2016-12-21
Install XHProf by default, reverts a6683f5

oxyc:
Just realized we add a XHProf vhost while we don't actually have xhprof enabled by default. https://github.com/geerlingguy/drupal-vm/commit/157b0edc1e247ffa30b375846069364c30cb1e02 removed XHProf from the default `installed_extras`, I assume it was because we didn't switch the package automatically before. With https://github.com/geerlingguy/drupal-vm/pull/1043 we set the correct package automatically so this should be safe to re-enable again.

That being said, we might also consider removing the vhost as the future of XHProf on PHP 7.1 looks pretty dark https://github.com/geerlingguy/drupal-vm/issues/1031. I think this PR is the correct fix at the moment, but maybe it's not worth changing the default `installed_extras` as we're probably switching to 7.1 as the default soon anyways.

Edit: Currently `service apache2 restart` outputs:

> AH00112: Warning: DocumentRoot [/usr/share/php/xhprof_html] does not exist

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -123,6 +123,14 @@ script:
 
   - >
     if [ $TEST_INSTALLED_EXTRAS = true ]; then
+      docker exec ${container_id} curl -s --header Host:xhprof.${HOSTNAME} localhost \
+      | grep -q '<title>XHProf' \
+      && (echo 'XHProf install pass' && exit 0) \
+      || (echo 'XHProf install fail' && exit 1)
+    fi
+
+  - >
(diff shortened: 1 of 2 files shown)


geerlingguy:
I'd rather leave it disabled for now, and live with the apache warning. Same thing happens if some of the other subdomains aren't in use (e.g. if you customize installed_extras and remove some of the other things).

Until there's a clearer path to 7.1, of course.

oxyc:
Makes sense, it took me this long to notice that warning.

PR #1070, opened 2016-12-21
Configure hostname for environments other than VMs

oxyc:
None of these tasks should issue `changed` state when provisioning for the first time–They should do the exact same thing as Vagrant does. I still haven't made sure of that.

- Debian uses the short hostname in `/etc/hostname` while RedHat uses the FQDN.
- `/etc/mailname` is a debian thing

### DIFF
--- a/provisioning/tasks/hostname.yml
+++ b/provisioning/tasks/hostname.yml
@@ -1,0 +1,26 @@
+- name: Define fully qualified domain name.
+  set_fact:
+    hostname_fqdn: "{{ vagrant_hostname }}"
+  when: hostname_fqdn is undefined
+
+- name: Define short hostname.
+  set_fact:
+    hostname_short: "{{ hostname_fqdn|regex_replace('^([^.]+).*$', '\\1') }}"
+  when: hostname_short is undefined
+
+- name: Add hostname to /etc/hosts.
+  lineinfile:
+    dest: /etc/hosts
+    regexp: '.*\t{{ hostname_short }}$'
+    line: "127.0.0.1\t{{ hostname_fqdn }}\t{{ hostname_short }}"
+    state: present
+
+- name: Configure hostname.
+  copy:
+    content: "{{ (ansible_os_family == 'Debian') | ternary(hostname_short, hostname_fqdn) }}\n"
+    dest: /etc/hostname
+  register: set_hostname
+
+- name: Set the hostname for current session.
+  shell: hostname --file /etc/hostname
+  when: set_hostname.changed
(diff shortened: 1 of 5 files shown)


oxyc:
> An exception occurred during task execution. To see the full traceback, use -vvv. The error was: NameError: global name 'exception' is not defined

https://github.com/ansible/ansible/issues/18628

oxyc:
I'm not setting the containers hostname as that would cause this kind of `/etc/hosts`

```
docker exec --tty ${container_id} env TERM=xterm cat /etc/hosts
127.0.0.1	localhost
::1	localhost ip6-localhost ip6-loopback
fe00::0	ip6-localnet
ff00::0	ip6-mcastprefix
ff02::1	ip6-allnodes
ff02::2	ip6-allrouters
127.0.0.1	drupalvm.dev	drupalvm
172.17.0.2	drupalvm
```

With unsafe_writes (only enabled on docker) this works. I dont really understand unsafe writes though... @geerlingguy should we skip the tasks entirely for docker or is the `unsafe_write` option okay for docker?

oxyc:
@geerlingguy this gets my 👍, except I cant test Centos7 because of some version/box/whatever reason causing my VM not to boot. I dont see why it wouldnt work when it works for centos6 though

geerlingguy:
Yeah, the CentOS 7 thing is a bug in Vagrant 1.9.0/1.9.1: https://github.com/mitchellh/vagrant/issues/8096#issuecomment-268562601

provisioning/tasks/hostname.yml
```diff
@@ -0,0 +1,28 @@
...
+    regexp: '.*\t{{ hostname_short }}$'
+    line: "127.0.0.1\t{{ hostname_fqdn }}\t{{ hostname_short }}"
+    state: present
+    unsafe_writes: "{{ (ansible_virtualization_type == 'docker')|ternary('yes', 'no')|bool }}"
+
+- name: Configure hostname.
+  copy:
+    content: "{{ (ansible_os_family == 'Debian') | ternary(hostname_short, hostname_fqdn) }}\n"
+    dest: /etc/hostname
+    unsafe_writes: "{{ (ansible_virtualization_type == 'docker')|ternary('yes', 'no')|bool }}"
```
geerlingguy: Yeesh!

oxyc: See https://github.com/ansible/ansible/issues/13981, probably among many others.

geerlingguy: Also note that this will be the 2nd thing (besides `systemd` use in the Ansible role) that explicitly requires Ansible 2.2 (https://github.com/ansible/ansible/issues/13981#issuecomment-247031451). I might want to add a more detailed Ansible version checker thingie, because the current one only checks once the entire playbook is loaded and validated—meaning I get a lot of issue reports for people running 2.1 and hitting the systemd error.

oxyc: Should we do it in `Vagrantfile` already?

geerlingguy: @oxyc - I was thinking about that... if we do the version compare inside the check for `which ansible`, then I think I'd be game for that. It would prevent a lot of issues and pings on Twitter and chat, for sure.

PR #1071, opened 2016-12-21
Add docs on using HTTPS with Nginx

oxyc:
This is using @joestewart example https://github.com/geerlingguy/drupal-vm/issues/754#issuecomment-242744098, which worked when I tested it on Ubuntu 16.04.

---

I thought the magic jinja trick might be useful so I added a tips & tricks section. Prefer if I leave it out?

### DIFF
--- a/docs/extras/ssl.md
+++ b/docs/extras/ssl.md
@@ -2,7 +2,13 @@ To enable SSL support for you virtual hosts you first need a certificate file. Y
 
     openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout example.key -out example.crt
 
-Place the files in your project directory and add the following to your `config.yml`.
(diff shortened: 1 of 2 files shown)


oxyc:
> The directives ssl_protocols and ssl_ciphers can be used to limit connections to include only the strong versions and ciphers of SSL/TLS. By default nginx uses “ssl_protocols TLSv1 TLSv1.1 TLSv1.2” and “ssl_ciphers HIGH:!aNULL:!MD5”, so configuring them explicitly is generally not needed. Note that default values of these directives were changed several times.

Maybe just exclude those then.

geerlingguy:
At some point I'm going to work on making the Nginx and Apache roles a little more consistent with the idea of vhosts/server blocks... but until then, this is helpful.

Issue #1076, opened 2016-12-22
undefined symbol: zend_sort


sbubaron:
### Issue Type

 - Bug Report / Support Request

### Your Environment
```
Vagrant 1.9.1
bash: vboxmanage: command not found
VirtualBox
bash: ansible: command not found

VirtualBox: Version 5.0.30 r112061
```

### Your OS
 - **Windows** 10

### Full console output

N/A

### Summary
After switching from php7.0 to 5.6 and running vagrant provision the following error would pop up in the console randomly while running drush commands:

```
Failed loading /usr/lib/php5/modules/xdebug.so:  /usr/lib/php5/modules/xdebug.so: undefined symbol: zend_sort
```

vagrant destroy and vagrant up seemed to fix it -- but oxyc requested I open a new issue.

oxyc:
Found the bug. It's because https://github.com/geerlingguy/ansible-role-php-xdebug/blob/master/tasks/main.yml#L33:L42 (and xhprof, uploadprogress) do not re-compile as there's no name change. It simply copies the old version in place.

> TASK [geerlingguy.php-xdebug : Build Xdebug.] **********************************
ok: [drupalvm_project] => (item=phpize)
ok: [drupalvm_project] => (item=./configure)
ok: [drupalvm_project] => (item=make)

For reference: https://github.com/geerlingguy/drupal-vm/pull/1043#issuecomment-268879545

oxyc:
ping @thom8 @geerlingguy maybe fix this in the upstream roles by changing it to something like:

```yaml
- name: Build Xdebug.
  shell: >
    {{ item }}
    chdir={{ workspace }}/xdebug-{{ php_xdebug_version }}-php{{ php_version }}
    creates={{ workspace }}/xdebug-{{ php_xdebug_version }}-php{{ php_version }}/modules/xdebug.so
  with_items:
    - phpize
    - ./configure
    - make
  notify: restart webserver
```

thom8:
@oxyc as a workaround we make `workspace` PHP version specific -- https://github.com/beetboxvm/beetbox/blob/master/provisioning/ansible/config/default.config.yml#L282

oxyc:
I just realized my suggestion didn't work as it was supposed to be the name of the downloaded file. 

Hmm sucks to recompile stuff such as solr etc because of a php version change though.

thom8:
agreed it's not ideal but it's only an issue when switching php versions on an existing VM.

[...]

PR #1079, opened 2016-12-23
Fix PHP modules not re-compiling on PHP version changes

oxyc:
Would be nicer if this could be done upstream in the php extension roles though... Now stuff such as solr is rebuilt as well.

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -314,7 +314,6 @@ solr_xmx: "128M"
 selenium_version: 2.53.0
 
 # Other configuration.
-workspace: "/root/php{{ php_version }}"
 dashboard_install_dir: /var/www/dashboard
 known_hosts_path: ~/.ssh/known_hosts
 #ssh_home: "{{ drupal_core_path }}"
--- a/default.config.yml
+++ b/default.config.yml
@@ -314,6 +314,7 @@ solr_xmx: "128M"
 selenium_version: 2.53.0
 
 # Other configuration.
+workspace: "/root/php{{ php_version }}"
 dashboard_install_dir: /var/www/dashboard
 known_hosts_path: ~/.ssh/known_hosts
 #ssh_home: "{{ drupal_core_path }}"
--- a/provisioning/playbook.yml
+++ b/provisioning/playbook.yml
@@ -58,6 +58,12 @@
         xhprof_download_folder_name: xhprof-master
       when: php_version == '5.5' or php_version == '5.6'
 
+    - name: Ensure PHP version -specific workspace directory exists.
+      file:
+        path: "/root/php{{ php_version }}"
+        state: directory
+        mode: 0775
+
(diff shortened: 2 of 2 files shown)


geerlingguy:
Hmm... I'd be willing to do things in the `php` role, but it's a lot more effort to try to change _all_ the `php-*` roles.

geerlingguy:
Especially since all the annoying changes are for Debian only (RHEL repos use the same package names, paths, etc. and unless you use Software Collections you just install one version at a time).

oxyc:
This one affects RHEL as well. The extension never recompiles even if the PHP version changes. It's just moved back into place.

oxyc:
Also, it's pretty rare that people switch PHP versions. Re-building solr/daemonize/etc isnt ideal, but it barely increases the build time anyways..

oxyc:
Nevermind, they are re-compiled, just not moved into place :) @joestewart https://github.com/geerlingguy/drupal-vm/issues/1076#issuecomment-268988244

geerlingguy:
This would still be needed after the upstream fix in the Xdebug role, correct? And would we need to fix anything in the XHProf role too?

oxyc:
@geerlingguy actually this is the only PR that is required for the fix. This is the solution suggested by @thom8 in https://github.com/geerlingguy/ansible-role-php-xdebug/pull/41#issuecomment-269059900.

Unfortunately I don't have time to test it and I'll be offline until Monday. 

In addition to this @joestewart (https://github.com/geerlingguy/drupal-vm/issues/1076#issuecomment-268988244) mentioned that the XDebug extension doesn't upgrade when you change the version (the extension version), this is a separate issue though. https://github.com/geerlingguy/ansible-role-php-xdebug/pull/41 has his fix rebased but bundled with some other commits that should not be merged if we go with assigning the `workspace` here in the playbook instead.

provisioning/tasks/init-debian.yml
```diff
@@ -55,15 +55,6 @@
...
 
-- name: Purge PHP modules directory.
-  file:
-    path: "{{ item }}"
-    state: absent
-  when: php_purge.changed
-  with_items:
-    - "/usr/lib/php5/modules"
-    - "/usr/lib/php/modules"
-
```
oxyc: Actually, I think we need to test that memcached/redis extensions still work on version change without this. Unfortunately I wont have time to do that until monday.

geerlingguy:
@oxyc - I merged https://github.com/geerlingguy/ansible-role-php-xdebug/pull/42, so it might be easier to just pull in the updated version of that role (2.3.0).

geerlingguy:
Ah, true. Post-holiday confusion from having like 40 new issues across my roles :D

geerlingguy:
(Apparently everyone but me works on my roles over Christmas.)

Issue #1047, opened 2016-12-12
Update docs: Add DrupalConsole remote command execution instructions.


jmolivas:
### Issue Type

 - Documentation Report

### Summary

Update documentation add DrupalConsole remote command execution example

Drupal Console allows you to run commands on your local server but actually execute them on a remote server. Instructions to setup remote command execution.

Add a new sites file `~/.console/sites/drupalvm.yml` at host computer.
```
dev:
  root: /var/www/drupalvm/drupal
  host: 192.168.88.88
  user: vagrant
  password: vagrant
```
NOTE: This site file was created using the default configuration that ships with Drupal VM.

Execute from host machine using `--target` option.
```
$ bin/drupal --target=drupalvm.dev site:status

 System Info
 Drupal                    8.2.4
 Access to update.php      Protected
 Configuration files       Protected
 Cron maintenance tasks    Last run 2 days 9 hours ago
 Database system           MySQL, MariaDB, Percona Server, or equivalent
 Database system version   5.7.16-0ubuntu0.16.04.1-log
 Database updates          Up to date
 Drupal core update status No update data available
 Entity/field definitions  Up to date
 File system               Writable (public download method)
 GD library                2.1.1
 GD library PNG support    2.1.1
 Image toolkit             gd
 Node Access Permissions   Disabled
 PHP                       7.0.8-0ubuntu0.16.04.3 (more information)
 PHP extensions            Enabled
 PHP memory limit          192M
 PHP OPcode caching        Enabled
 Search index progress     100% (0 remaining)
 Trusted Host Settings     Not enabled
 Unicode library           PHP Mbstring Extension
 Update notifications      Enabled
 Upload progress           Not enabled
 Web server
 Hash salt                 n4X2BOJGCeSh6UfkyOmhssdyiofbrla_fBzIHfz1PiJ6GyiiFKSkG3r6YsSP84bwfLig3vvRhw
 Drupal Console            1.0.0-rc10

 Database connection
 Driver              mysql
 Host                localhost
 Database connection drupal
 Port
 Username            drupal
 Connection          mysql//drupal:drupal@localhost/drupal

 Themes
 theme_default bartik
 theme_admin   seven

 Directories
 Site root directory      /var/www/drupalvm/drupal/web
 Site temporary directory /tmp
 Default theme directory  /core/themes/bartik
 Admin theme directory    /core/themes/seven
```

geerlingguy:
[...]

PR #1085, opened 2016-12-29
Issue #1047: Add docs on Drupal Console remote command execution

oxyc:


oxyc:
Or we could just scaffold it automatically...

geerlingguy:
This is good enough for now... let's leave out the automatic generation until it's a little more stable—I'd rather wait until a while after the 1.x final release before committing to a format for the automated alias file generation :)

Issue #1055, opened 2016-12-16
Drupal core redirects fail on fresh installation


ba66e77:
### Issue Type

 - Bug Report / Support Request

### Your Environment

```
Vagrant 1.8.6
VirtualBox 5.1.6r110634
ansible 2.2.0.0
  config file =
  configured module search path = Default w/o overrides
```

### Your OS

 - **macOS** 10.12.1

### Full console output

*Curling path to test redirect*
```
$ curl -I http://d8sandbox.local/install.php
HTTP/1.1 404 Not Found
Date: Fri, 16 Dec 2016 22:27:13 GMT
Server: Apache/2.4.18 (Ubuntu)
X-Powered-By: PHP/7.0.14-2+deb.sury.org~xenial+1
Content-Type: text/html; charset=UTF-8
```

*Curl path to test clean urls are working*
```
$ curl -I http://d8sandbox.local/user/login
HTTP/1.1 200 OK
Date: Fri, 16 Dec 2016 22:30:52 GMT
Server: Apache/2.4.18 (Ubuntu)
X-Powered-By: PHP/7.0.14-2+deb.sury.org~xenial+1
Cache-Control: must-revalidate, no-cache, private
X-Drupal-Dynamic-Cache: MISS
X-UA-Compatible: IE=edge
Content-language: en
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
Expires: Sun, 19 Nov 1978 05:00:00 GMT
X-Generator: Drupal 8 (https://www.drupal.org)
X-Drupal-Cache: MISS
Content-Type: text/html; charset=UTF-8
```

*Verify mod_rewrite is loaded in apache*
```
vagrant@d8sandbox:~$ apache2ctl -M
AH00112: Warning: DocumentRoot [/usr/share/php/xhprof_html] does not exist
Loaded Modules:
...
 rewrite_module (shared)
...
```

### Summary
On a clean installation of DrupalVM, I'm seeing mod_rewrite redirects set by Drupal core not working as expected.  In .htaccess, Drupal 8 includes a redirect rule to reroute /install.php to /core/install.php.  As you can see from the above, .htaccess redirects (as an example) are not working even though clean URLs are working and mod_rewrite is loaded in the VM.

thom8:
See https://github.com/geerlingguy/drupal-vm/issues/945#issuecomment-258744112

ba66e77:
[...]

Issue #945, opened 2016-10-26
Drupal 8 not recognizing clean url support


shawnachieve:
Whenever I setup a new DrupalVM site for Drupal 8 (or try to reinstall an existing D8 site), Drupal always comes up and says that DrupalVM does not support clean URLs.

The actual message that shows up during the requirements check is:

> Your server is capable of using clean URLs, but it is not enabled. Using clean URLs gives an improved user experience and is recommended.

This message comes up whether I install through the browser going to http://mysite.local/core/install.php or via drush site-install.

I am able to install the site and everything works correctly if I click the "continue anyway" link on the install.php page, but there doesn't seem to be a way around this from drush.

Since I haven't seen anyone else posting about this issue I'm assuming it's something specific to my setup.

Any thoughts on what may be causing this and how to fix it?

Here is some info on my setup:
- DrupalVM 3.4.0 (checked out tag from github) running on Mac OS 10.12
  - I had this same problem with DrupalVM 3.2.x and Mac OS 10.11 as well.
- Drupal 8.1.10 
- Pretty much a stock config.yml file.  
  - The main differences are using the auto-network plugin with an IP setting of 0.0.0.0
  - Setting up the vagrant_synced_folders to use rsync.
  - Turned off the automatic make file and site install commands.

Thanks.

geerlingguy:
@shawnachieve - I've noticed the same thing. Not entirely sure why Drupal fails that initial check—my guess is a bug in Drupal core :)

It would be nice to figure out what's happening, but as it doesn't really cause any functional issues, I haven't ever had it as a priority.

shawnachieve:
I'm still working through what's going on, but so far what I can tell is that the following rewrite command in the .htaccess file is not triggering:

```
RewriteRule ^core/install.php core/install.php?rewrite=ok [QSA,L]
```

The above command is suppose to append a "rewrite" query parameter to the core/install.php URL.  If the install script doesn't see that rewrite parameter it assumes that mod_rewrite is disabled.

So far what I've seen with mod_rewrite debugging enabled is that this rewrite rule runs against everything but "core/install.php" (e.g. CSS files, theme files, etc.)  

Still trying to figure out the underlying cause though.

geerlingguy:
Weird... I wonder if somehow the ProxyPass setup is not proxying that particular rewrite correctly or something? Maybe it's an Apache bug?

thom8:
[...]

Issue #617, opened 2016-05-13
Apache/Nginx/PHP-FPM not handling index.html as index file.


sarahjean:
Working with the zen theme, which uses a static styleguide, and it looks like Apache is not loading index.html as an index file. 

To reproduce you can start up a fresh drupalvm instance, install drupal 8.2.x and the zen theme + components module. If you visit http://drupalvm.dev/themes/contrib/zen/STARTERKIT/styleguide/ you get a 404, but you can visit http://drupalvm.dev/themes/contrib/zen/STARTERKIT/styleguide/index.html directly

Have errors similar to http://serverfault.com/questions/450628/apache-2-4-php-fpm-proxypassmatch in /var/log/apache2/error.log.

derimagia:
Took a look at this for a bit to see exactly what was happening here and why it wasn't happening in nginx. 

This answer: http://serverfault.com/a/672969 (From the link you linked) seems to have a fix for apache and the reason FilesMatch/SetHandler wasn't used before is because of the fact that drupal-vm supported Apache 2.2 before which didn't support sockets. We should take a look if we can pull this out of every virtualhost so clean everything up. We may be able to remove the DirectoryIndex at the top of the file.

Nginx didn't have this issue because in apache the path is being hit by .htaccess and then the path is being redirected to index.php and then it's failing there. Nginx has a try_files clause which checks if the file (and directory) exist which then hits the index clause...

geerlingguy:
Taking this further, we might be able to figure out a way to use something like:

```
<FilesMatch \.php$>
   SetHandler "proxy:unix:/path/to/socket.sock|fcgi://unique-domain-name-string/"
 </FilesMatch>
```

Instead of the current `ProxyPassMatch` in a way that's slightly easier to maintain. Maybe... Another option is to try to figure out how to get ProxyPassMatch to work okay with `index.html`.

attheshow:
Same issue here. We're switching one of our sites from being Drupal based to become a more static site with an index.html but we're attempting to run it on our VM. Right now we can view it but we have to be sure that we explicitly type "/index.html" onto the end of the domain name.

geerlingguy:
@attheshow - Thanks for the +1! For now, the main fix is to use Nginx instead of Apache; though that's not ideal. I just haven't had the time to work on a more permanent fix for Apache.

attheshow:
[...]

Issue #876, opened 2016-08-19
PHP-FPM doesn't respond to .html paths, only .php


bhasselbeck:
When trying to reach an .html asset located in the Drupal filesystem for Drupal 8.  I'm receiving a File Not Found error.  Error is similar to https://forums.cpanel.net/threads/php-fpm-directory-index-index-html-file-not-found-problem.529701/

We should allow for config settings on config.yml that allows for phpfpm to listen to a configurable array of filetypes.

geerlingguy:
Yeah, similar issue with Apache here: https://github.com/geerlingguy/drupal-vm/issues/617

geerlingguy:
(Note that for my other infra, my Nginx server config does route .html files appropriately, so maybe there's a clue in there for a simple fix.)

geerlingguy:
Closing in favor of #617, merging to there.

oxyc:
Fixed in #1086.

PR #1086, opened 2016-12-29
Use SetHandler fixing mod_rewrite and DirectoryIndex issues

oxyc:
Adding the PR for travis as well as further review and discussion.

- [x] Both dashboard index.html and php_info.php work.
- [x] Default Drupal installation works
  _Status reports only complaint is that trusted hosts settings is disabled._
- [x] Ensure index.html files are used as `DirectoryIndex` #617
  _Added Zen and http://drupalvm.dev/themes/contrib/zen/STARTERKIT/styleguide/ displays `index.html`_
- [x] Fix: Drupal core redirects fail on fresh installation #1055
  > curl -I http://drupalvm.dev/install.php
HTTP/1.1 301 Moved Permanently
  > 
- [x] Drupal 8 not recognizing clean url support #945
  _Adding `install_site: false` and running the installation through `/install.php` adds the `rewrite=ok` GET parameter._

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -106,29 +106,38 @@ apache_vhosts:
     serveralias: "www.{{ drupal_domain }}"
     documentroot: "{{ drupal_core_path }}"
     extra_parameters: |
-          ProxyPassMatch ^/(.*\.php(/.*)?)$ "fcgi://127.0.0.1:9000{{ drupal_core_path }}"
+          <FilesMatch \.php$>
+            SetHandler "proxy:unix:{{ php_fpm_listen }}|fcgi://{{ drupal_domain }}/"
+          </FilesMatch>
 
   - servername: "adminer.{{ vagrant_hostname }}"
     documentroot: "{{ adminer_install_dir }}"
     extra_parameters: |
-          ProxyPassMatch ^/(.*\.php(/.*)?)$ "fcgi://127.0.0.1:9000{{ adminer_install_dir }}"
+          <FilesMatch \.php$>
+            SetHandler "proxy:unix:{{ php_fpm_listen }}|fcgi://adminer.{{ vagrant_hostname }}/"
+          </FilesMatch>
 
   - servername: "xhprof.{{ vagrant_hostname }}"
     documentroot: "{{ php_xhprof_html_dir }}"
     extra_parameters: |
-          ProxyPassMatch ^/(.*\.php(/.*)?)$ "fcgi://127.0.0.1:9000{{ php_xhprof_html_dir }}"
+          <FilesMatch \.php$>
+            SetHandler "proxy:unix:{{ php_fpm_listen }}|fcgi://xhprof.{{ vagrant_hostname }}/"
+          </FilesMatch>
 
   - servername: "pimpmylog.{{ vagrant_hostname }}"
     documentroot: "{{ pimpmylog_install_dir }}"
     extra_parameters: |
-          ProxyPassMatch ^/(.*\.php(/.*)?)$ "fcgi://127.0.0.1:9000{{ pimpmylog_install_dir }}"
+          <FilesMatch \.php$>
+            SetHandler "proxy:unix:{{ php_fpm_listen }}|fcgi://pimpmylog.{{ vagrant_hostname }}/"
+          </FilesMatch>
 
   - servername: "{{ vagrant_ip }}"
     serveralias: "dashboard.{{ vagrant_hostname }}"
     documentroot: "{{ dashboard_install_dir }}"
     extra_parameters: |
-          ProxyPassMatch ^/(.*\.php(/.*)?)$ "fcgi://127.0.0.1:9000{{ dashboard_install_dir }}"
-          DirectoryIndex index.html
+          <FilesMatch \.php$>
+            SetHandler "proxy:unix:{{ php_fpm_listen }}|fcgi://dashboard.{{ vagrant_hostname }}/"
+          </FilesMatch>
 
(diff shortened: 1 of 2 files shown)


provisioning/templates/nginx-vhost.conf.j2
```diff
@@ -34,7 +34,7 @@ server {
         fastcgi_intercept_errors on;
         fastcgi_read_timeout {{ php_max_execution_time }};
         include fastcgi_params;
-        fastcgi_pass {{ php_fpm_listen }};
+        fastcgi_pass unix:{{ php_fpm_listen }};
```
oxyc: Would be nice if we could support both unix and tcp sockets.

oxyc: We could also just stick with tcp sockets. Or are the performance benefits of unix sockets noticeable?

geerlingguy: If possible, I'd like to stick with TCP sockets for now... they are slightly more debuggable (in my experience), and the performance impact is negligible in most real-world use cases. Sockets are very slightly faster, but we're talking < 1ms difference in most Drupal usage.

default.config.yml
```diff
@@ -98,37 +98,36 @@ drupalvm_cron_jobs: []
 # this variable is 'true'.
 configure_drush_aliases: true
 
+apache_vhosts_extra_parameters: |
+    <FilesMatch \.php$>
+      SetHandler "proxy:fcgi://{{ php_fpm_listen }}"
+    </FilesMatch>
```
oxyc: Something like this would be possible. What do you think? Too confusing?

geerlingguy: I think that might be helpful... It's so verbose the way it's set up right now :/

geerlingguy: If you could just add a comment above the variable explaining this is to help configure the PHP-FPM connection to Apache... something like that would be adequate.

oxyc:
Bedtime, I'll test it more thoroughly tomorrow.

oxyc:
@geerlingguy any other scenarios that should be tested? All those issues are now fixed with this PR.

geerlingguy:
> P.S. We're running out of bug/feature issues 👍 

Haha, nice!

geerlingguy:
@oxyc - Does this fix the index.html index for Nginx as well?

oxyc:
Probably not, I thought that nginx worked. I'll provision an nginx box and test it.

oxyc:
Nginx:
- [ ] `/install.php` redirects to `core/install.php`
  _Does not work. You have to visit core/install.php yourself_
- [ ] `rewrite=ok` parameter wasn't added during the install step, but the installation didn't complain about URL rewriting either?
- [x] no complaints from _Status report_
- [ ] Both dashboard index.html and php_info.php work.
  _Erm, the php file didn't work. It was just downloaded. We've forgotten to add the `is_php` attribute on the nginx vhost. If we add it, `index.html` is not shown though._
- [ ] http://drupalvm.dev/themes/contrib/zen/STARTERKIT/styleguide/ displays index.html
  _Nope_

oxyc:
@geerlingguy should we add https://github.com/drupal/drupal/blob/8.3.x/.htaccess#L126 to nginx as well?

One key difference here is that on Apache, you cannot access random php files outside of `core/`. On nginx you can due to this redirect not existing.

geerlingguy:
@oxyc - I would... just because I have had a few people ask my why "install.php" doesn't work (a lot of old scripts/assumptions still use it there and 'just work' in D8 because of that silly redirect :P

oxyc:
This is how I would translate the apache rewrite to nginx.

Apache:
```apache
  RewriteCond %{REQUEST_URI} ^(.*)?/(install.php) [OR]
  RewriteCond %{REQUEST_URI} ^(.*)?/(rebuild.php)
  RewriteCond %{REQUEST_URI} !core
  RewriteRule ^ %1/core/%2 [L,QSA,R=301]
```

nginx:

```nginx
    # Redirect common PHP files to their new locations.
    location ~ ^((?!.*(?:core)).*)/(install.php|rebuild.php) {
         return 301 $scheme://$host$2/core/$1;
    }

# drupalvm.dev/install.php -> drupalvm.dev/core/install.php
# drupalvm.dev/foo/install.php -> drupalvm.dev/foo/core/install.php
```

The problem is we only need this for the drupal vhost, and we need it before passing the file on to fastcgi. `extra_parameters` doesn't work as it's after the fastcgi rule.

oxyc:
This still doesn't add the security measurements the htaccess does. https://github.com/drupal/drupal/blob/8.3.x/.htaccess#L141:L155

geerlingguy:
@oxyc - Yeah... and I don't think we can solve that problem (Drupal community applying security through regex in htaccess files) easily in this PR :)

So I'm okay if a few edge cases aren't working.

geerlingguy:
Otherwise... this looks awesome, and is a great improvement. For some users on really old Ubuntu 12 configurations or still using CentOS 6, this might be the straw that breaks the camel's back—but they can still use the old ProxyPassMatch method (if they desire) by overriding the new var.

Also, existing configs overriding the Vhosts should still work correctly after this merge, so no reason to do a major version bump. It's something that people can take their time to upgrade to (if they need to).

geerlingguy:
So... how many issues did this fix?

- https://github.com/geerlingguy/drupal-vm/issues/617
- https://github.com/geerlingguy/drupal-vm/issues/876
- https://github.com/geerlingguy/drupal-vm/issues/1055

PR #1090, opened 2016-12-29
Docs: Update examples to use SetHandler instead of ProxyPassMatch

oxyc:


oxyc:
Wait, forgot one.

geerlingguy:
G'ah! How did that comment pop in _after_ I clicked merge?

Issue #649, opened 2016-05-19
Make Drupal VM work with vagrant-lxc


nortmas:
Hi! First of all, Nice Job! Very helpful Project!

**I have two suggestions:**

1) Have you ever considered to add an ability for linux users to use https://github.com/fgrehm/vagrant-lxc , it works much faster then virtual box.
I forced it to work, but I faced with number of challenges.

**What I did:**

a) Added lines to Vagrantfile.

```
 # LXC.
  config.vm.provider :lxc do |lxc|
    # Same effect as 'customize ["modifyvm", :id, "--memory", "1024"]' for VirtualBox
    # 1lxc.customize 'cgroup.memory.limit_in_bytes', '1024M'
  end
```

b) config.yml

`vagrant_box: fgrehm/trusty64-lxc`

c) provisioning/tasks/init-debian.yml removed lines.

```
- name: Disable the ufw firewall (since we use a simple iptables firewall).
  service: name=ufw state=stopped
  when: ansible_distribution == "Ubuntu"
```

It would be great if linux users have such ability from the box.

2) Is it possible to add oh-my-zsh as an optional possibility? It's a popular and very helpful tool.

Thanks!

oxyc:
In the docs we have a section on extending Drupal VM with custom providers, and an [example using the `vagrant-aws` provider](http://docs.drupalvm.com/en/latest/other/overriding-configurations/#example-using-the-vagrant-aws-provider). You can use the same setup for lxc as well.

> c) provisioning/tasks/init-debian.yml removed lines.

Hm why did you have to remove the firewall rules? Maybe that's something we can improve.

> 2) Is it possible to add oh-my-zsh as an optional possibility? It's a popular and very helpful tool.

Yes you can use the `post_provision_scripts` variable in `config.yml` to run arbitrary shell scripts, eg for setting up zsh. You can read more about the scripts in the [docs](http://docs.drupalvm.com/en/latest/extras/scripts/#shell-scripts). There's also an [example script for configuring solr](https://github.com/geerlingguy/drupal-vm/blob/master/examples/scripts/configure-solr.sh), that you can use as a boilerplate for your own script.

_Note: if you're reading the docs about pre- and post provision scripts. The section on ansible task files is for the upcoming Drupal VM 3.0.0, and they're not available in 2.5.1_

geerlingguy:
Also, for `oh-my-zsh`, if it's just a package install that you need, you can add it to the list of `extra_packages`.

[...]

PR #1094, opened 2016-12-30
Issue #649: Add docs on how to use vagrant-lxc

oxyc:


oxyc:
Would be nice if we could detect if ufw/ip6tables/etc was actually functioning before calling them. Seems it's always some kernel requirements missing.

oxyc:
I reverted the first commit which defaulted the `drupalvm_disable_ufw_firewall` variable. Maybe not worth it as we have docs anyways...

Thought I'd let you have a say if you prefer defaulting it rebasing it away though.

Also these docs still need some testing.

oxyc:
@geerlingguy could you test these docs at some point when you're using your linux machine? I might just set up my old thinkpad for some testing purposes as well...

oxyc:
Fixed `modprobe` only accepting a single module.

oxyc:
Also there seems to be some sort of race condition going on.

From time to time I get:

> NFS requires a host-only network to be created. Please add a host-only network to the machine (with either DHCP or a static IP) for NFS to work.

If I re-run `vagrant destroy; vagrant up --provider=lxc` it eventually (90% of the time on the second try already) works.

geerlingguy:
@oxyc - I will try to do it this week. Just have to reboot the ol' ThinkPad into Fedora.

geerlingguy:
Still haven't found the time. I even had everything set up last week before a couple prod issues resulted in back-to-back hotfixes and no time :(

geerlingguy:
I think I'll merge the other PR for docs overhaul before incorporating this one. This week I've had a bunch of audio gear sitting on top my PC laptop, so I haven't had the chance to even open it up!

geerlingguy:
@oxyc can you update the PR for the new layout? I'm comfortable merging it in, then anyone who finds issues can open up follow-up tickets.

Issue #1093, opened 2016-12-30
#1090 SetHandler PR broke Ubuntu 14.04


oxyc:
### Issue Type

- Bug Report / Support Request

### Full console output

```
$ curl http://drupalvm.dev/
<?php

/**
 * @file
 * The PHP page that serves all page requests on a Drupal installation.
 *
 * All Drupal code is released under the GNU General Public License.
 * See COPYRIGHT.txt and LICENSE.txt files in the "core" directory.
 */

use Drupal\Core\DrupalKernel;
use Symfony\Component\HttpFoundation\Request;

$autoloader = require_once 'autoload.php';

$kernel = new DrupalKernel('prod', $autoloader);

$request = Request::createFromGlobals();
$response = $kernel->handle($request);
$response->send()
```

### Summary

We were too hasty :( The PR broke older Ubuntu versions (just tested on 14.04) which install Apache 2.4.7. Apparently SetHandler requires 2.4.9.

oxyc:
I guess we either revert the PR, or add ondrejs apache ppa.

geerlingguy:
@oxyc let's go with 2.4.9+.

Ugh, Ubuntu!!!

geerlingguy:
Fixed in #1093

geerlingguy:
Tagged in 4.1.1 https://github.com/geerlingguy/drupal-vm/releases/tag/4.1.1

geerlingguy:
Hmm... on one of my own projects, the page was loading plain text .php files, and when I logged in and checked, I got:

```
$ apache2 -v
Server version: Apache/2.4.7 (Ubuntu)
Server built:   Jul 15 2016 15:34:04
```

Do we need to add anything to the `geerlingguy.apache` role to support different repos for the Apache install? https://github.com/geerlingguy/ansible-role-apache/blob/master/tasks/setup-Debian.yml#L6

oxyc:
Shouldn't this cover it? https://github.com/geerlingguy/drupal-vm/blob/master/provisioning/tasks/init-debian.yml#L24:L28

geerlingguy:
One would think... Testing on another Mac.

oxyc:
> Server built:   Jul 15 2016 15:34:04

So the provision never triggered a re-install. We need `state=latest` rather than `installed`?

geerlingguy:
@oxyc ah, that could be it... maybe if that's not configurable, it _should_ be configurable in the role.

[...]

PR #1095, opened 2016-12-30
Issue #1093: Install Apache 2.4.9+ on Ubuntu 14.04

oxyc:


geerlingguy:
I thought we had an ubuntu1404 test running in Docker through Travis... maybe we need to add one back in—it looks like we're only testing 1604 and centos7. A lot of devs are still on 14.04 :(

oxyc:
Hmm 14.04 is failing on MySQL role https://travis-ci.org/geerlingguy/drupal-vm/jobs/187783415#L976

geerlingguy:
Hmm maybe so... but just needs: https://github.com/geerlingguy/ansible-role-mysql/blob/master/tests/test.yml I think.

geerlingguy:
@oxyc - I won't have time to add in a 14.04 test today, but I'll tag the new release so we get 4.1.x working again for 14.04 users.

PR #1105, opened 2017-01-05
Upgrade Apache packages on provision, fixing SetHandler usage on previously built Ubuntu 14.04 boxes

oxyc:


oxyc:
Now that I think about it. This is a bit dangerous, and at minimum needs a note in the production docs.

geerlingguy:
@oxyc - Can you add it as 'installed' (the default) in the prod config example? Otherwise, this feels pretty safe to me. Anyone not on 2.4 today better have a good reason for it!

Also, the default configuration changes will break your site loading if it's not updated as part of the general 4.x Drupal VM upgrade, so this will actually do a better job of fixing anyone upgrading from 3.x Drupal VM configs, assuming they haven't gone full-on custom on everything.

Issue #1133, opened 2017-01-20
Improve the Solr script in example folder.


jimmyko:
### Issue Type

 - Feature Idea

### Summary

In example folder, the solr script is the recommended way to configure solr in documentation. But this script only replace the conf folder of the default collection. My idea is that this script should create a new collection.

If someone wants to create multiple solr collection, it can be a starting point to customize in their environment.

oxyc:
This was merged in the above PR, thanks for your work @jimmyko!

geerlingguy:
Ah... keeping issue count below 25 :)

PR #1134, opened 2017-01-20
Create new solr collection instead of using the default one.

jimmyko:
This pull request has implemented the suggestion I raised in #1133

### DIFF
--- a/examples/scripts/configure-solr.sh
+++ b/examples/scripts/configure-solr.sh
@@ -8,14 +8,14 @@
 # of Apache Solr Search or Search API Solr.
 
 NAME="d8"
-SOLR_SETUP_COMPLETE_FILE="/etc/drupal_vm_solr_config_complete_"$NAME
+SOLR_SETUP_COMPLETE_FILE="/etc/drupal_vm_solr_config_complete_$NAME"
 
 # Search API Solr module.
 SOLR_DOWNLOAD="http://ftp.drupal.org/files/projects/search_api_solr-8.x-1.x-dev.tar.gz"
 SOLR_DOWNLOAD_DIR="/tmp"
 SOLR_MODULE_NAME="search_api_solr"
 SOLR_VERSION="5.x"
-SOLR_CORE_PATH="/var/solr/data/"$NAME
+SOLR_CORE_PATH="/var/solr/data/$NAME"
 
 # Check to see if we've already performed this setup.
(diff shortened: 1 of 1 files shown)


examples/scripts/configure-solr.sh
```diff
@@ -7,29 +7,27 @@
 # `solr` in the `installed_extras`, and is currently set up for the D7 versions
 # of Apache Solr Search or Search API Solr.
 
-SOLR_SETUP_COMPLETE_FILE=/etc/drupal_vm_solr_config_complete
+NAME="drupalvm"
```
geerlingguy: I like this change—one thing though, let's set the default to `d8`, just because that's what the D8 version of the Search API Solr Defaults module sets up in it's default index.

This would make the setup pretty much plug-and-play, as you'd immediately have a populated and working search index and search page after enabling the defaults module, as shown here: http://www.jeffgeerling.com/blog/2017/setting-faceted-apache-solr-search-drupal-8

jimmyko: I second that motion! I didn't know there is a default provided in Search API Solr Search because I am currently using Apache Solr Search. Let me do the change.

examples/scripts/configure-solr.sh
```diff
@@ -7,29 +7,27 @@
 # `solr` in the `installed_extras`, and is currently set up for the D7 versions
 # of Apache Solr Search or Search API Solr.
 
-SOLR_SETUP_COMPLETE_FILE=/etc/drupal_vm_solr_config_complete
+NAME="d8"
+SOLR_SETUP_COMPLETE_FILE="/etc/drupal_vm_solr_config_complete_"$NAME
```
geerlingguy: This `$NAME` should go inside the double quotes.

examples/scripts/configure-solr.sh
```diff
@@ -7,29 +7,27 @@
...
 
 # Search API Solr module.
 SOLR_DOWNLOAD="http://ftp.drupal.org/files/projects/search_api_solr-8.x-1.x-dev.tar.gz"
 SOLR_DOWNLOAD_DIR="/tmp"
 SOLR_MODULE_NAME="search_api_solr"
 SOLR_VERSION="5.x"
-SOLR_CORE_PATH="/var/solr/data/collection1"
+SOLR_CORE_PATH="/var/solr/data/"$NAME
```
geerlingguy: Same here (inside double quotes).

PR #1156, opened 2017-02-07
Fix nginx bad gateway errors

ccjjmartin:
I experienced this on a client site so it is pretty hard to add reproducing steps but the idea is if you have a page that needs a higher fastcgi buffer size you get a 502 error "Bad Gateway" from nginx.

This took a while to track back down but I isolated it to adding these two lines to the nginx-vhost.conf.j2 file.

NOTE: I expect there will be changes to this before it is merged.  These settings match Pantheon's settings so maybe they should be configurable in config.yml instead of directly adding them to this file?  I do think that a default setting for these could be higher though, so that it just works out of the box.

### DIFF
--- a/provisioning/templates/nginx-vhost.conf.j2
+++ b/provisioning/templates/nginx-vhost.conf.j2
@@ -40,6 +40,8 @@ server {
         fastcgi_read_timeout {{ php_max_execution_time }};
         include fastcgi_params;
         fastcgi_pass {{ php_fpm_listen }};
+        fastcgi_buffers 16 16k;
+        fastcgi_buffer_size 32k;
     }
 
     location @rewrite {


joestewart:
We've experienced this issue too but added the changes via "extra_parameters".

```
extra_parameters: |
      fastcgi_buffers 16 16k;
          fastcgi_buffer_size 32k;
```

geerlingguy:
Yeah, I'd recommend going @joestewart's route—I don't want to add such a fine grained default (especially hardcoded) to the Nginx file, unless it's something that is necessary on 80%+ sites.

Thanks for the PR, though!

ccjjmartin:
@joestewart Thanks for the tip!

b-connect:
We have the same problems with big drupal projects.

How about a global config variable like nginx_fastcgi_buffers

geerlingguy:
@b-connect - It would need to be a configuration per-server-block, and I am also leery of making it too customizable, because if you don't know what you're doing/why that error occurs, you could shoot yourself in the foot by setting a value higher/too high. See: https://gist.github.com/magnetikonline/11312172#introduction

I haven't personally run into this issue on a D8 build yet, so I'm wondering if, in the cases where this happens, something's storing a ton of data in cookies or other headers?

sukhjindersingh:
I'm trying to add above entries in `nginx_vhosts: []` like below

```
nginx_vhosts: [
    extra_parameters: |
          fastcgi_buffers 16 16k;
              fastcgi_buffer_size 32k;
  ]
```

but it giving error

```
ERROR! Syntax Error while loading YAML.
  found character '|' that cannot start any token

The error appears to have been in '/Users/sukhjindersingh/Sites/drupal-vm/provisioning/roles/geerlingguy.nginx/defaults/main.yml': line 57, column 23, but may
be elsewhere in the file depending on the exact syntax problem.

The offending line appears to be:

nginx_vhosts: [
    extra_parameters: |
                      ^ here
```

geerlingguy:
@sukhjindersingh - That looks like a YAML syntax error.

You'd need to add a full host like:

```yaml
nginx_hosts:
  - server_name: "{{ drupal_domain }} www.{{ drupal_domain }}"
    root: "{{ drupal_core_path }}"
    is_php: true
    extra_parameters: |
          listen 443 ssl;
          ssl_certificate     /vagrant/example.crt;
          ssl_certificate_key /vagrant/example.key;
          ssl_protocols       TLSv1.1 TLSv1.2;
          ssl_ciphers         HIGH:!aNULL:!MD5;
```

PR #1159, opened 2017-02-09
Fixes #1158, Follow-up #1086: Fix Drupal 7 and nginx breaking install.php

### LINKED ISSUE #1158: install.php redirects to /core/install.php (which then 404s) even in D7?
### Issue Type

### Your Environment

```
Vagrant 1.9.1
VirtualBox 5.0.26r108824
ansible 2.2.1.0
  config file = 
  configured module search path = Default w/o overrides
```

### Your OS

 - **macOS** (10.11.6)

### Summary

I've got the single site install turned off and am using Drupal VM to run  a few different drupal 7 sites at different vhosts, the docroots of which I'm syncing via nfs. The folders seem to be syncing fine - I can create html files in them and go to myhostname.dvm/file.html and see it. However, when I put drupal core into the folder, if I visit the site in a browser, `install.php` does a 301 redirect to `/core/install.php` which does not exist.  This seems to be Drupal 8 behavior getting applied to a Drupal 7 site? 

I was using nginx for my webserver, I'm currently rebuilding with apache to see if the problem is happening when using apache also.

oxyc:
Any better suggestion?

### DIFF
--- a/provisioning/templates/nginx-vhost.conf.j2
+++ b/provisioning/templates/nginx-vhost.conf.j2
@@ -26,10 +26,12 @@ server {
         return 404;
     }
 
+{% if drupal_major_version == 8 %}
     # Redirect common PHP files to their new locations.
     location ~ ^((?!.*(?:core)).*)/(install.php|rebuild.php) {
          return 301 $scheme://$host$1/core/$2$is_args$args;
     }
+{% endif %}
 
     # Use fastcgi for all php files.
     location ~ \.php$|^/update.php {


geerlingguy:
Makes sense. I've segregated all my D7 and D8 sites, so their Nginx configs are customized for each major Drupal version—and don't install the sites locally on Drupal VM anyways, which is why I've never seen this issue crop up.

But it makes perfect sense, and should fix the issue. Merging!

geerlingguy:
(Well, once tests have passed).

PR #1175, opened 2017-02-21
Adding an option to enable the gui in the vagrant box...

finnlewis:
... for times when developers might not want a headless box.

Hi Jeff! 
Thanks for the drupal-vm, it's great!
I'm working with a team who are on old-ish Windows machines, and would rather work inside a Linux VM with a desktop, which requires enabling the gui the Vagrant options.
This allows that, while keeping the default option to headless.
Any chance it would be useful to merge in?
Cheers,
FInn

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -160,7 +160,7 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
     # HGFS kernel module currently doesn't load correctly for native shares.
     override.vm.synced_folder host_project_dir, '/vagrant', type: 'nfs'
 
-    v.gui = false
+    v.gui = vconfig['vagrant_gui']
     v.vmx['memsize'] = vconfig['vagrant_memory']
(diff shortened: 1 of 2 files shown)


oxyc:
You can also add this using a [`Vagrantfile.local`](http://docs.drupalvm.com/en/latest/extending/vagrantfile/)

```rb
config.vm.provider :virtualbox do |v|
  v.gui = true
end
```

finnlewis:
Thanks oxyc, I hadn't realised or thought about that. 
I can certainly see the argument for separating out vagrant specific changes in that way, but at the same time it is kind of nice to have those options in config.yml to allow keep to a single local config file. There are also plenty of other options that alter the vagrant config in config.yml, so I guess it seemed the natural place for it.

geerlingguy:
I like having it configurable like this—there have been a few times when I've even wanted to pop the GUI for certain debugging/testing, so I'm fine with this var (instead of requiring a Vagrantfile.local to do it).

PR #1184, opened 2017-02-28
Fix varnish repo id on CentOS 6

derekderaps:
Fixes use of Jinja2 string concatenation by setting an extra fact to
represent the repository identifier. Alternative would be to use an
array of strings with the
[join](http://jinja.pocoo.org/docs/2.9/templates/#join) filter.


geerlingguy:
@derekderaps - Can you please file this against the upstream role? https://github.com/geerlingguy/ansible-role-varnish

derekderaps:
Ah yes, sure thing!

PR #1187, opened 2017-02-28
Indicate "less than" version constraint with "<"

derekderaps:
Currently, it uses a "greater than" sign (">").


geerlingguy:
Are you sure that's correct? It should be 'PHP greater than 5.5', I think... technically greater than or equal to—5.4 and below didn't have native opcache support?

derekderaps:
Oh! Nevermind then, I was confused...

Issue #1171, opened 2017-02-15
Support Drush and xdebug inside DrupalVM


japerry:
### Issue Type
 - Feature Idea
 - Documentation Report

### Your Environment
Vagrant 1.9.1
VirtualBox 5.1.4r110228
ansible 2.2.0.0
  config file = 
  configured module search path = Default w/o overrides

### Your OS
 - **macOS** 10.11.6

### Summary
Drush+xdebug is a powerful tool for stepping through code that isn't available via the UI, with migrate being the biggest example. 

DrupalVM is almost capable of allowing for this, however it needs to accept the following three environment variables: 
```bash
# Allow client to pass locale environment variables
AcceptEnv PHP_IDE_CONFIG
AcceptEnv XDEBUG_CONFIG
AcceptEnv PHP_OPTIONS
```

oxyc:
Also https://github.com/geerlingguy/drupal-vm/issues/716#issuecomment-267700090

heddn:
I'm able to test drush & migrate in d8 with xdebug right now inside the VM. You just have to setup things inside the VM to send its xdebug stuff to outside so PHPStorm can catch things.

oxyc:
@japerry does the above PR solve it? any chance you could provide an example on how to use it? Or some docs would be greatly appreciated.

geerlingguy:
Oh haha, I just read this comment after I made an almost identical comment in the PR. @japerry ^^

oxyc:
Ops, forgot to mention it was WIP. 

Or @mirsoftacquia, any chance you could [provide a good example](https://github.com/geerlingguy/drupal-vm/issues/716#issuecomment-267700090) we could add to the docs?

oxyc:
As I understand we can load XDebug on demand with this (I use vim and haven't set up an IDE so can't test this right now though).

```sh
export PHP_OPTIONS="-d xdebug.profiler_enable=1"
```

and then xdebug will be enabled in the CLI whenever you (or your IDE) SSH into Drupal VM.

Tested that the following scenario works:

```sh
PHP_OPTIONS="-d xdebug.profiler_enable=1" drush @drupalvm.drupalvm.dev php-eval 'echo file_get_contents(xdebug_get_profiler_filename());'
```

while the following does not:

```sh
drush @drupalvm.drupalvm.dev php-eval 'echo file_get_contents(xdebug_get_profiler_filename());'
```

[...]

Issue #716, opened 2016-06-09
Xdebug PHPUnit Tests from within VM - CLI and/or PHPStorm


justinlevi:
I realize this is a bit off topic but thought I'd throw out the question here anyway. I'm writing some PHPUnit tests for a D8 custom module and doing the development work on the VM. XDebug is setup and working fine for breakpoints if I load the site through my browser. However, I'm running into an exception for one of my tests and I'd like to debug within PHPStorm and I don't want to run my tests by using the testing module. I'm not even sure if that would work in the first place...

I've tried to create quite a few different PHPStorm debug configurations both using the default PHPUnit configuration and the php script configuration with no luck. I've also tried to run my tests from the command line, which works fine, but I can't get XDebug to trip. 

Here's the command line I'm trying to run
```
export XDEBUG_CONFIG="idekey=PHPSTORM"; sudo php docroot/core/scripts/run-tests.sh --url http://dvm.dev --verbose --color
--class "Drupal\soapwebservice\Tests\SoapWebServiceTest"
```

Here's a shell script I created to make my life a bit easier
```
#!/usr/bin/env bash 
export XDEBUG_CONFIG="idekey=PHPSTORM" 
echo $1 
php /var/www/dvm/docroot/core/scripts/run-tests.sh --url http://dvm.dev --verbose --color --execute-test $1
```

source: http://vlikin.blogspot.com/2013/11/debuggingtracing-drupal-8-tests.html

---

I've also tried setting up a debug configuration in PHPStorm with the following settings, per slide 22 of the talk below: 
**Xdebug and Drupal8 tests (PhpUnit and Simpletest)**
http://www.slideshare.net/FranciscoSeva/xdebug-and-drupal

![debug-tests-run-tests](https://cloud.githubusercontent.com/assets/1117989/15936116/c838753c-2e37-11e6-8a63-7ff8385da727.jpg)

Here's the thing that's totally throwing me off though. This site is a Bolt derivative, so all dependencies are managed with composer, and the tests within `/var/www/dvm/tests/phpunit` all run as expected in PHPStorm. I was thinking this could be an Autoloader issue but I tried to run google_analytics tests from within PHPStorm and I'm getting the same errors. 

What is the the workflow for debugging(xdebug) PHPUnit tests for a custom module from within PHPStorm? If I can get this all setup and working I will definitely do a blog post on this one...

geerlingguy:
[...]

PR #1189, opened 2017-02-28
Issue #1171: Allow client to pass PHP and XDEBUG environemnt variables over SSH

oxyc:
~todo: docs~

### DIFF
--- a/docs/extras/xdebug.md
+++ b/docs/extras/xdebug.md
@@ -48,6 +48,27 @@ To use XDebug with NetBeans, change the `php_xdebug_idekey` variable as shown be
 php_xdebug_idekey: netbeans-xdebug
 ```
 
+### XDebug over SSH/Drush
+
+As long as `xdebug` is listed in `installed_extras` Drupal VM is configured to accept the `PHP_IDE_CONFIG`, `XDEBUG_CONFIG` and `PHP_OPTIONS` environment variables over SSH and this can be used to set up some IDE's as well as enable XDebug on a per request basis:
+
+```
+PHP_OPTIONS="-d xdebug.profiler_enable=1" drush @drupalvm.drupalvm.dev migrate-import
+```
+
+To send the environment variables when using `vagrant ssh`, [create a `Vagrantfile.local`](../extending/vagrantfile.md) with:
+
+```
+config.ssh.forward_env = ['PHP_IDE_CONFIG', 'XDEBUG_CONFIG', 'PHP_OPTIONS']
+```
+
+And you can run:
+
+```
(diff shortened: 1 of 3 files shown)


geerlingguy:
@oxyc - Can you document somewhere how to use this?

oxyc:
I'll try and play around with some examples for this during the weekend but I'm now thinking that maybe it's useful to have this toggled on by a config option instead. It could be useful outside the realm of xdebug and seems odd that it's depending on it.

Maybe add `ssh_accept_php_vars: true`

FinBoWa:
@oxyc just hit me that i commented on the https://github.com/geerlingguy/drupal-vm/issues/716#issuecomment-283272705 and that was closed all ready. 

So wouldn't it be better that ansible generated those lines to the vagrant users bashrc and based them on the config.yaml settings.

oxyc:
I don't think they should be hardcoded to be honest. I feel it's better if especially `PHP_OPTIONS` is set only when required, and not on every CLI request.

oxyc:
I applied for a PhpStorm opensource license, let's see if I get it before the weekend and I can do some testing.

oxyc:
I quickly remembered why I'm still using vim... I have to postpone getting into this whole IDE thing again...

geerlingguy:
@oxyc - You and me both. Getting Xdebug happy through an IDE with PHP reminds me of developing for Java... way overcomplicated just to print out a variable :P

FinBoWa:
For me getting xdebug working with Vim would be problematic also :). Anyways the phpstorms https://www.jetbrains.com/help/phpstorm/2016.3/zero-configuration-debugging.html is quite a breez to configure but the documentation for it is way overly verbose. 

I could translate my guide on getting phpstorm to work somewhat right in 5 steps tomorrow.

geerlingguy:
Is this still necessary/required? I have things working with Sublime Text 3 without needing to set any additional vars like those in the PR... I'm still not entirely clear on why this is required... but I also don't use PHPStorm much.

thom8:
I think this PR is for CLI XDEBUG via SSH.

geerlingguy:
@thom8 - Oh, I was totally misreading that. In that case, does anyone have any objections to the way @oxyc implemented this? I figure we could sneak it into the 4.5.0 release as well.

bdone:
no objections. it would be helpful and appreciated. thanks!

geerlingguy:
I'd like some form of docs (even if it's just inline in a comment on this PR) before merging, though.

oxyc:
I verified that the above examples work (after stumbling around because the ssh service isn't restarted on changes).

oxyc:
Probably better to use `default_enable` in the examples as that's the more likely use case. I just verified that it worked by checking that the `/tmp/cachegrind.out.*` files were generated correctly when using `profiler_enable`.

geerlingguy:
Restarted build, hopefully it comes back all green this time.

PR #1198, opened 2017-03-06
Use VAGRANT_HOME to get ssh key directory

derimagia:
Fixes having a VAAGRANT_HOME set to a different directory.

### DIFF
--- a/provisioning/templates/drupalvm.aliases.drushrc.php.j2
+++ b/provisioning/templates/drupalvm.aliases.drushrc.php.j2
@@ -13,7 +13,7 @@ $aliases['{{ host }}'] = array(
   'root' => '{{ root }}',
   'remote-host' => '{{ host }}',
   'remote-user' => '{{ vagrant_user }}',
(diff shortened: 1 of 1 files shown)


oxyc:
Makes sense to me. I guess the use case is placing the VMs on a separate drive?

geerlingguy:
Might just need a rebase, or centos7 docker image might need a manual rebuild again...

derimagia:
Vagrant info (I used XDG Base DIr Standard (https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html). VMs are stored based on the provider you use (For virtualbox it's completely separate and not related to VAGRANT_HOME).

The other way this can be done is to get the key path from vagrant itself, but I didn't look into that too much.

geerlingguy:
Makes sense... and timely, too, since I was just testing things on an older Windows PC where I had to put `VAGRANT_HOME` on a separate path for performance's sake.

PR #1221, opened 2017-03-15
Switch to set PHP's default to 7.1

### LINKED ISSUE #1031: PHP 7.1 - Timeline for switching to default

heddn:
Fixes #1031

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -239,7 +239,7 @@ firewall_disable_ufw: true
 
 # PHP Configuration. Currently-supported versions: 5.6, 7.0, 7.1.
 # See version-specific notes: http://docs.drupalvm.com/en/latest/configurations/php/
-php_version: "7.0"
+php_version: "7.1"
 php_install_recommends: no
 php_memory_limit: "192M"
 php_display_errors: "On"


geerlingguy:
@oxyc had questioned whether we'd need a major release bump for this... what do you think? I'd be fine with a minor release bump, especially considering that 7.1 is such a small transition from 7.0 vs 5.6 to 7.0... Plus it's not like there was any other architectural change besides bumping a version number.

Issue #1220, opened 2017-03-15
Unable to upload images inline-images


skorasaurus:
### Your Environment

```
Vagrant 1.9.1
VirtualBox 5.1.16r113841
ansible 2.2.1.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
drupal-vm version: of https://github.com/geerlingguy/drupal-vm/commit/1a981762840072121c4c1b102e6ff242ce6cfede
```

### Your OS

ubuntu 16.04.02

### Full console output
none, see below

### Summary

Unable to upload an image in a page in the default site that was created. Vanilla setup, no modifications to the config file. 

Here's the error that I receive in the log messages: ![selection_005](https://cloud.githubusercontent.com/assets/955351/23964344/8c4f3ad6-098a-11e7-8ba6-bbce99f2cd0a.png)

or as text ```The upload directory public://inline-images for the file field could not be created or is not accessible. A newly uploaded file could not be saved in this directory as a consequence, and the upload was canceled.```

I found this had occurred in the past [with #99](https://github.com/geerlingguy/drupal-vm/issues/99) and was fixed at https://github.com/geerlingguy/drupal-vm/commit/1a39996fb2f9c6e54e78f6522647eb684484eee7  but apparently the issue still exists or a regression occurred. 

My file permissions inside the VM suggest that I should be able to create them:
```
vagrant@drupalvm:/var/www/drupalvm/drupal/web/sites/default/files$ ls -lah
total 64K
drwxrwxrwx 8 1000 admin 4.0K Mar 15 18:08 .
dr-xr-xr-x 3 1000 admin 4.0K Mar 14 16:36 ..
drwxrwxr-x 3 1000 admin 4.0K Mar 14 16:37 config_rAIeGI6LbF2nVldKW-LdnNL_9Rur7HDdFi5zv1Q9b1me2-bmNRjnJ5VCMrjIrVYBTyXWgLmBUA
drwxrwxr-x 2 1000 admin  12K Mar 15 18:14 css
-r--r--r-- 1 1000 admin  487 Mar 14 16:37 .htaccess
drwxrwxr-x 2 1000 admin 4.0K Mar 15 18:08 inline-images
drwxrwxr-x 2 1000 admin  12K Mar 15 18:14 js
-rw-rw-r-- 1 1000 admin 9.5K Mar 15 18:04 logo.png
[...]

PR #1223, opened 2017-03-15
Issue #1220: Fix to ensure www-data is in the same group as NFS synced dirs

oxyc:
This is still WIP. Not the nicest solution and it needs proper testing.

Using `stat.gr_name` would be way nicer and could remove the shell task, unfortunately I can't get this to work though. It's never defined in my case.

Also, if the user sets `vagrant_synced_folder_default_type` to something other than `nfs` while using `nfs` for the primary synced folder, this wont help. Other than that I don't see this causing issues. Maybe with smb somehow?

### DIFF
--- a/provisioning/tasks/www.yml
+++ b/provisioning/tasks/www.yml
@@ -4,24 +4,60 @@
     path: /vagrant
   register: vagrant_directory
 
-- name: Detect if group used to sync directories already exist.
-  shell: "getent group {{ vagrant_directory.stat.gid }} | cut -d':' -f1"
-  register: vagrant_directory_groupname
-  when: vagrant_directory.stat.exists
-
-- name: Ensure a group with the same GID as used to sync directories exist.
+# When using NFS the group id of a folder will be identical to that of the host
+# machine, but the groupname will differ or not exist. For the latter case
+# we create a group called `vagrant_group`.
+#
+# In Ansible 2.3+ the gr_name will be set if the GID is mapped to an existing
+# group. If the GID doesn't exist, gr_name will be undefined.
+- name: Ensure a group with the same GID as used to sync directories exist (Ansible 2.3+).
   group:
     gid: "{{ vagrant_directory.stat.gid }}"
     name: vagrant_group
     state: present
-  when: vagrant_directory.stat.exists and vagrant_directory_groupname.stdout == ''
+  when: >
+    ansible_version.full | version_compare('2.3', '>=') and
+    vagrant_directory.stat.exists and
+    vagrant_directory.stat.gr_name is undefined
 
-- name: Ensure www-data user is in the same group as the owner of synced directories.
+- name: Ensure the webserver user is in the same group as the owner of synced directories (Ansible 2.3+).
+  user:
+    name: www-data
+    append: yes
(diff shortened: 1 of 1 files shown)


geerlingguy:
Yeesh, it looks a little fragile (and definitely a bit obtuse), but let me play around when I get some time and see if there's a better way I can think of (probably not... but who knows!).

I'd rather not merge this without considering it a while.

Also, the build failures:

https://travis-ci.org/geerlingguy/drupal-vm/jobs/211519983#L2205 and https://travis-ci.org/geerlingguy/drupal-vm/jobs/211519985 have:

```
/opt/mailhog/mhsendmail: not found
```

Maybe just a fluke? I'll kick those builds again.

oxyc:
I went down the rabbit hole trying to figure out why `stat.gr_name` isn't available for the directory but couldn't find anything. I'll look into it again later as well.

oxyc:
Okay I figured out why gr_name isnt working. In ansible 2.2.0.0 before https://github.com/ansible/ansible-modules-core/commit/52e0536fae2ae8df752f41868ed50486b5653292 `getgrgid` never ran if `getpwuid` failed, which it did because of the nonexisting 501 user.

Unfortunately the commit above isn't in a tagged release yet. Once it is, the shell task can be removed.

oxyc:
Unfortunately the commit isnt in `v2.2.2.0-0.2.rc2` either. But it is in `v2.3.0.0-0.1.rc1`

oxyc:
Also just realized this only fixes the issue for Debian. We need to use the httpd user for centos.

oxyc:
@geerlingguy I pushed a commit so you can see how it will look once Ansible 2.3 is a dependency.

One nice thing about this is I believe it will solve almost all permission related errors people have. Out of the box.

oxyc:
Added a safety check in case the group resolves to root (in case someone somehow runs ansible with sudo). Maybe it's not needed but feels a bit safer...

geerlingguy:
Interesting... it's definitely looking less shaky, and in a quarter or two, once 2.3 has been stable for a while, we could ditch the older tasks. I'll give it a spin.

oxyc:
For rererence, if someone syncs using the default VirtualBox synced folders method, the owner:group will be `vagrant:vagrant`.

As long as the user also changed `vagrant_synced_folder_default_type: ""` www-data will end up being part of the `vagrant` group. If they don't change the default folder type, the group will resolve to dialout on osx. I dont see this scenario happening though.

What's more likely to happen is that a Windows user sets

```yaml
vagrant_synced_folder_default_type: ""
vagrant_synced_folders:
  - local_path: .
    type: ""
    create: true
```

and thus the owner:group becomes `vagrant:vagrant` with www-data added to the `vagrant` group.

geerlingguy:
Looks good now—we'll just have to remember to drop the Ansible < 2.3 support sometime in the future. Thanks!

Hopefully this will head off the percentage of issues that keep popping up for permissions-related things!

nvahalik:
I just tried this and my NFS mount (using `ubuntu1204`) is still `root:root`. This is on a completely new box.

I'm on Windows 7 and I'm using the `vagrant-winnfsd` plugin.

Issue #1213, opened 2017-03-12
Add specific tags to speed up common configuration updates


geerlingguy:
### Issue Type

 - Feature Idea

### Summary

I've always been a bit more of a stickler against tagging _all the things_ in Ansible (and still am), but see some value in adding tags to specific roles, includes, etc. which are often helpful in speeding up the deployment of specific config changes.

There are a few of these specific things which I think we could support out of the box in Drupal VM and wouldn't add too much weight to the main playbook:

  - PHP (options like memory limit changes)
  - XDebug configuration changes (quite often requested)
  - Drupal deployment (the new `drupal` role, basically)
  - Apache/Nginx ('webserver') changes

I hope some other people may chime in and offer suggestions of other things that could be tagged.

The basic process is:

First, add a tag to a part (or relevant parts) of the playbook like:

```yaml
    # Build and configure the Drupal site.
    - include_role: name=geerlingguy.drupal
      static: no
      tags: ['drupal']
```

Then, if you _just_ want to call that tag, you can specify the tags arg with `DRUPALVM_ANSIBLE_ARGS` like so:

    DRUPALVM_ANSIBLE_ARGS="--tags=drupal" vagrant provision

Any other sets of tags people think would be nice to have? Parts of the VM you might frequently change and don't want to wait the full 2-4 minutes for a provision run.

oxyc:
Hmm maybe add `DRUPALVM_ANSIBLE_TAGS` for convenience? Seems more like a "supported feature" then.

```rb
if tags = ENV['DRUPALVM_ANSIBLE_TAGS']
  ansible.tags = tags
end
```

geerlingguy:
@oxyc - True, would accept a PR for that—that way you could specify tags through raw args _or_ that env var.

PR #1224, opened 2017-03-15
Follow-up #1213: Add support for DRUPALVM_ANSIBLE_TAGS env variable

oxyc:
I refactored the provisioner declaration so that it's more DRY. Doesnt matter much now but maybe if we add SKIP_TAGS or some more config variables in the future.


geerlingguy:
LGTM!

Issue #1199, opened 2017-03-07
How to change owner or group when syncing with rsync?


pminf:
Currently I'm struggling with permission issues in my Drupal installation. I'm syncing with rsync which leads to owner and group `vagrant` of synced files and folders. Because apache is running with user `www-data` files cannot be written in `sites/default/files`. That's why I'm trying to change the group of synced files to `www-data`. How do I accomplish this?

### Issue Type
 - Support Request

### My Environment
```
Vagrant 1.9.1
sh.exe": vboxmanage: command not found
VirtualBox
sh.exe": ansible: command not found
```

### My OS

Microsoft Windows [Version 10.0.14393]

### Summary

I've already tried the following settings in `config.yml`:
```
vagrant_synced_folders:
  - local_path: C:\#\myproject
    destination: /var/www/myproject.dev
    type: rsync
    create: true
    options_override:
      group: www-data
```
or
```
vagrant_synced_folders:
  - local_path: C:\#\myproject
    destination: /var/www/myproject.dev
    type: rsync
    create: true
    group: www-data
```
These don't take effect after `vagrant reload`. When I check `.vagrant/machines/mydrupalvmbox/virtualbox/syncedfolders` `group` is still `vagrant`. Changing the group in this temp file and doing a `vagrant rsync` results the correct group for rsynced files and directories. But after `vagrant reload` those temp settings are gone and group `vagrant` is back again.

I've also tried to change the group via `rsync_args` with no success:
```
vagrant_synced_folders:
  - local_path: C:\#\myproject
    destination: /var/www/myproject.dev
    type: rsync
    create: true
   options_override:
     rsync__args: [
       "--verbose", "--archive", "--delete",
       "--chmod=gu=rwX,o=rX",
       "--group", # required for the following command
       "--groupmap=*:www-data"
     ]
```
I get an error: `Error: rsync: --groupmap=*:www-data: unknown option`.

geerlingguy:
I'll leave this ticket open here for a time to let anyone else using rsync have a chance to chime in, but I wonder if this might be something better suited to a Vagrant issue queue or stack overflow, since it's not really Drupal-VM-specfic.

geerlingguy:
Also, have you read through Vagrant's `rsync_chown` documentation? https://www.vagrantup.com/docs/synced-folders/rsync.html#rsync_chown

pminf:
[...]

PR #1225, opened 2017-03-16
Issue #1199: Fix synced folder examples in docs

oxyc:
While reading through the synced folder docs I also tested out one of the examples that seemed incorrect. NFS does not support `dmode` or `fmode`, and produces an invalid mount command. Also `owner`/`group` doesn't affect NFS.


geerlingguy:
Thanks!

PR #1226, opened 2017-03-16
Follow-up #1222: Update PHP 7.0 references to 7.1

oxyc:


geerlingguy:
I'll update the public site codebase too!

PR #1228, opened 2017-03-16
Remove PHP 5.5 logic as it's not supported anymore

oxyc:


geerlingguy:
Nice! I love removing code.

PR #1233, opened 2017-03-16
Cleanup playbook

oxyc:
Now when all the drupal stuff is moved to it's own role (:tada:) the playbook feels so clean. Here's a few cleanup suggestions.

### DIFF
--- a/provisioning/playbook.yml
+++ b/provisioning/playbook.yml
@@ -5,11 +5,9 @@
   vars_files:
     - vars/main.yml
     - ../default.config.yml
+    - "vars/{{ ansible_os_family }}.yml"
 
   pre_tasks:
-    - name: Include OS-specific variables.
-      include_vars: "{{ ansible_os_family }}.yml"
-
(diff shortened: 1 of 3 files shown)


geerlingguy:
LOVE this. I want to keep the main playbook down to as slim and trim as we can keep it—this also helps when there is a time to break off a role separately, too... e.g. I've been wondering if I might want to abstract all the PHP version stuff into a `geerlingguy.php-versions` role that can be included to set things up for PHP 5.6, 7.0, 7.1 (and beyond) for all the current OS releases (e.g. take all the repo setup and package setup and extract it into a role).

oxyc:
That would be awesome! That's most of the content in both `init-debian.yml` and `init-redhat.yml`...

Another one I would suggest to split to it's own repo is `tasks/hostname.yml` together with `Configure /etc/mailname` in `tasks/init-debian.yml`. Seems like it could be a useful as an independent role?

PR #1238, opened 2017-03-19
Add an intro message for vagrant up/reload

oxyc:
Just posting ideas...

```sh
==> drupalvm_project: Mounting NFS shared folders...
==> drupalvm_project: Mounting shared folders...
    drupalvm_project: /var/www/drupalvm => /Users/cindy/Projects/Personal/drupal-vm
==> drupalvm_project: Machine already provisioned. Run `vagrant provision` or use the `--provision`
==> drupalvm_project: flag to force provisioning. Provisioners marked to run always will still run.

==> drupalvm_project: Machine 'drupalvm_project' has a post `vagrant up` message. This is a message
==> drupalvm_project: from the creator of the Vagrantfile, and not from Vagrant itself:
==> drupalvm_project:
==> drupalvm_project: Your Drupal VM Vagrant box is ready to use!
==> drupalvm_project: * Visit the dashboard for an overview of your site: http://dashboard.drupalvm-project.dev (or http://192.168.99.88)
==> drupalvm_project: * You can SSH into your machine with `vagrant ssh`.
==> drupalvm_project: * Find out more in the Drupal VM documentation at http://docs.drupalvm.com
```

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -99,6 +99,13 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # Vagrant box.
   config.vm.box = vconfig['vagrant_box']
 
+  unless vconfig['vagrant_skip_post_up_message']
+    config.vm.post_up_message = 'Your Drupal VM Vagrant box is ready to use!'\
+      "\n* Visit the dashboard for an overview of your site: http://dashboard.#{vconfig['vagrant_hostname']} (or http://#{vconfig['vagrant_ip']})"\
+      "\n* You can SSH into your machine with `vagrant ssh`."\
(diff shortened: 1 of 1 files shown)


oxyc:
My reasoning is that rather than account for all the different scenarios, just make the message useful for beginners and let eg teams who have some fancier setup (disabled dashboard, blabla) either a) disable the message by setting `vagrant_post_up_message: false` or b) set their own welcome message to introduce co-workers to the VM.

geerlingguy:
I like this idea! Does `vagrant_post_up_message: false` actually work, though? It seems like it would cause the else to be triggered...

oxyc:
Oh right! I'll fix that.

geerlingguy:
Works for me!

PR #1240, opened 2017-03-19
Issue #1199: Fix options_override causing duplicate json entries

oxyc:
See https://github.com/geerlingguy/drupal-vm/issues/1199#issuecomment-287633956


geerlingguy:
I'll take your word for it; thanks for doing this deep testing!

PR #1255, opened 2017-03-24
Allow provided templates to be extended through template inheritance

oxyc:
~This is entirely untested so still WIP.~

### DIFF
--- a/docs/configurations/webservers-nginx.md
+++ b/docs/configurations/webservers-nginx.md
@@ -40,7 +40,7 @@ If you are using Ubuntu as your base OS and you want to get started quickly with
         ssl_ciphers         HIGH:!aNULL:!MD5;
 ```
 
-### Customizing server block configuration
+## Customizing server block configuration
 
 If you can't customize via variables because an option isn't exposed, you can override the template used to generate the the virtualhost configuration file.
 
@@ -48,9 +48,9 @@ If you can't customize via variables because an option isn't exposed, you can ov
 nginx_vhost_template: "{{ config_dir }}/templates/nginx-vhost.conf.j2"
 ```
 
-You can either copy and modify the provided `nginx-vhost.conf.j2` template, or extend it and and override the specific template block you need to change.
+You can either copy and modify the provided `nginx-vhost.conf.j2` template, or extend it and override the specific template block you need to change.
 
-_If you extend Drupal VM's provided base template, the path referenced should to be relative to the playbook.yml._
+_If you extend Drupal VM's provided base template, the path referenced should to be relative to playbook.yml._
 
 ```
 {% extends 'templates/nginx-vhost.conf.j2' %}
--- a/docs/configurations/webservers-nginx.md
+++ b/docs/configurations/webservers-nginx.md
@@ -1,6 +1,6 @@
 To use Nginx instead of Apache, change the `drupalvm_webserver` variable inside your customized `config.yml`, from `apache` to `nginx`.
 
-Because Nginx server directives behave a little differently than Apache's VirtualHosts, Drupal VM includes a custom Drupal-optimized Nginx server block configuration, and you can control all the servers ('virtual hosts') Nginx will run using the `nginx_hosts` configuration. A few simple examples are shown in `default.config.yml`, but you have some extra flexibility if you need it. See the `nginx-vhost.conf.j2` template for more information.
+Because Nginx server directives behave a little differently than Apache's VirtualHosts, Drupal VM includes a custom Drupal-optimized Nginx server block configuration, and you can control all the servers ('virtual hosts') Nginx will run using the `nginx_vhosts` configuration. A few simple examples are shown in `default.config.yml`, but you have some extra flexibility if you need it. See the `nginx-vhost.conf.j2` template for more information.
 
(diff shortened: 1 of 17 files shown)


docs/configurations/webservers-nginx.md
```diff
@@ -39,3 +39,43 @@ If you are using Ubuntu as your base OS and you want to get started quickly with
...
+
+If you can't customize via variables because an option isn't exposed, you can override the template used to generate the the virtualhost configuration file.
+
+```yaml
+nginx_vhost_template: "{{ config_dir }}/nginx-vhost.conf.j2"
+```
+
+You can either copy and modify the provided `nginx-vhost.conf.j2` template, or extend it and and override the specific template block you need to change.
+
+_If you extend Drupal VM's provided base template, the path referenced should to be relative to the playbook.yml._
```
oxyc: Not sure this is true, need to check it.

oxyc: It is true.

oxyc:
Made some quick tests by following the docs I've added. Everything seems to be in order.

provisioning/tasks/www.yml
```diff
@@ -88,7 +88,7 @@
 
 - name: Copy Nginx vhosts into place.
   template:
-    src: ../templates/nginx-vhost.conf.j2
+    src: "{{ nginx_vhost_template }}"
     dest: "{{ nginx_vhost_path }}/{{ item.server_name.split(' ')[0] }}.conf"
     force: yes
     owner: root
```
oxyc: @geerlingguy any chance you'd consider making your nginx role support multiple vhost files rather than looping server directives within one vhost file? Then we could remove this task :)

geerlingguy: Yes, in fact that's how I do it for my main Drupal multisite server (home of www.midwesternmac.com, www.jeffgeerling.com, www.beforeidostl.com among others...). I just never pulled that back into the Nginx role.

It makes things easier to manage, but I've been selfish and never contributed it back! :P

oxyc: https://github.com/geerlingguy/ansible-role-nginx/issues/70

geerlingguy:
@oxyc - I'm going to try to get this done before or during DrupalCon Baltimore.

oxyc:
Might be better to partially upstream these to the nginx and the varnish role as well.

geerlingguy:
@oxyc - Yeah, the goal would be to make sure nginx, apache, and varnish roles have some form of parity in terms of how you define configs. Right now those three roles are a bit disjointed if you look at them as something that could potentially work together in one playbook.

geerlingguy:
I'm going to tag a new patch release, I think, prior to merging this work. Since it's slightly more structural, I want to leave this for a new release—and if we find that it disrupts nothing/doesn't cause any backwards compatibility issues whatsoever, it can be another patch release. But if we end up changing anything else we can go to 4.5.

geerlingguy:
@oxyc - In light of https://github.com/geerlingguy/ansible-role-nginx/pull/91 — did you already have any work to adapt that vhost config into Drupal VM... or do you think we should just continue to maintain the complete template for now?

geerlingguy:
Also, looks like this just needs a rebase (I did so briefly to verify tests are still passing in https://github.com/geerlingguy/drupal-vm/pull/1343 — but I closed that PR so you can finish off this one when ready.)

oxyc:
I had an untested example here https://github.com/geerlingguy/ansible-role-nginx/pull/92#issuecomment-289242138 but as we're almost overriding everything anyways maybe it's not worth extending and just maintain our own?

geerlingguy:
Good point. PHP projects like Drupal are unique in that they require a lot more custom path-based rules to match what's in the .htaccess file.

oxyc:
Let me go through the changes once more before merging, especially to check the block names. I'll get to it either today or tomorrow and let you know.

geerlingguy:
Thanks! I have a session to present tomorrow, so I'll probably be heads-down on that until late Thursday. Then there's Trivia Night, so I may not be back to a functional state until Friday sometime :)

geerlingguy:
Oh, also, I just pushed the Nginx role version update... let me know if something's still wrong there!

geerlingguy:
Looks good to me!

lib/drupalvm/vagrant.rb
```diff
@@ -95,7 +95,8 @@ def get_vhost_aliases(vconfig)
   if vconfig['drupalvm_webserver'] == 'apache'
     aliases = get_apache_vhosts(vconfig['apache_vhosts'])
   else
-    aliases = get_nginx_vhosts(vconfig['nginx_hosts'])
+    # @todo shim for `nginx_hosts`.
```
geerlingguy: The `@todo` here - is that just meant to say 'remove the shim part at some point in the future'?

provisioning/tasks/drush-aliases.yml
```diff
@@ -18,7 +18,7 @@
 # Note that this doesn't work for Windows, since Ansible's running in the VM.
 - name: Configure host machine drush aliases.
   template:
-    src: ../templates/drupalvm.aliases.drushrc.php.j2
```
geerlingguy: At some point, maybe we could pull drush aliases out into a separate role too.

lib/drupalvm/vagrant.rb
oxyc: yeah

Issue #1247, opened 2017-03-23
Composer plugin?


thom8:
### Issue Type

 - Feature Idea

### Summary

@geerlingguy have you considered making DVM a composer plugin so you don't need a 
[delegating Vagrantfile](http://docs.drupalvm.com/en/latest/deployment/composer-dependency/#create-a-delegating-vagrantfile)?

see https://github.com/beetboxvm/beetbox/blob/master/composer/src/Plugin.php

This creates/updates a `Vagrantfile` during a `composer install`

Then you'd just ignore the `Vagrantfile` in `.gitignore`

geerlingguy:
Interesting... definitely something worth considering. It would take one extra step out of integrating Drupal VM straight into a project.

geerlingguy:
The main concern I'd have is how easy it would be to allow customizations per project (e.g. if you want your config_dir to be `vm` or `box` or `drupalvm` or whatever).

thom8:
Completely understand, which is why it would be difficult to jump to PR stage.

We've also been testing the concept of adding config directly to `composer.json` -- https://github.com/thom8/drupal8-vagrant/blob/master/composer.json#L27

thom8:
Also for a bit of background we're also moving to only support composer installations to help with breaking backwards compatibility -- https://github.com/beetboxvm/beetbox/issues/391

oxyc:
So with beetbox as an example, we could do something like this to support specifying the config_dir in `composer.json` as well.

```diff
diff --git a/Vagrantfile b/Vagrantfile
index 386e9d0..46fb62e 100644
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -14,6 +14,16 @@ guest_config_dir = ENV['DRUPALVM_CONFIG_DIR'] ? "/vagrant/#{ENV['DRUPALVM_CONFIG

 drupalvm_env = ENV['DRUPALVM_ENV'] || 'vagrant'

+if File.exist?("#{host_project_dir}/composer.json")
+  require 'json'
+  composer_conf = JSON.parse(File.read("#{host_project_dir}/composer.json"))
+  config_dir = composer_conf['extra']['drupalvm']['config_dir'] rescue nil
+  if config_dir.is_a?(String)
+    host_config_dir = "#{host_project_dir}/#{config_dir}"
+    guest_config_dir = "/vagrant/#{config_dir}"
+  end
+end
+
 # Cross-platform way of finding an executable in the $PATH.
[...]

PR #1256, opened 2017-03-24
[WIP] Add Composer plugin which scaffolds a Vagtranfile

oxyc:
Proof of concept for #1247 which supports all earlier setups with the addition of reading `config_dir` from `composer.json`.

Might want to drop support for some scenarios at some point tbh...


geerlingguy:
Like basically say "Add Drupal VM as a composer dependency, or use it standalone, and that's it"?

oxyc:
Yeah, that would probably make it easier.

Just realized that making this change would break all existing composer dependency setups as the old `Vagrantfile` will be overridden and the `DRUPALVM_CONFIG_DIR` definition will disappear.

geerlingguy:
> break all existing composer dependency setups

😩

geerlingguy:
Also, with the Vagrantfile ever-expanding in size, I wonder if it might be good to have a '`ruby-src` subfolder somewhere (something like that... whatever the Ruby convention would be) and break out business logic in there?

Or if there's some other way of building up a complicated Vagrantfile that's a better pattern. It's just getting unwieldy at this point.

oxyc:
Only one I've seen that breaks out the logic is homestead.

https://github.com/laravel/homestead/blob/master/Vagrantfile
https://github.com/laravel/homestead/blob/master/scripts/homestead.rb

oxyc:
This is what it would look like if we dropped support for a delegating Vagrantfile

```diff
diff --git a/Vagrantfile b/Vagrantfile
index ae5021f..3e4b5da 100644
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -6,18 +6,6 @@ require 'yaml'

 VAGRANTFILE_API_VERSION = '2' unless defined? VAGRANTFILE_API_VERSION

-# Absolute paths on the host machine.
-host_drupalvm_dir = File.dirname(File.expand_path(__FILE__))
-host_project_dir = ENV['DRUPALVM_PROJECT_ROOT'] || host_drupalvm_dir
-host_config_dir = ENV['DRUPALVM_CONFIG_DIR'] ? "#{host_project_dir}/#{ENV['DRUPALVM_CONFIG_DIR']}" : host_project_dir
-
-# Absolute paths on the guest machine.
-guest_project_dir = '/vagrant'
-guest_drupalvm_dir = ENV['DRUPALVM_DIR'] ? "/vagrant/#{ENV['DRUPALVM_DIR']}" : guest_project_dir
-guest_config_dir = ENV['DRUPALVM_CONFIG_DIR'] ? "/vagrant/#{ENV['DRUPALVM_CONFIG_DIR']}" : guest_project_dir
-
-drupalvm_env = ENV['DRUPALVM_ENV'] || 'vagrant'
-
 # Cross-platform way of finding an executable in the $PATH.
 def which(cmd)
   exts = ENV['PATHEXT'] ? ENV['PATHEXT'].split(';') : ['']
@@ -44,6 +32,12 @@ def walk(obj, &fn)
   end
 end

+drupalvm_env = ENV['DRUPALVM_ENV'] || 'vagrant'
+
+# Default paths when the project is based on Drupal VM.
+host_project_dir = host_drupalvm_dir = host_config_dir = File.dirname(File.expand_path(__FILE__))
+guest_project_dir = guest_drupalvm_dir = guest_config_dir = '/vagrant'
+
 if File.exist?("#{host_project_dir}/composer.json")
   cconfig = {}
   composer_conf = JSON.parse(File.read("#{host_project_dir}/composer.json"))
@@ -51,18 +45,16 @@ if File.exist?("#{host_project_dir}/composer.json")
     cconfig = composer_conf['extra']['drupalvm']
   end

-  # If Drupal VM is a Composer dependency set the correct paths.
+  # If Drupal VM is a Composer dependency set the correct path.
   if Dir.exist?("#{host_drupalvm_dir}/vendor/geerlingguy/drupal-vm")
-    host_project_dir = File.dirname(File.expand_path(__FILE__))
     host_drupalvm_dir = "#{host_project_dir}/vendor/geerlingguy/drupal-vm"
-    host_config_dir = ENV['DRUPALVM_CONFIG_DIR'] ? "#{host_project_dir}/#{ENV['DRUPALVM_CONFIG_DIR']}" : host_project_dir
-    guest_drupalvm_dir = '/vagrant/vendor/geerlingguy/drupal-vm'
+    guest_drupalvm_dir = "#{guest_project_dir}/vendor/geerlingguy/drupal-vm"
   end

   # Read config_dir from composer.json if set.
-  if !ENV['DRUPALVM_CONFIG_DIR'] && cconfig.include?('config_dir')
+  if cconfig.include?('config_dir')
     host_config_dir = "#{host_project_dir}/#{cconfig['config_dir']}"
-    guest_config_dir = "/vagrant/#{cconfig['config_dir']}"
+    guest_config_dir = "#{guest_project_dir}/#{cconfig['config_dir']}"
   end
 end
```

Vagrantfile
```diff
@@ -40,7 +32,32 @@ def walk(obj, &fn)
...
+  composer_conf = JSON.parse(File.read("#{host_project_dir}/composer.json"))
+  if composer_conf['extra'] && composer_conf['extra']['drupalvm']
+    cconfig = composer_conf['extra']['drupalvm']
+  end
+
+  # If Drupal VM is a Composer dependency set the correct path.
+  if Dir.exist?("#{host_drupalvm_dir}/vendor/geerlingguy/drupal-vm")
+    host_drupalvm_dir = "#{host_project_dir}/vendor/geerlingguy/drupal-vm"
+    guest_drupalvm_dir = "#{guest_project_dir}/vendor/geerlingguy/drupal-vm"
+  end
```
oxyc: One annoying thing here is that this wouldn't support forks...

thom8:
> break all existing composer dependency setups

This is actually the reason we are dropping all other installation methods so it's easier/safer to make BC breaks - https://github.com/beetboxvm/beetbox/issues/391#issue-209679900

But yeah these changes are only coming in the next minor version bump.

thom8:
@oxyc there's also a test Class which could be useful -- https://github.com/beetboxvm/beetbox/blob/master/tests/PluginTest.php

oxyc:
- [ ] update docs
- [ ] should we include tests for the composer plugin?
- [x] way more readable logic

oxyc:
> Also, with the Vagrantfile ever-expanding in size, I wonder if it might be good to have a 'ruby-src subfolder somewhere (something like that... whatever the Ruby convention would be) and break out business logic in there?

https://github.com/geerlingguy/drupal-vm/pull/1332

geerlingguy:
@oxyc - Also, note that I think I'd like to include #1332 in a 4.x release, but I'm considering saving this change (along with more of a "this is how you do it _the Drupal VM way_" philosophy) for a 5.0 release. What do you think?

oxyc:
Yeah I agree this should be saved for 5.0.

Btw, is there a way to detect which version you're updating from in a composer plugin? If it is, we could issue a warning stating that the `Vagrantfile` has been overridden and a `config_dir` now needs to be set in `composer.json`.

geerlingguy:
> Is there a way to detect which version you're updating from in a composer plugin?

That... I'm not sure of. I'm pretty new to Composer plugins myself.

oxyc:
I came pretty far and managed to detect updates from <5.0.0 but then I realized that it won't run when the package is first updated, it will run the previous versions code until the next update.

So I went a different, slightly hackier, way.

screen shot 2017-04-23 at 19 10 32

oxyc:
> Also, note that I think I'd like to include #1332 in a 4.x release, but I'm considering saving this change (along with more of a "this is how you do it the Drupal VM way" philosophy) for a 5.0 release. What do you think?

Or we could just aim to release 5.0.0 at or shortly after DrupalCon? Like we did with 3.0.0 last year. Maybe too early for another major version bump...

oxyc:
Regarding docs for this I guess we could entirely delete _Drupal VM as a Composer Dependency_, or move it as a short page under_Getting Started_ section somewhere. In the Installation docs we point people to the _Quick Start Guide_. I guess we should either update the Quick Start Guide, or update these sections with how to set it up a project Drupal VM as a Composer dependency.

Basically the install instructions are

>1. Require Drupal VM as a Composer development dependency.
>
>        composer require --dev geerlingguy/drupal-vm
>
> 2. Create and configure a config.yml file, eg. in `vm/config.yml`.
>
> 3. If you placed the config.yml file in a subdirectory tell Drupal VM about it
>
>         composer config extra.drupalvm.config_dir vm

geerlingguy:
@oxyc - I think one other thing that would be helpful to document somewhere/somehow is how you could still patch Drupal VM using something like https://github.com/cweagans/composer-patches — sometimes people need to either use a change in `master` but not a tagged release, or patch one specific task somewhere for their project...

oxyc:
Maybe add as the first page _Installation / How to integrate Drupal VM_ (I'm terrible at naming things)

Where we point out that there are two ways of integrating Drupal VM with your project.

1. You download Drupal VM and build a quick site within Drupal VM using `vagrant up`. Point out the downsides.

2. You require Drupal VM as a development dependency to your existing project. Show a quick start using `drupal-composer/drupal-project`

Then we also point to this page from the separate Install pages so people consider the options.

geerlingguy:
@oxyc - I'm good with that. As long as you don't make people who use it standalone feel shameful, since that's probably going to still be the default/majority use case for a while until more of the community adopts Composer.

oxyc:
>  that's probably going to still be the default/majority use case for a while

Yeah this would probably still be the first option. What about the Quick Start Guide though? Should that be updated to reflect both or should we keep composer dependency as an _advanced feature_?

> patch Drupal VM using something like https://github.com/cweagans/composer-patches

Mind blown! I hadn't even considered that.

docs/deployment/composer.md
```diff
@@ -1,5 +1,7 @@
 Drupal VM is configured to use `composer create-project` to build a Drupal 8 codebase by default but supports building Drupal from a custom `composer.json` file as well.
 
+_Note that if you already have a project built using a `composer.json` file you should instead add [Drupal VM as a dependency](composer-dependency.md) of your project. The method described below will make it somewhat difficult to manage your `composer.json` together with Drupal VM as the file will be copied from `drupal_composer_path` into `drupal_composer_install_dir` as a secondary `composer.json`._
```
oxyc: Are you okay with this? Honestly I think we should default `drupal_composer_path` to `no` in favor of Drupal VM as a dependency. I think we've seen a few issues due to this particular setup being a bit weird.

oxyc: Or well that's a bit odd too. Maybe it should be renamed to `drupal_composer_copy_path` so it makes more sense as to why you need to set it to `false` for the composer dependency setup.

oxyc:
Do we want to add support for storing configuration values in the `composer.json` [like beetbox](https://github.com/thom8/drupal8-vagrant/blob/master/composer.json#L27)?

We could either a) scaffold the file like they do or b) merge it into `extra_vars` as we already need to pass `config_dir` there anyways.

thom8:
@oxyc we decided to let config in `composer.json` overwrite the normal config file so it wasn't possible to manage config in both places, mainly to reduce the WTF factor...

Otherwise, debugging could get pretty difficult and confusing as to which takes precedence.

When using composer to manage config it's recommended to ignore the normal config files -- https://github.com/thom8/drupal8-vagrant/blob/master/.gitignore

Also run into a few issues with data type conversion from `JSON` to `YAML`, so it's still WIP ;)

oxyc:
> we decided to let config in composer.json overwrite the normal config file so it wasn't possible to manage config in both places, mainly to reduce the WTF factor...

Yeah it could easily get very confusing! However I don't see much reason for the feature when the file needs to be created anyways.  Drupal VM won't be providing defaults so the use case would mainly be forks or downstream projects (such as sprintbox) I guess?

Vagrantfile
```diff
@@ -3,17 +3,28 @@
...
+# Default paths when the project is based on Drupal VM.
+host_project_dir = host_drupalvm_dir = host_config_dir = __dir__
+guest_project_dir = guest_drupalvm_dir = guest_config_dir = '/vagrant'
 
-drupalvm_env = ENV['DRUPALVM_ENV'] || 'vagrant'
+if File.exist?("#{host_project_dir}/composer.json")
+  cconfig = load_composer_config("#{host_project_dir}/composer.json")
+
+  # If Drupal VM is a Composer dependency set the correct path.
+  drupalvm_path = "#{vendor_dir}/geerlingguy/drupal-vm"
```
oxyc: `vendor_dir` being a function seems very unclear to me coming from PHP. Rubocop complains if I add unnecessary parenthesis though.

Vagrantfile
```diff
@@ -3,17 +3,28 @@
 
 require './lib/drupalvm/vagrant'
```
oxyc: Derp this fails now.

Vagrantfile
```diff
@@ -1,19 +1,37 @@
...
+if File.exist?("#{host_project_dir}/composer.json")
+  cconfig = {}
+  composer_conf = JSON.parse(File.read("#{host_project_dir}/composer.json"))
+  if composer_conf['extra'] && composer_conf['extra']['drupalvm']
+    cconfig = composer_conf['extra']['drupalvm']
+  end
+
+  # If Drupal VM is a Composer dependency set the correct path.
+  vendor_dir = composer_conf.fetch('extra', {}).fetch('vendor-dir', 'vendor')
+  drupalvm_path = "#{vendor_dir}/geerlingguy/drupal-vm"
```
oxyc: One problem here. There are some Windows developers who might prefer to code entirely within the VM right? This is why I'm avoiding calls to `composer` without the `which` function which is now in the lib we dont have access to (d'oh).

This setup sort of works, but it still requires the user to run `composer install` to get Drupal VM. That's probably alright, I mean all window devs using this setup would have composer installed, right?

Also, as I'm avoiding calling `composer config` to retrieve the vendor-dir path, this could theoretically be wrong if the user somehow set a custom path using a different method than the `vendor-dir` extra-field.

oxyc: Could also allow the user to specify the path to Drupal VM in the `composer.json`. So that it doesnt have to reside within the vendor-dir. But I doubt anyone would need to use that.

thom8: I like this better than depending on composer being available in the current session.

Also need to consider env vars -- https://getcomposer.org/doc/03-cli.md#composer-vendor-dir

geerlingguy: >  There are some Windows developers who might prefer to code entirely within the VM right?

Yes. (And the occasional Mac or Linux user who likes the isolation.)

> Also this setup requires the user to run `composer install` to download Drupal VM. That's probably alright, I mean all window devs using this setup would have composer installed, right?

Preferably, yes. But right now there are a lot of Windows devs who neither want nor are allowed to set up PHP and/or Composer on Windows... but it could be possible we could make that a requirement for 5.0.0+...

oxyc: > But right now there are a lot of Windows devs who neither want nor are allowed to set up PHP and/or Composer on Windows.

Would these devs still be using Drupal VM as a composer dependency?

thom8: There's also the possibility that someone commits all dependencies and therefore not all Devs would need composer..

thom8: So for example a tech lead might use composer on behalf of the rest of the team and track all dependencies - https://www.codeenigma.com/build/blog/do-you-really-need-composer-production

I personally prefer an isolated build process but some might use DVM as composer dep without composer

docs/deployment/composer-dependency.md
```diff
@@ -1,151 +1,38 @@
...
+    drupal_build_composer_project: false
 
-## Drupal VM in a subdirectory without composer
+    # Tell Drupal VM that the composer.json file already exists and doesn't need to be transfered.
+    drupal_composer_path: false
 
-If you're not using `composer` in your project you can still download  Drupal VM (or add it as a git submodule) to any subdirectory in your project. As an example let's name that directory `box/`.
+    # Set the location of the composer.json file and Drupal core.
+    drupal_composer_install_dir: "/var/www/drupalvm"
+    drupal_core_path: "{{ drupal_composer_install_dir }}/web"
```
oxyc: Would it be too much to print this in the console if Drupal VM is added and no prior `Vagrantfile` exists? It would just make getting started so much easier if you wouldn't have to visit the docs...

geerlingguy:
Just wanted to note that I grabbed some of the work here and dumped it into https://github.com/geerlingguy/drupal-vm-docker, which is more of a quick proof-of-concept just using the Drupal VM Docker container as a starting point.

I am still planning on eventually converting Drupal VM itself into a Composer Plugin—maybe for 5.0.0 still—but I don't think the timing is right yet; it seems the majority of people using Drupal VM still download or git clone it, and aren't ready to take the full Composer plunge.

But we can use https://github.com/geerlingguy/drupal-vm-docker as a testing ground for this, too.

geerlingguy:
Dropping to 6.0.0 milestone. The Drupal community's adoption of Composer is just not there yet to default this thing to Composer-first :(

Issue #1242, opened 2017-03-21
Composer timeouts and slowness due to default use of vboxsf on Windows


fb-multimedia:
### Issue Type

 - Bug Report / Support Request

### Your Environment

```
vagrant 1.9.2
Virtualbox 5.1.10

```

### Your OS

 - Windows 8.1

### Summary

Hello,
I know it's not the first report for this error, but I haven't find solution.
After many try I can't use composer to create a new drupal 8 project :
composer create-project drupal-composer/drupal-project:8.x-dev drupal8 --stability dev --no-interaction
I always have a composer timeout and the message "couldn't delete web/core"
I have try to install in another directory instead of the share folder but same problem.
I have change the COMPOSER_PROCESS_TIMEOUT: 1200
Can somebody confirm that this command actually work good with drupal vm under windows with default share folder ?
Thanks.

geerlingguy:
@fb-multimedia - If you're on an older computer (especially if you don't have an SSD), or if you're using native shared folders, or if you have a slower Internet connection, or if your Internet connection is behind a firewall or proxy, you may have issues with the command.

Are you having trouble running that command manually, or does Drupal VM show a failure during `vagrant up` or `vagrant provision` when it builds a Drupal codebase as per its default configuration?

Did you set the timeout like: `COMPOSER_PROCESS_TIMEOUT=2000 composer create-project drupal-composer/drupal-project...` or some other way?

And does it do anything, or does it just hang after you run the command? If it's doing stuff, and failing at 1200 seconds, then maybe you need to increase the timeout to 2000, 3000, or even more (again, depends on your environment).

fb-multimedia:
Thanks for your answer.
It fails on vagrant up or if I run the command under the VM. Finally the composer command work under /var but not in /var/www (my default shared folder)
So I think the problem come from the shared folder type. 
I have try to change for nfs but It doesn't work (can't see any winnfsd.exe process launch with vagrant-winnfsd installed). 
If I change type to smb vagrant up fails too. Should I install samba on the VM before changing the type on config.yml ? (I haven't found explainations for smb share)
Thanks

geerlingguy:
@fb-multimedia - Ah, in that case it was definitely the shared folder type. The default `vboxsf` type is extremely slow in Windows 7/8 (and even in 10), and that will definitely lead to a Composer timeout.

[...]

PR #1259, opened 2017-03-24
Add docs on how to improve performance by mixing synced folder types

oxyc:
See https://github.com/geerlingguy/drupal-vm/issues/1242#issuecomment-289130181.

How's that? Or maybe under the synced folder section as it's more a feature than a performance improvement suggestion

### DIFF
--- a/docs/other/performance.md
+++ b/docs/other/performance.md
@@ -67,9 +67,10 @@ vagrant_synced_folders:
 
 By default, if you use the _NFS_ synced folder type, Vagrant will ignore this directive and use the native (usually slow) VirtualBox shared folder system instead. You can get higher performance by doing one of the following (all of these steps require a full VM reload (`vagrant reload`) to take effect):
 
+  1. **Use PhpStorm or a reverse-mounted synced folder for working from inside the VM.** Read [Drupal VM on Windows - a fast container for BLT project development](https://www.jeffgeerling.com/blog/2017/drupal-vm-on-windows-fast-container-blt-project-development) for instructions and recommendations.
   1. **Install the `vagrant-winnfsd` plugin**. See the 'NFS' section later for more details and caveats.
-  2. **Use `smb` for the synced folder's type.**
-  2. **Use `rsync` for the synced folder's type.** This requires that you have `rsync` available on your Windows workstation, which you can get if you install a substitute CLI like [Cygwin](https://www.cygwin.com/) or [Cmder](http://cmder.net/).
+  1. **Use `smb` for the synced folder's type.**
+  1. **Use `rsync` for the synced folder's type.** This requires that you have `rsync` available on your Windows workstation, which you can get if you install a substitute CLI like [Cygwin](https://www.cygwin.com/) or [Cmder](http://cmder.net/).
 
 ### NFS
 
--- a/docs/other/performance.md
+++ b/docs/other/performance.md
@@ -41,6 +41,30 @@ vagrant_synced_folders:
 
 _This example assumes that you have Drupal in a directory called `drupal/public`._
 
+### Mixing synced folder types
+
+You can also mix the synced folder types and use a fast one-way rsync for your primary codebase and then a slower but two-way sync for Drupal's configuration sync directory.
+
+```yaml
+vagrant_synced_folders:
+  - local_path: .
+    destination: /var/www/drupal
+    id: drupal
+    type: rsync
+    create: true
+    excluded_paths:
+      - drupal/private
+      - drupal/public/.git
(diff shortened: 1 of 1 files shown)


geerlingguy:
This placement seems appropriate to me.

geerlingguy:
Going to do a little testing before merging, though (I trust you, but I want to see how much of an impact this could make for poor forlorn Windows non-10 devs.

oxyc:
Don't merge just yet btw, I'll test this too and probably change the example to use drupal-project structure `drupal/web` rather than `drupal/public` so it's easier for people to understand.

oxyc:
So it didn't work :) It seems that since we updated the examples to use `local_path: .` the excluded_path examples have been incorrect. You cant reference a parent directory in the path.

oxyc:
These are the rsync executions

```sh
rsync --verbose --archive --delete -z --chmod=ugo=rwX --no-owner --no-group \
--rsync-path "sudo rsync" -e "ssh -p 2200 -o LogLevel=FATAL  -o ControlMaster=auto -o ControlPath=/var/folders/d8/8h53d56x645c5p32t6_44l_40000gn/T/ssh.160 -o ControlPersist=10m  -o IdentitiesOnly=yes -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -i '/Users/cindy/.vagrant.d/insecure_private_key'" \
--exclude ".vagrant/" \
--exclude "private" \
--exclude ".git" \
--exclude "web/sites/default/files" \
--exclude "tmp" \
--exclude "config/drupal" \
/Users/cindy/Projects/Personal/drupal-vm/ vagrant@127.0.0.1:/var/www/drupal
```

```
find /var/www/drupal \
-path /var/www/drupal/private -prune \
-o -path /var/www/drupal/.git -prune \
-o -path /var/www/drupal/web/sites/default/files -prune \
-o -path /var/www/drupal/tmp -prune \
-o -path /var/www/drupal/config/drupal -prune \
-o '!' -type l -a '(' ! -user vagrant -or ! -group vagrant ')' \
-exec chown vagrant:vagrant '{}'
````

**Comment by geerlingguy** [comment]
I'm also investigating doing reverse shared folders (e.g. sync folder from guest to host, using sshfs or other options).

**Comment by yareckon** [comment]
Sharing guest to host is **great** for php performance on large codebases. We do it for some vagrant projects.  I would suggest samba / smb as a protocol since all hosts have passable clients for it.

The drawbacks are:
1. Not having access to the codebase when the VM is shut down (you can't check on something quick without waiting for it to boot)
1. All the compatibility, configuration and performance problems move to the IDE / text editor, which now has to not choke on the "remote" codebase, or use inappropriate features on it like SASS compilation or git management.

**Comment by geerlingguy** [comment]
@yareckon - Especially "performance problems move to the IDE / text editor" — global find-and-replace is an especially slow performer.

**Comment by geerlingguy** [comment]
@oxyc - I'm okay with merging this as-is, if the way it's currently written works.

Maybe we could also add a link to the article [Drupal VM on Windows - a fast container for BLT project development](https://www.jeffgeerling.com/blog/2017/drupal-vm-on-windows-fast-container-blt-project-development), which has some recommendations around reverse-mounting a Samba share, or using SSH sync like with PhpStorm?

File: docs/other/performance.md [hunk 1]
```diff
@@ -67,9 +67,10 @@ vagrant_synced_folders:
...
 
+  1. **Use PhpStorm or a reverse-mounted synced folder for working from inside the VM.** Read [Drupal VM on Windows - a fast container for BLT project development](https://www.jeffgeerling.com/blog/2017/drupal-vm-on-windows-fast-container-blt-project-development) for instructions and recommendations.
   1. **Install the `vagrant-winnfsd` plugin**. See the 'NFS' section later for more details and caveats.
-  2. **Use `smb` for the synced folder's type.**
-  2. **Use `rsync` for the synced folder's type.** This requires that you have `rsync` available on your Windows workstation, which you can get if you install a substitute CLI like [Cygwin](https://www.cygwin.com/) or [Cmder](http://cmder.net/).
+  1. **Use `smb` for the synced folder's type.**
+  1. **Use `rsync` for the synced folder's type.** This requires that you have `rsync` available on your Windows workstation, which you can get if you install a substitute CLI like [Cygwin](https://www.cygwin.com/) or [Cmder](http://cmder.net/).
```
oxyc: I recall a discussion somewhere where you preferred this way of numbering markdown lists.

geerlingguy: Yeah; it's easier to update them without renumbering stuff. Good catch!

Issue #1289, opened 2017-04-12
exportfs: /home/wings/Projects/drupal-vm does not support NFS export


Zorlin:
### Issue Type
 - Bug Report / Support Request

### Your Environment

```
Vagrant 1.9.3
VirtualBox 5.1.18r114002
ansible 2.2.2.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```

### Your OS
 - **Linux** (Ubuntu 16.04 LTS 64-bit)

### Full console output

```
wings@bar:~/Projects/drupal-vm$ vagrant up
Bringing machine 'drupalvm' up with 'virtualbox' provider...
==> drupalvm: Importing base box 'geerlingguy/ubuntu1604'...
==> drupalvm: Matching MAC address for NAT networking...
==> drupalvm: Checking if box 'geerlingguy/ubuntu1604' is up to date...
==> drupalvm: Setting the name of the VM: drupalvm.dev
 ==> drupalvm: Clearing any previously set network interfaces...
==> drupalvm: Preparing network interfaces based on configuration...
    drupalvm: Adapter 1: nat
    drupalvm: Adapter 2: hostonly
==> drupalvm: Forwarding ports...
    drupalvm: 22 (guest) => 2222 (host) (adapter 1)
==> drupalvm: Running 'pre-boot' VM customizations...
==> drupalvm: Booting VM...
==> drupalvm: Waiting for machine to boot. This may take a few minutes...
    drupalvm: SSH address: 127.0.0.1:2222
    drupalvm: SSH username: vagrant
    drupalvm: SSH auth method: private key
==> drupalvm: Machine booted and ready!
[drupalvm] GuestAdditions 5.1.18 running --- OK.
==> drupalvm: Checking for guest additions in VM...
==> drupalvm: Setting hostname...
==> drupalvm: Configuring and enabling network interfaces...
==> drupalvm: Exporting NFS shared folders...
==> drupalvm: Preparing to edit /etc/exports. Administrator privileges will be required...
● nfs-server.service - NFS server and services
   Loaded: loaded (/lib/systemd/system/nfs-server.service; enabled; vendor preset: enabled)
   Active: active (exited) since Thu 2017-04-06 11:38:22 AWST; 19min ago
 Main PID: 3069 (code=exited, status=0/SUCCESS)
   CGroup: /system.slice/nfs-server.service

Apr 06 11:38:22 bar systemd[1]: Starting NFS server and services...
[...]

PR #1290, opened 2017-04-12
Issue #1289: Add encryption as a possible cause for NFS issues on linux

oxyc:
I felt it was too hidden away when added to the numbered list of possible fixes. I think an encrypted home directory is pretty common so rather highlight this as the first debugging step.


geerlingguy:
Definitely helpful. I just need better memory so I can point people to this docs page in the future.

PR #1298, opened 2017-04-14
Read [host-os].config.yml if available

oxyc:
Toyed around with an idea and thought I'd post it for feedback or future reference. This would try and load a host os specific config file so that teams could provide a `windows.config.yml` with

```yaml
vagrant_synced_folder_default_type: "smb"
vagrant_synced_folders:
  - local_path: .
    destination: /var/www/drupal
    type: smb
    create: true
```

Very untested.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -14,6 +14,10 @@ guest_config_dir = ENV['DRUPALVM_CONFIG_DIR'] ? "/vagrant/#{ENV['DRUPALVM_CONFIG
 
 drupalvm_env = ENV['DRUPALVM_ENV'] || 'vagrant'
 
+host_os = %w[darwin bsd freebsd linux solaris windows].reduce(nil) do |detected_os, os|
+  detected_os || (Vagrant::Util::Platform.send("#{os}?") ? os : nil)
+end
+
 # Cross-platform way of finding an executable in the $PATH.
 def which(cmd)
   exts = ENV['PATHEXT'] ? ENV['PATHEXT'].split(';') : ['']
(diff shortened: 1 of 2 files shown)


Vagrantfile
```diff
@@ -14,6 +14,10 @@ guest_config_dir = ENV['DRUPALVM_CONFIG_DIR'] ? "/vagrant/#{ENV['DRUPALVM_CONFIG
 
 drupalvm_env = ENV['DRUPALVM_ENV'] || 'vagrant'
 
+host_os = %w[darwin bsd freebsd linux solaris windows].reduce(nil) do |detected_os, os|
```
oxyc: Vagrant::Util::Platform also has a `cygwin?` method, but that is probably not needed. Need to ensure that `windows?` returns true for cygwin.

provisioning/playbook.yml
```diff
@@ -20,6 +20,7 @@
         - "{{ config_dir }}/config.yml"
         - "{{ config_dir }}/local.config.yml"
         - "{{ config_dir }}/secrets.yml"
+        - "{{ config_dir }}/{{ drupalvm_host_os|default(ansible_system.lower()) }}.config.yml"
```
oxyc: `ansible_system` doesnt entirely match Vagrant, but it still returns the main ones `windows`, `linux` and `darwin`. http://stackoverflow.com/questions/31107019/what-are-the-potential-results-for-platform-system

oxyc:
I'll close this as it's never even been requested and didn't get any reactions since posted.

Issue #1313, opened 2017-04-20
Fix automated test failures 


geerlingguy:
### Issue Type

 - Bug Report

### Full console output

```
TASK [geerlingguy.drupal : Install Drupal with drush.] *************************
ERROR! Unexpected Exception: '0242ac11-0002-c14e-7964-000000002436'
to see the full traceback, use -vvv
```

### Summary

Tests are failing. That makes me sad.

Relevant commit: https://github.com/geerlingguy/drupal-vm/commit/e631891b922afdca4f1d1c3f916ae14ee391e787
Travis failure: https://travis-ci.org/geerlingguy/drupal-vm/builds/224132998

geerlingguy:
The failures were introduce upon updating the Drupal role to `1.9.0`, so it's likely one of those changes (though that particular role is working okay in its own tests... so maybe it's an environment thing?).

geerlingguy:
And of course it _yet again_ reminds me that I should always push a branch, even for what I think are inconsequential changes :P

geerlingguy:
/me needs to give up on this for the day; worst case we can rollback the 1.9.0 drupal role update. But I'd like to find out what's up then push 1.9.1 soon. Note that it worked in my local testing against prod.drupalvm.com :/

oxyc:
This is reproducible with a standard Vagrant setup.

oxyc:
```
TASK [geerlingguy.drupal : Install Drupal with drush.] *************************
task path: /Users/cindy/Projects/Personal/drupal-vm/provisioning/roles/geerlingguy.drupal/tasks/install-site.yml:21
Using module file /usr/local/lib/python2.7/site-packages/ansible/modules/commands/command.py
<127.0.0.1> ESTABLISH SSH CONNECTION FOR USER: vagrant
[...]

PR #1316, opened 2017-04-20
Issue #1313: Redefine geerlingguy.drupal handler due to ansible bug

oxyc:
Let's see if it passes. Maybe better to revert the upstream commit but wanted to check.


geerlingguy:
@oxyc - I'll merge this for now... mostly because I want to keep moving forward with the Drupal role, and the new handler is a nice improvement. It might also be a good idea to consider moving the other prep stuff that runs after general role provisioning into a role (even if it's just a role specific for Drupal VM) so we can run the `drupal` role with the other roles instead of using `include_role`...

Issue #1324, opened 2017-04-22
Problems installing on OS X


justkristin:
### Issue Type

 - Bug Report / Support Request

### Your Environment

```
Vagrant 1.9.3
VirtualBox 5.1.20r114629
-bash: ansible: command not found

```

### Your OS

 - **macOS** (version)
Version 10.12.4 (16E195)

### Full console output

(sorry - not full)
https://gist.github.com/justkristin/39fdf1d83b6bc293bb450d8ec6e713bc

### Summary

The errors are included in the gist pasted above.

I am simply trying to install the drupalvm, but my machine/my brain are lacking somehow. :) The first time there was an error with sendmail install, and the second time it died rather spectacularly at nginx.

justkristin:
Latest attempt:
https://gist.github.com/justkristin/39e15c117953e8180f6aaacd946b7519

thom8:
@justkristin can you try installing [ansible](https://github.com/geerlingguy/drupal-vm#1---install-vagrant-and-virtualbox) and make sure you're using the [latest version](https://github.com/geerlingguy/drupal-vm/releases/tag/4.4.3)

Would also recommend running `vagrant destroy` then `vagrant up` again.

Gugols:
I get the same error and same default apache2 default page. Error during install: https://pastebin.com/Bji1sGcy

I have the following setup:
OSX: 10.10.1 (Yosemite)
Vagrant: 1.9.3
Virtualbox: 5.1.18 r114002 (Qt5.6.2)
Ansible: 2.1.0.0

Gugols:
I just updated Ansible to 2.3.0.0 (since 2.1. didn't meet requirements). Then I destroyed, and rebuilt the box and got the same problem.

However, I changed the variable drupalvm_webserver from apache to nginx and the built was successful. So it seems that for some reason reason it tries to check whether NGINX is is intalled (which it is not since Apache is used) and then fails:

> TASK [geerlingguy.nginx : Ensure nginx is installed.] **************************

justkristin:
I would prefer an apache environment so if Nguni isn't needed can we just comment that line out somewhere?

Also, could you link to an a subtle download page? The only free version I have found is a 30-day free trial version of Tower...?

oxyc:
This is currently a bug. `drupalvm.www` lists `geerlingguy.nginx` as a dependency so it runs just before the `drupalvm.www` role even if `drupalvm_webserver` is set to `apache`.

[...]

PR #1326, opened 2017-04-22
Issue #1324: Remove geerlingguy.nginx hard dependency from drupalvm.www

oxyc:
Or should we use:

```yaml
dependencies:
  - { role: geerlingguy.nginx, when: drupalvm_webserver == 'nginx' }
```

### DIFF
--- a/provisioning/roles/drupalvm.www/README.md
+++ b/provisioning/roles/drupalvm.www/README.md
@@ -12,7 +12,7 @@ There are a few defaults defined, but you shouldn't really need to worry about t
 
 ## Dependencies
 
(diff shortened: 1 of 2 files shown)


geerlingguy:
@oxyc if the conditional works as it looks like it should in your comment, I'd rather go that route, since this role definitely depends on the `nginx` role when the Nginx server is used.

oxyc:
I'll verify it locally once more, but I'm pretty sure it did work when I tested it.

oxyc:
Verified that it works as expected, the tasks are triggered but skipped.

PR #1332, opened 2017-04-23
Vagrantfile cleanup

oxyc:
Inspired by https://github.com/roots/trellis/pull/828.

A first stab at a simple Vagrantfile logic split. Thoughts?

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -74,20 +74,20 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
   # Synced folders.
   vconfig['vagrant_synced_folders'].each do |synced_folder|
     options = {
-      type: synced_folder.include?('type') ? synced_folder['type'] : vconfig['vagrant_synced_folder_default_type'],
+      type: synced_folder.fetch('type', vconfig['vagrant_synced_folder_default_type']),
       rsync__auto: 'true',
       rsync__exclude: synced_folder['excluded_paths'],
       rsync__args: ['--verbose', '--archive', '--delete', '-z', '--chmod=ugo=rwX'],
       id: synced_folder['id'],
-      create: synced_folder.include?('create') ? synced_folder['create'] : false,
-      mount_options: synced_folder.include?('mount_options') ? synced_folder['mount_options'] : []
+      create: synced_folder.fetch('create', false),
+      mount_options: synced_folder.fetch('mount_options', [])
     }
     if synced_folder.include?('options_override')
       synced_folder['options_override'].each do |key, value|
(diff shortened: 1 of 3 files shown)


Vagrantfile
```diff
@@ -1,6 +1,7 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
-VAGRANTFILE_API_VERSION = '2' unless defined? VAGRANTFILE_API_VERSION
```
oxyc: I don't really see the point of defining this when it could technically be overridden by an earlier `vagrant` command.

Vagrantfile
```diff
@@ -74,23 +30,23 @@ else
...
+    ip: vconfig['vagrant_ip'],
+    auto_network: vconfig['vagrant_ip'] == '0.0.0.0' && Vagrant.has_plugin?('vagrant-auto_network')
 
-  if !vconfig['vagrant_public_ip'].empty? && vconfig['vagrant_public_ip'] == '0.0.0.0'
-    config.vm.network :public_network
-  elsif !vconfig['vagrant_public_ip'].empty?
-    config.vm.network :public_network, ip: vconfig['vagrant_public_ip']
+  if vconfig['vagrant_public_ip']
+    config.vm.network :public_network,
+      ip: vconfig['vagrant_public_ip'] != '0.0.0.0' ? vconfig['vagrant_public_ip'] : nil
```
oxyc: My personal preference but I know some prefer to avoid ternary expressions so let me know if you want me to revert it. Also I'm only assuming `nil` will be the same as not setting the option at all. Should verify that.

Vagrantfile
```diff
@@ -137,20 +72,20 @@ Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
...
+      create: synced_folder.fetch('create', false),
+      mount_options: synced_folder.fetch('mount_options', [])
     }
     if synced_folder.include?('options_override')
       synced_folder['options_override'].each do |key, value|
         options[key.to_sym] = value
       end
     end
-    config.vm.synced_folder synced_folder['local_path'], synced_folder['destination'], options
+    config.vm.synced_folder synced_folder.fetch('local_path'), synced_folder.fetch('destination'), options
```
oxyc: As these keys are required, it will now throw a _key not found (KeyError)_ if not set.

geerlingguy:
While we're on the 'potential to break small things' release train... I'll loop this into the 4.5.0 release as well.

Vagrantfile
geerlingguy: +1

PR #1333, opened 2017-04-23
Add docs on using Drupal VM with other application eg Wordpress

oxyc:
As Drupal VM is so flexible I thought it would be useful to point this out in the docs. Included an example to integrate it with the [bedrock project](https://github.com/roots/bedrock).

### DIFF
--- a/docs/other/wordpress-other-applications.md
+++ b/docs/other/wordpress-other-applications.md
@@ -1,0 +1,107 @@
+While Drupal VM caters specifically to Drupal, and support for and compatibility with other PHP applications isn't guaranteed, Drupal VM is flexible enough to work with PHP applications besides Drupal.
+
+## Wordpress
+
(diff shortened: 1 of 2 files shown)


docs/other/wordpress-other-applications.md
```diff
@@ -0,0 +1,107 @@
+Drupal VM is flexible enough to work with other applications than Drupal and you can easily run any PHP application on it.
```
geerlingguy: Might want to make it something along the lines of

> While Drupal VM caters specifically to Drupal, and support for and compatibility with other PHP applications isn't guaranteed, Drupal VM is flexible enough to work with PHP applications besides Drupal.

I want to make it clear that the mission of Drupal VM is primarily to make it the best _Drupal_ dev environment possible. That it makes it a fine _PHP_ development environment is mostly an indication of how well Drupal's been sailing away from it's little island, ala Moana.

oxyc: Updated.

Issue #1335, opened 2017-04-24
Travis fails on geerlingguy.drupal-console : Run Drupal Console init


oxyc:
### Issue Type

 - Bug Report / Support Request

### Your Environment

Travis

### Full console output

> TASK [geerlingguy.drupal-console : Run Drupal Console init.] *******************
fatal: [localhost]: FAILED! => {"changed": true, "cmd": "php /usr/local/bin/drupal init", "delta": "0:00:00.062620", "end": "2017-04-24 01:22:32.724793", "failed": true, "rc": 1, "start": "2017-04-24 01:22:32.662173", "stderr": "", "stderr_lines": [], "stdout": "Could not find Drupal in the current path.", "stdout_lines": ["Could not find Drupal in the current path."]}

### Summary

https://travis-ci.org/geerlingguy/drupal-vm/jobs/225029757#L1304

oxyc:
Restarted a travis build for master to check if it's upstream https://travis-ci.org/geerlingguy/drupal-vm/builds/225003977

oxyc:
https://github.com/hechoendrupal/drupal-console/issues/3279

geerlingguy:
https://github.com/hechoendrupal/drupal-console/issues/3280

geerlingguy:
Haha, @oxyc you beat me by a couple minutes!

jmolivas:
@geerlingguy @oxyc `init` command is no longer valid.

geerlingguy:
Opened https://github.com/geerlingguy/ansible-role-drupal-console/issues/11 upstream in the role repo.

@jmolivas - Is there another command that should be substituted? Or do we not need to run any setup command anymore... just place the .phar in the right location and we're done?

jmolivas:
Exactly @geerlingguy no more setup command to run, since DrupalConsole should be installed per site now.

The `phar` is only a launcher tool now.

ArtuGit:
So we should exclude `drupalconsole` from `installed_extras`?

oxyc:
For now yes.

geerlingguy:
@oxyc - So is the solution we just pull `drupalconsole` from installed extras entirely (and drop the role) and then recommend that if someone wants to use it, they should do so via Composer per Drupal project?

I could then mark the `geerlingguy.drupal-console` role as deprecated, if that's the case.

oxyc:
[...]

PR #1339, opened 2017-04-25
Added drupal console to composer_global_packages.

thom8:
Might still be useful to include the drupal console launcher globally.

Re: #1335

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -262,6 +262,7 @@ composer_home_path: "/home/{{ drupalvm_user }}/.composer"
 composer_home_owner: "{{ drupalvm_user }}"
 composer_home_group: "{{ drupalvm_user }}"
 composer_global_packages:
+  - { name: drupal/console, release: '~1.0' }
   - { name: hirak/prestissimo, release: '^0.3' }
 
 # Run specified scripts before or after VM is provisioned. Path is relative to


geerlingguy:
@thom8 - Even so... I think we've been burned multiple times over supporting Drupal Console as it's gone through dozens of release candidates... I'd rather leave it out as a default, and let projects include it either globally in their override config, or per-project.

The error in this case is:

```
The requested package drupal/console ~1.0 is satisfiable by drupal/console[1.0.0-alpha1, 1.0.0-alpha2, 1.0.0-beta1, 1.0.0-beta2, 1.0.0-beta3, 1.0.0-beta4, 1.0.0-beta5, 1.0.0-rc1, 1.0.0-rc10, 1.0.0-rc11, 1.0.0-rc12, 1.0.0-rc13, 1.0.0-rc14, 1.0.0-rc15, 1.0.0-rc16, 1.0.0-rc17, 1.0.0-rc2, 1.0.0-rc3, 1.0.0-rc4, 1.0.0-rc5, 1.0.0-rc6, 1.0.0-rc7, 1.0.0-rc8, 1.0.0-rc9] but these conflict with your requirements or minimum-stability
``

Since there's no stable release, Composer chokes on that as a default req (see [minimum stability requirements](https://getcomposer.org/doc/04-schema.md#minimum-stability)). And the only way we'd be able to work around that is to hack together some more stuff allowing dev dependencies.

thom8:
Yeah, makes sense to wait for a stable release

PR #1347, opened 2017-04-28
Fixes Error Message "Message: LoadError: cannot load such file -- ./l…

vml-jbartlett:
Fixes Error Message "Message: LoadError: cannot load such file -- ./lib/drupalvm/vagrant" when running vagrant up or destroy.

In this particular instance, the error arose when using a build similar to that described in the [Composer Dependency](http://docs.drupalvm.com/en/latest/deployment/composer-dependency) documentation.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,7 +1,7 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-require './lib/drupalvm/vagrant'
+require_relative 'lib/drupalvm/vagrant'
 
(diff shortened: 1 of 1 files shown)


oxyc:
Oops, seems I was too hasty in my testing! Also fixed in #1256 but that wont get merged until 5.0.0 so this looks good!

geerlingguy:
This will be included with some other recent changes in a 4.5.0 release! Hopefully I'll get that done tomorrow during DrupalCon sprints!

PR #1350, opened 2017-04-28
to pass the Web Server Debug Validation

cdmo:
This solves the `Remote host is configured as 'localhost' despite server host is probably not local` problem in the validator. See http://stackoverflow.com/a/36152477/721065

Forgive me if I didn't follow the editing procedure correctly 😬

### DIFF
--- a/docs/extras/xdebug.md
+++ b/docs/extras/xdebug.md
@@ -14,6 +14,11 @@ To use XDebug with PHPStorm, change the `php_xdebug_idekey` variable as shown be
 ```yaml
 php_xdebug_idekey: PHPSTORM
 ```
+In order to pass the Web Server Debug Validation, you'll need to also add the following your `config.yml`
+
+```yaml
+php_xdebug_remote_host: "{{ vagrant_hostname }}"
+```
 
 ### Sublime Text and XDebug with Drupal VM
 
--- a/docs/extras/xdebug.md
+++ b/docs/extras/xdebug.md
@@ -14,11 +14,6 @@ To use XDebug with PHPStorm, change the `php_xdebug_idekey` variable as shown be
 ```yaml
 php_xdebug_idekey: PHPSTORM
 ```
-In order to pass the Web Server Debug Validation, you'll need to also add the following your `config.yml`
-
-```yaml
-php_xdebug_remote_host: "{{ vagrant_hostname }}"
-```
 
 ### Sublime Text and XDebug with Drupal VM
 
--- a/default.config.yml
+++ b/default.config.yml
@@ -310,6 +310,7 @@ php_xdebug_remote_connect_back: 1
 # Use PHPSTORM for PHPStorm, sublime.xdebug for Sublime Text.
 php_xdebug_idekey: PHPSTORM
 php_xdebug_max_nesting_level: 256
+php_xdebug_remote_host: "{{ ansible_default_ipv4.gateway }}"
 
 # Solr Configuration (if enabled above).
 solr_version: "5.5.3"


oxyc:
Should this be set in `default.config.yml` maybe?

cdmo:
I think that makes sense, yes

oxyc:
Related https://github.com/geerlingguy/drupal-vm/issues/1171#issuecomment-287469544.

I remember 10.0.0.2 had a meaning in VM world but cant recall what right now. Does the hostname work just as well?

joelpittet:
Not sure what the hostname would be but if it resolves to the 'host' michine or that ip it should work.

cdmo:
Ah, that makes sense that the `vagrant_ip` would work too. Not sure if one other is better.

oxyc:
Ah 10.0.0.2 points to the host machine (in some cases, this varies though). @cdmo but for you it works using the guest machine's hostname as well?

joestewart:
We use "php_xdebug_remote_host: 10.0.2.2"

cdmo:
@oxyc yes, I have it pointing to `vagrant_hostname` and it works fine. In my case this happens to be `ul.dev`.  I added my documentation change just to PHPStorm. Not sure how this affects other IDEs. Hard-coding the private  IP address would work, but it should be the same as `vagrant ip` I would think.

geerlingguy:
Yeah, I usually use the hostname and it works okay on my Mac with Sublime Text. Of course IP would be slightly more consistent, but I like the hostname :)

I haven't tried with PHPStorm on my Mac yet (only Windows).

docs/extras/xdebug.md
```diff
@@ -14,6 +14,11 @@ To use XDebug with PHPStorm, change the `php_xdebug_idekey` variable as shown be
 ```yaml
 php_xdebug_idekey: PHPSTORM
 ```
+In order to pass the Web Server Debug Validation, you'll need to also add the following your `config.yml`
+
+```yaml
+php_xdebug_remote_host: "{{ vagrant_hostname }}"
```
geerlingguy: Can you just add this alongside the other xdebug options in the default.config.yml file, and leave out any changes in the docs here?

Also, it'd be good for at least one other person to manually validate the hostname works like we think it does.

thom8: `php_xdebug_remote_host` should be set to the IP/hostname of the host machine.

`vagrant_hostname` is the hostname given to the VM.

With default settings the host IP this is either `10.0.2.2` or `192.168.88.1` but it could be different depending on your network configuration.

oxyc: The host IP definitely varies on my boxes. I think we could use `{{ ansible_eth0.ipv4.address }}` to figure it out but that's probably not very stable.

thom8: Would be nice if Vagrant automatically added a `host` entry to `/etc/hosts` similar to docker.

Might be a Vagrant plugin which does this?

cdmo: I might be misunderstanding @thom8 but DrupalVM endorses `vagrant-hostsupdater` and `vagrant-auto_network` which I use in my local workflow. This completely manages IP assignment and `hosts` updating.

geerlingguy: I'm guessing @thom8 means _inside_ the VM (e.g. the guest); hostsupdater works on the host only.

thom8: @geerlingguy correct, AFAIK there's no stable way to get the IP/hostname of the host (machine running the VM) from _inside_ the VM.

It looks like [vagrant-hostmanager](https://github.com/devopsgroup-io/vagrant-hostmanager) also manages the hosts file of the guest (VM).

cdmo: So, I don't know why `vagrant_hostname` works when the value is supposed to be the IP of the host machine, but it is working for me. `10.0.2.2` does appear to be the IP of the host for me on my Mac too fwiw.

oxyc: fwiw mine is `10.0.2.15` for all 4 of my currently running boxes.

thom8: @cdmo if you're using PHPStorm I don't think you need to set it as the IP is automatically resolved with the network connection of the browser running on the host.

However, I've found you do need to set it when using xdebug in CLI.

joestewart:
Should we just find the default gateway inside vagrant?  Something like one of these?  

    netstat -r |awk '$1=="default"{print $2}'
    route -n | grep 'UG[ \t]' | awk '{print $2}'
    ip route show default | awk '/default/ {print $3}

oxyc:
Derp sorry, mine is actually `10.0.2.2` as well! Here's some of the interesting ansible facts we can use as wlel.

```
drupalvm_project | SUCCESS => {
    "ansible_facts": {
        "ansible_all_ipv4_addresses": [
            "10.0.2.15",
            "192.168.99.88"
        ],
        "ansible_default_ipv4": {
            "address": "10.0.2.15",
            "alias": "enp0s3",
            "broadcast": "10.0.2.255",
            "gateway": "10.0.2.2",
            "interface": "enp0s3",
            "macaddress": "08:00:27:79:98:32",
            "mtu": 1500,
            "netmask": "255.255.255.0",
            "network": "10.0.2.0",
            "type": "ether"
        },
        "ansible_enp0s3": {
            "active": true,
            "device": "enp0s3",
            "features": {},
            "ipv4": {
                "address": "10.0.2.15",
                "broadcast": "10.0.2.255",
                "netmask": "255.255.255.0",
                "network": "10.0.2.0"
            },
            "ipv6": [
                {
                    "address": "fe80::a00:27ff:fe79:9832",
                    "prefix": "64",
                    "scope": "link"
                }
            ],
            "module": "e1000",
            "mtu": 1500,
            "pciid": "0000:00:03.0",
            "promisc": false,
            "speed": 1000,
            "type": "ether"
        },
        "ansible_enp0s8": {
            "active": true,
            "device": "enp0s8",
            "features": {},
            "ipv4": {
                "address": "192.168.99.88",
                "broadcast": "192.168.99.255",
                "netmask": "255.255.255.0",
                "network": "192.168.99.0"
            },
            "ipv6": [
                {
                    "address": "fe80::a00:27ff:fe3d:5c2b",
                    "prefix": "64",
                    "scope": "link"
                }
            ],
            "module": "e1000",
            "mtu": 1500,
            "pciid": "0000:00:08.0",
            "promisc": false,
            "speed": 1000,
            "type": "ether"
        },
        "ansible_interfaces": [
            "lo",
            "enp0s3",
            "enp0s8"
        ],
    },
    "changed": false
}
```

docs/extras/xdebug.md
cdmo: @thom8 I never tested it without setting `php_xdebug_remote_host` mainly because I wanted to and thought I needed to pass the "Web Server Debug Validation" wizard thing. Thanks.

thom8: I believe it is still required for other xdebug clients like sublime, PHPStorm just makes the setup super easy.

thom8:
@oxyc I get  `{{ ansible_default_ipv4.gateway }} == 10.0.2.2`  too.

This is going to save me heaps of time with xdebug CLI setup 😄

cdmo:
@thom8 @oxyc I just used this in my `config.yml`:

`php_xdebug_remote_host: "{{ ansible_default_ipv4.gateway }}"`

And, voila, it worked. Passed the "Web Server Debug Validation" thing too. So, it probably makes more sense to make this part of the `default.config.yml` file.

geerlingguy:
@cdmo - In that case, can you update the PR and I'll go ahead and merge it?

geerlingguy:
Heh... all that for one line changed code :)

cdmo:
This is how the sausage of open source is made. Ain't pretty I guess ;) Thanks.

oxyc:
This should be good to merge no? Beetbox did already.

geerlingguy:
Heh, yes. It just slipped under the radar. I love how the PR ended up having 30+ comments and a few rewrites... and in the end it's a one line change 😆

Issue #1348, opened 2017-04-28
Use more recent Node.js version


yannickoo:
### Issue Type

Feature Idea

### Summary

It feels weird that the default Node.js version in [`default.config.yml`](https://github.com/geerlingguy/drupal-vm/blob/master/default.config.yml#L284-L285) says `0.12`. The current LTS version would be `6.x` so I would suggest to stick to a more recent version so you don't need to increase the version manually. Also mentioning `7.x` in the comment could be helpful.

geerlingguy:
This was fixed by #1353 - I didn't realize it was still stuck back on 0.12!

yannickoo:
Cool 💪

PR #1353, opened 2017-04-30
Update default node version to 6.x

thom8:
Updating default node version to `6.x` which is the same as the [role default](https://github.com/geerlingguy/drupal-vm/blob/master/provisioning/roles/geerlingguy.nodejs/defaults/main.yml).

Therefore, it could potentially be removed from `default.config.yml` but it might be useful to keep so it exposes the config.

Relates to #1348


geerlingguy:
++

PR #1358, opened 2017-05-05
Encapsulate the documentroot in the vhost template

maestrojed:
The apache2 vhosts' documentroot paths were not encapsulated in quotation marks. This would cause an apache2 startup error for paths with spaces or other restricted characters.


oxyc:
You should probably open an issue in the apache role instead https://github.com/geerlingguy/ansible-role-apache

geerlingguy:
@maestrojed - Yes, can you please file a PR against the upstream role so I can merge the fix there and then it will make its way back into Drupal VM?

PR #1384, opened 2017-05-24
Update solr.md

danepowell:
I was confused when trying to set up a local search core, because I used `collection1` as the core name (according to the docs), and then got an error message that schema.xml wasn't created correctly. Turns out the post-provision script creates a new search core `d8` that should be used instead of `collection1`.


geerlingguy:
Good catch—`d8` is the default search core name that the Solr defaults module ships with (which is why I chose that core name for the default).

Issue #1375, opened 2017-05-17
/vagrant default share has no override_options


yareckon:
### Issue Type

 - Bug Report / Support Request

### Your OS

 - **Linux** (Debian Stretch)

### Summary

DrupalVM keeps the default /vagrant share inside the VM, allowing it to pick up the user's configured default type (https://github.com/geerlingguy/drupal-vm/blob/master/Vagrantfile#L95).  There is no chance as far as I can see to add override_options like the custom mounts from the config.yml to the default share.  Why this is important is for instance if you need to switch to nfs version 4 or tcp rather than udp nfs.  Could you a way to either disable or configure the default /vagrant share?  I'm listing this as a bug, because vagrant will break if you have no way to change incompatible mount options.

Thanks for a great project!

oxyc:
You can also set this using a `Vagrantfile.local` for now:

```rb
config.vm.synced_folders['/vagrant'][:nfs_version] = 4
config.vm.synced_folders['/vagrant'][:nfs_udp] = false
```

yareckon:
Thank you for that suggestion -- that looks like my solution for now!

For anyone reaching this issue late, heres' the current (mid-2017) drupalvm doc page on Vagrantfile.local http://docs.drupalvm.com/en/2.5.1/other/local-vagrantfile/

I would still question whether a default mount not visible in config.yml should exist in drupalvm by default.  If it does for reasons of not rocking the vagrant boat, we should at least be able to manipulate its options in the same config files as everything else.

geerlingguy:
> If it does for reasons of not rocking the vagrant boat

That's the main reason, yes.

> We should at least be able to manipulate its options in the same config files as everything else.

Yeah, probably so. One question, though—was it causing provisioning to fail entirely, or was there another reason you needed to change the mount options? (Just want to document an error if there was one so google can pick up this issue for other Linux users).

yareckon:
It was about being able to work on the machine after vagrant up.  After typing in host passwords to mount nfs, there was a hang for about 5 minutes until nfs guest gave up trying to mount.

oxyc:
Related https://github.com/mitchellh/vagrant/issues/2304 and https://www.vagrantup.com/docs/synced-folders/nfs.html#other-notes

geerlingguy:
Fixed via just-merged PR #1392

PR #1392, opened 2017-05-30
Issue #1375: Allow /vagrant synced folder to be overridden

oxyc:
How about leveraging the same logic as we already use for `vagrant_synced_folders`?

```
vagrant_synced_folders:
  - local_path: .
    type: nfs
    create: true
  - local_path: .
    destination: /vagrant
    type: rsync
    options_override: [...]
```

Would need to add docs for this though.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -75,6 +75,14 @@ Vagrant.configure('2') do |config|
     config.hostmanager.aliases = aliases
   end
 
+  # Sync the project root directory to /vagrant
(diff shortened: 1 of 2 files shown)


geerlingguy:
Looks good to me; don't know if we need docs on it? Maybe in the Linux section, a note about adding override options if you hit slowness...

oxyc:
👍 tomorrow I'll add a note in the linux section and run some tests to verify that it works as expected.

geerlingguy:
@oxyc ++ thanks for running with it! I'm knee-deep in some other automation projects currently, but will try to get some time for a few more of these interim improvements while Drupal VM 4.5.x simmers a bit.

Issue #1370, opened 2017-05-15
Instability stymies team use


frob:
### Issue Type

 - Feature Idea
 - Documentation Report

### Your Environment

```
Vagrant 1.9.4
VirtualBox 5.1.22r115126
ansible 2.2.1.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
```

### Your OS

 - **macOS** (10.12.4)

### Summary

Every time I need to provision or start a new project I need to go through at least a half hour of failure, troubleshooting, and updating/downgrading to find the correct version of things.

Is there any type of precheck command or composer plugin that we can use to make the task of version finding a bit easier. It seems like even if I stay on the same major version of DrupalVM I get breaking issues from Vagrant/VirtualBox.

frob:
I honestly don't have much issue making my tools work. My team members don't like using DrupalVM because even though it is working on my machine doesn't mean it will work on their machine.

geerlingguy:
> Is there any type of precheck command or composer plugin that we can use to make the task of version finding a bit easier.

Currently, the presumption is that any versions greater than the minimums defined by Drupal VM (e.g. in `default.config.yml`) should work. Basically, if you run `vagrant up` and _don't_ get a fatal error, everything should work.

That said...

> It seems like even if I stay on the same major version of DrupalVM I get breaking issues from Vagrant/VirtualBox.

Welcome to my special ring of hell :)

Almost every time I think to myself "oh, Vagrant has finally gotten stable for [Ubuntu 16|Ubuntu 14|CentOS 7|CentOS 6|Debian 8]", a new version comes out that breaks something in one of those environments. Or breaks shared folders on Windows, or breaks networking on Mac OS 10.11, but not 10.12. Etc.

And if it's not Vagrant, than VirtualBox releases an update that breaks networking, or provisioning, or shared folders, or something else :P

It's kind of the nature of virtualization that it will never be perfect.

_That said_, I maintain a dozen or so Drupal projects that have Drupal VM integrated via Composer, and as long as I've updated the variables that are listed in `breaking changes` in the CHANGELOG file, they've been working for the past year across my two Macs, a Windows 10 laptop, and my test CI environment (Jenkins on another Mac).

If you can provide more detail on what exactly is breaking, maybe we can find ways to document better workarounds that work in more situations, or find ways to bypass the functionality or software that causes that breaking behavior.

[...]

PR #1401, opened 2017-06-02
Issue #1370: Add VirtualBox version requirement check

oxyc:
As I mentioned in #1370, this is not 100% reliable as it enforces the version requirements as long as VirtualBox is available, regardless if the provider is used or not. 

As there's no way to determine the current provider in a `Vagrantfile`, I don't think we can improve this. Feel free to close if you think the false positives outweigh the benefit.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -39,6 +39,7 @@ end
 # Verify version requirements.
 require_ansible_version ">= #{vconfig['drupalvm_ansible_version_min']}"
 Vagrant.require_version ">= #{vconfig['drupalvm_vagrant_version_min']}"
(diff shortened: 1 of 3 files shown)


geerlingguy:
I'm still torn on this. There are enough reports of weird things happening for people stuck on 5.0.x—but then this might introduce new reports of "I'm not using VirtualBox, why is Drupal VM complaining about it!" and other things like that.

It's kind of six of one, half dozen the other... and closing this out means less code to maintain, so that's one benefit of _not_ including it.

geerlingguy:
Especially if we consider adding Vagrant-based Docker support, and people want to play around with that, but still have an older release of VirtualBox installed for some reason.

PR #1402, opened 2017-06-02
Update default.config.yml add some documentation on vagrant plugins

frob:
Add a bit of documentation around excluding vagrant plugins.

This is in conjunction with @1378

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -33,7 +33,8 @@ vagrant_synced_folders:
 vagrant_memory: 2048
 vagrant_cpus: 1
 
-# Ensure vagrant plugins are installed.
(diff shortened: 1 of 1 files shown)


geerlingguy:
I think it's okay as-is, as it's the same case with any other configuration value inside `config.yml`—installed_extras, apache_vhosts, etc. all work the same, where if you want to override, you have to copy out the entire variable then remove what you don't want, and add what you do.

oxyc:
It's also somewhat documented here http://docs.drupalvm.com/en/latest/getting-started/configure-drupalvm/

> Note: The merge of variables in these files is shallow, if you want to override a single item in a list, you will need to re-define all items in that list.

Issue #1386, opened 2017-05-26
Update example Acquia config file


geerlingguy:
### Issue Type

 - Documentation Report

### Summary

Someone was asking me about updating the example Acquia config.yml file today. It's probably a good idea to take a glance and make sure the settings in that file jibe with what's currently used in Acquia Cloud, as last time I looked at that file was probably a year or more ago, and bits of the platform have changed.

Note that BLT provides its own overrides, so that's a good starting point—or it might be that I just recommend people take a look at BLT's defaults, and take out the example config from Drupal VM (since I don't have much interest in maintaining a config totally separate from BLT).

geerlingguy:
BLT's drop-in `config.yml` is here: https://github.com/acquia/blt/blob/8.x/scripts/drupal-vm/config.yml

geerlingguy:
Fixed by removing example and replacing with documentation in https://github.com/geerlingguy/drupal-vm/pull/1405

PR #1405, opened 2017-06-03
Issue #1386: Remove Acquia example and add docs section pointing to BLT

oxyc:
How's that? BLT has quite a few "extras" so I thought it's still useful to mention what users should be looking for. I've never had the opportunity to test out Acquia Cloud unfortunately so I'm not entirely certain about this PR. Tell me if there's anything I should add/change.

My vote still goes for removing the example and pointing to BLT, partly because I feel examples should be available at http://docs.drupalvm.com/en/latest/.


geerlingguy:
This works for now—maybe at some point in the future, we could rename that doc to "Emulating Cloud provider environments" and expand it to include defaults suggestions for Pantheon, Platform.sh, Blackmesh, etc. (basically any hosting company that has some emphasis on Drupal).

Issue #1411, opened 2017-06-05
.gitignore errors in latest version


inversed:
### Issue Type

 - Bug Report / Support Request

### Your Environment
NOTE: I'm running ansible through the GUEST via the `force_ansible_local: true` setting.
```
Vagrant 1.9.2
VirtualBox 5.1.18r114002
bash: ansible: command not found
```

### Your OS

 - **Windows** (10)

### Full console output

https://gist.github.com/inversed/b469c0e64709b2828f95b735a81aa8b4

### Summary

Drupal-VM's .gitignore excludes a required file (provisioning/tasks/config.yml). 
When trying to do a `vagrant up` on a project where Drupal-VM itself has been included in a repo, that file (and `tests/config.yml`) is missing so the spin up operation fails.

This might not be the intended use case (docs specify running this Windows 10's Bash) but this approach does work fine. Seems like this is a minor bug that's easy to fix but I don't know if it's actually an intentional configuration issue.

geerlingguy:
Technically this is how I originally intended it to work, but I see your point. For applications like BLT, the `box/config.yml` is explicitly committed (e.g. `git commit -f box/config.yml`) and then it's part of the repository.

But it should probably only be excluded in the root directory (for testing or using Drupal VM completely standalone), not in subdirectories.

It looks like @oxyc is working on a fix already 😀

oxyc:
Just trying to find the git version when leading / was added :smile:

inversed:
Thanks for the update and great tool, BTW!

geerlingguy:
That change has been merged in, so it will be part of the next minor release... which I hope to push out this week (was going for the weekend, but life got in the way, so it'll be sometime this week).

PR #1412, opened 2017-06-06
Issue #1411: Fix gitignore applying rules to files in subdirectories

oxyc:
Not entirely sure when this was added to git but it's definitely been in for a while. Somewhere between ~1.9.3 and 2.2.1~

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,14 +16,11 @@ vagrant_ansible_inventory_default
 
 # Drupal VM specific files.
 local.config.yml
-config.yml
-docker-compose.yml
-drupal.make.yml
 Vagrantfile.local
-examples/prod/inventory
-examples/prod/bootstrap/vars.yml
-scripts/
-drupal/
-
-# Some roles include 'config.yml' files.
-!provisioning/roles/**/config.yml
+/config.yml
+/docker-compose.yml
+/drupal.make.yml
+/examples/prod/inventory
+/examples/prod/bootstrap/vars.yml
+/scripts/
+/drupal/


oxyc:
Found the commit message https://github.com/git/git/commit/f87f94974869b1ddd579d4c6aaff12a5f86ca26e. Seems to have been in since forever.

.gitignore
```diff
@@ -15,15 +15,12 @@
...
-drupal.make.yml
-Vagrantfile.local
-examples/prod/inventory
-examples/prod/bootstrap/vars.yml
-scripts/
-drupal/
-
-# Some roles include 'config.yml' files.
-!provisioning/roles/**/config.yml
+/local.config.yml
```
geerlingguy: Since `local.config.yml` should only be used for a local instance configuration... maybe we leave that as globally ignored?

oxyc: Fixed, and `Vagrantfile.local` as well.

geerlingguy:
Since this has the potential to disrupt some projects (since previously-ignored files could be inadvertently committed), I think I may wait until after the next stable release to merge this. Not major enough to warrant a full version bump or anything, but I _am_ thinking it might be best to wait until the next minor version...

geerlingguy:
I might merge this in and bump the minor version since there's now a dozen or so smallish changes... what do you think?

oxyc:
Sounds good. We don't really have any PRs or issues that would warrant a minor version bump so no need to wait I guess. Except for the composer one and that's for 5.0.0

oxyc:
Also the automagic installation of vagrant plugins might justify a minor bump.

geerlingguy:
Yeah, I'm thinking the next release will be a minor bump—a lot of small things that together could cause at least one breaking change in more complex scenarios.

PR #1419, opened 2017-06-09
Issue 1418: Allow default distribution php packages to be used

oxyc:
Built successfully with this `config.yml` on Ubuntu 16.04:

```yaml
---
php_version: ''

installed_extras:
  - adminer
  - blackfire
  # - drupalconsole
  - drush
  # - elasticsearch
  # - java
  - mailhog
  - memcached
  # - newrelic
  # - nodejs
  - pimpmylog
  - redis
  # - ruby
  # - selenium
  # - solr
  - tideways
  - upload-progress
  - varnish
  - xdebug
  # - xhprof # use `tideways` if you're installing PHP 7+
```

That last sentence in the docs might need some tuning.

### DIFF
--- a/provisioning/playbook.yml
+++ b/provisioning/playbook.yml
@@ -40,7 +40,7 @@
     - { role: geerlingguy.apache, when: drupalvm_webserver == 'apache', tags: ['webserver']}
     - { role: geerlingguy.apache-php-fpm, when: drupalvm_webserver == 'apache', tags: ['webserver'] }
     - { role: geerlingguy.nginx, when: drupalvm_webserver == 'nginx', tags: ['webserver'] }
-    - { role: geerlingguy.php-versions, tags: ['php', 'xdebug', 'database'] }
+    - { role: geerlingguy.php-versions, when: php_version != '', tags: ['php', 'xdebug', 'database'] }
     - { role: geerlingguy.php, tags: ['php'] }
     - { role: geerlingguy.php-pecl, tags: ['php'] }
     - { role: geerlingguy.composer }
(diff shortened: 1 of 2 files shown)


oxyc:
Do you think we should add examples instead of just pointing at `geerlingguy.php` (which in fact doesn't really document much about this)? I was avoiding examples as this can change so quickly with PHP 7.2, Ubuntu 16.10 etc etc.

provisioning/playbook.yml
```diff
@@ -40,7 +40,7 @@
     - { role: geerlingguy.apache, when: drupalvm_webserver == 'apache', tags: ['webserver']}
     - { role: geerlingguy.apache-php-fpm, when: drupalvm_webserver == 'apache', tags: ['webserver'] }
     - { role: geerlingguy.nginx, when: drupalvm_webserver == 'nginx', tags: ['webserver'] }
-    - { role: geerlingguy.php-versions, tags: ['php', 'xdebug', 'database'] }
+    - { role: geerlingguy.php-versions, when: php_version != '', tags: ['php', 'xdebug', 'database'] }
```
geerlingguy: Could just do `when: php_version` since it's falsey... I mean that's just as egregious as comparing a version number to a string :D

oxyc: Egregious haha, word of the day. Hmm maybe we should add `php_version|default('')` in the few places we use it? I think it's the dashboard and php extension workspaces that uses it. Then it would work with the value being `false`, `no` and `''`.

Doing `when: php_version` and then later on `workspace: "/root/php{{ php_version }}"` definitely raises an eyebrow for me

geerlingguy: K, so vote to leave as-is?

oxyc: Yes in my opinion, unless we default the variable to an empty string everywhere.

If someone sets `php_version: no` php extensions would be built in `"/root/phpFalse"`

PR #1420, opened 2017-06-09
Update comment in Solr post-provisioning script

eelkeblok:
Still says it is set up for D7, but is set up for D8.


geerlingguy:
Thanks, and good catch!

PR #1426, opened 2017-06-14
Using Vagrant behind corporate proxy

denics:
using the module vagrant-proxyconf it is possible to configure the proxy during the machine start. As a bonus, the fix for bug 589 :)


geerlingguy:
See docs: http://docs.drupalvm.com/en/latest/extending/vagrantfile/#example-using-drupal-vm-behind-a-corporate-proxy-with-vagrant-proxyconf

I'd rather not add the proxy stuff to the global default config, because it's a very small percentage of users who have to deal with all of it.

PR #1428, opened 2017-06-15
Use .localhost instead of .dev as default hostname.

betz:


geerlingguy:
There are a number of other changes that would also need to happen. I think some docs, some test pieces, etc. (to account for a TLD change).

I'm 50/50 on whether I'd go with `localhost` (loooots of letters to type out a zillion times), or with `test` for the default.

betz:
yeah, don't understand they didn't reserve .dev also to be honest.

geerlingguy:
Can we switch to .test? Just because it's only one more character than dev.

geerlingguy:
I'm going to pull in your commit but rebase and merge it to `.test` instead.

default.config.yml
```diff
@@ -8,7 +8,8 @@ vagrant_gui: false
 
 # If you need to run multiple instances of Drupal VM, set a unique hostname,
 # machine name, and IP address for each instance.
-vagrant_hostname: drupalvm.dev
+# rfc2606 limits to only use one of .test, .example, .invalid, or .localhost
```
geerlingguy: Also going to remove this line just because a lot of people (myself included) use something like `local.mysitename.com` for local, and this comment could make that use case confusing.

geerlingguy:
Merged via separate betz-master branch: https://github.com/geerlingguy/drupal-vm/commit/83547893fce01d0628bb8e7f930499f9b980ab57 (and subsequent commit to switch to .test).

Issue #1432, opened 2017-06-16
vagrant exec pwd is /vagrant by default


mikemadison13:
### Issue Type

 - Bug Report / Support Request

### Your Environment

```
Vagrant 1.9.5
VirtualBox 5.1.22r115126
ansible 2.3.0.0
  config file = 
  configured module search path = Default w/o overrides
  python version = 2.7.13 (default, Apr  4 2017, 08:46:44) [GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)]
```

### Your OS

 - **macOS** 10.11.6
### Full console output

```
$ vagrant exec pwd
/vagrant
```

### Summary

This is a cross post from https://github.com/acquia/blt/issues/1681 where running Behat commands with BLT w/in DrupalVM are failing. Digging into this, it seems that my vagrant config is looking in the wrong place (/vagrant vs. /var/www/<project>). As a result, I am getting composer autoload errors. 

It seems like we may need to set the exec pwd path in the VM so that things are running the proper location?

cc @grasmash and @geerlingguy

oxyc:
This isn't configured out of the box, but there's a docs page on it http://docs.drupalvm.com/en/latest/extras/drupal-console/#remote-command-execution-using-vagrant-exec

geerlingguy:
It looks like there's a PR #1435 related to this issue.

I'm still on the fence on whether I want to add explicit conditional config for this as part of Drupal VM. For my own needs, outside of BLT, I've not yet had a use case where vagrant exec was essential... just a nice little tool if you really want to use it. But for most things, using `vagrant ssh` or automating things via Ansible or Drush to work in the VM is easier/faster.

cc @grasmash

geerlingguy:
It looks like this was fixed downstream, in BLT; see https://github.com/acquia/blt/issues/1681 and https://github.com/acquia/blt/commit/1bc9e39e7fef63327d7abfdc1b4acd8905cb4b8c

PR #1435, opened 2017-06-19
Add default directory for `vagrant exec` commands.

grasmash:
This PR adds support for the Vagrant Exec plugin.

By default, using `vagrant exec [command]` will execute a command in the `/vagrant` dir on the VM. This can be problematic for a variety of reasons. 

In short, `/vagrant` isn't the directory that is considered the "drupal" directory. We should use `drupal_composer_install_dir` instead.

This PR will not affect users that do not use Vagrant Exec.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -152,6 +152,10 @@ Vagrant.configure('2') do |config|
       type: vconfig['vagrant_synced_folder_default_type']
     }
   end
+  
+  if Vagrant.has_plugin?('vagrant-exec')
+    config.exec.commands '*', directory: vconfig['drupal_composer_install_dir']
+  end
 
   # Allow an untracked Vagrantfile to modify the configurations
   [host_config_dir, host_project_dir].uniq.each do |dir|


geerlingguy:
See issue: https://github.com/geerlingguy/drupal-vm/issues/1432

Vagrantfile
```diff
@@ -152,6 +152,10 @@ Vagrant.configure('2') do |config|
       type: vconfig['vagrant_synced_folder_default_type']
     }
   end
+  
+  if Vagrant.has_plugin?('vagrant-exec')
+    config.exec.commands '*', directory: vconfig['drupal_composer_install_dir']
```
geerlingguy: This would definitely serve the use case of sites that are managed using the BLT directory structure, or the default for Drupal VM out of the box.

However... if you have a site where you're just sharing the codebase into the VM, and you set that path directly in `config.yml`, the path would likely not be `drupal_composer_install_dir`, and it might not be immediately obvious to the end user why `vagrant exec` isn't working correctly.

In this case, I think that it might make more sense to just mention this configuration in the docs (see: http://docs.drupalvm.com/en/latest/extras/drupal-console/#remote-command-execution-using-vagrant-exec), and then let end users (or something like BLT) add the explicit path to the Vagrantfile.local directly.

I can't think of a way (at least not a simple way) to support the flexibility we currently do (composer project or not, any kind of directory structure, possibility of using Drupal VM with things other than Drupal which would have an entirely different structure, etc.) while forcing the exec root directory to any of the default vars in the Drupal VM config...

geerlingguy:
@grasmash - Were you able to implement this within BLT?

geerlingguy:
Closing as it looks like it was implemented downstream.

Issue #1437, opened 2017-06-19
Allow multiple pre/post provision task directories


oxyc:
### Issue Type

 - Feature Idea

### Summary

Currently we're only allowing users to set one path for pre and post provision tasks. I'm not sure why we did this, but it might be the [vague documentation](http://docs.ansible.com/ansible/playbooks_loops.html#id4).

At least with ansible 2.3.1.0 `with_fileglob` supports both strings and lists. I guess we believed it had to be a list? Both of the following work as expected:

```yaml
    - name: Copy each file over that matches the given pattern
      copy:
        src: "{{ item }}"
        dest: "/etc/fooapp/"
      with_fileglob:
        - "/playbooks/files/fooapp/*"

    - name: Copy each file over that matches the given pattern
      copy:
        src: "{{ item }}"
        dest: "/etc/fooapp/"
      with_fileglob: "/playbooks/files/fooapp/*"
```

I suggest we change how Drupal VM runs the pre/post provision tasks so that user is free to use either a single path as a string or a list to multiple paths.

```yaml
pre_provision_tasks_dir: "{{ config_dir }}/tasks/pre/*.yml" 
post_provision_tasks_dir:
 - "{{ config_dir }}/tasks/post/*.yml" 
 - "{{ config_dir }}/tasks/drupal/*.yml"

geerlingguy:
I like this, but as in the PR, my only concern is that it's compatible with Ansible 2.2+ — otherwise we'll need to version bump the minimum variable in default.config.yml.

geerlingguy:
Merged PR above. Seems like a low-impact feature addition.

PR #1438, opened 2017-06-19
Issue #1437: Add support for list of paths in pre/post_provision_tasks_dir

oxyc:
Might need testing in older versions of Ansible but at least as of 2.3.1.0 this has backwards compatibility. This is also what we added to [`geerlingguy.mac-dev-playbook`](https://github.com/geerlingguy/mac-dev-playbook/pull/43/files) btw.

### DIFF
--- a/provisioning/playbook.yml
+++ b/provisioning/playbook.yml
@@ -25,8 +25,7 @@
       include: "{{ outer_item }}"
       loop_control:
         loop_var: outer_item
-      with_fileglob:
-        - "{{ pre_provision_tasks_dir|default(omit) }}"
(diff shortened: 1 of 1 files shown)


oxyc:
https://github.com/geerlingguy/drupal-vm/pull/623 https://github.com/geerlingguy/drupal-vm/pull/633

geerlingguy:
++ - I'd like to test in 2.2 if we want to keep supporting it for now—that's still the minimum set in default.config.yml right now.

oxyc:
Looking at this unrelated issue https://github.com/ansible/ansible/issues/17136#issuecomment-258616881, it seems to have worked both with lists and with strings at last in 2.1.3

PR #1193, opened 2017-03-02
Fixes #1192: Extract Drupal install and deployment to geerlingguy.drupal role.

geerlingguy:
See #1192 for more info.

PR #1444, opened 2017-06-23
Fix outdated comment for drupal_install_site variable

danylevskyi:
Just found the outdated comment in `default.config.yml`.
The variable has been renamed in #1193


geerlingguy:
Right you are! Thanks.

PR #1446, opened 2017-06-23
Fix linux instructions on how to configure ansible_python_interpreter

oxyc:
While debugging I realized that our instructions for changing the python interpreter won't work anymore as we now load our config files dynamically.


oxyc:
Actually I'm a bit confused by this. This seems to set the python path on the VM, not on the host. https://github.com/geerlingguy/drupal-vm/issues/530

oxyc:
Also it works even if set dynamically.

PR #1474, opened 2017-07-14
Path ammend

kevinhowbrook:
I think the path inside the container needs to be drupal (like with your live example)

### DIFF
--- a/example.docker-compose.yml
+++ b/example.docker-compose.yml
@@ -25,7 +25,7 @@ services:
       - '8.8.4.4'
 
     volumes:
-      - ./:/var/www/drupalvm/:rw,delegated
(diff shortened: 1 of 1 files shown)


geerlingguy:
It depends on how you structure your project. The web root needs to be `/var/www/drupalvm/drupal/web` to be picked up by the default config. So it depends on whether your project directory has `drupal/web`, or is the drupal project directory directly...

I think in my initial testing, I was always testing by downloading Drupal VM, then running things inside the Drupal VM directory (which would then have a `drupal/web` directory).

Can you give a little more detail about how this tripped you up / your use case?

oxyc:
I'll close this PR as there hasn't been any response since July.

Issue #1476, opened 2017-07-17
vagrant up remains on 'Generate Drupal project with composer package'


skorasaurus:
### Issue Type

 - Bug Report / Support Request

### Your Environment

```
Vagrant 1.9.1
VirtualBox 5.1.22r115126
ansible 2.3.1.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
  python version = 2.7.12 (default, Nov 19 2016, 06:48:10) [GCC 5.4.0 20160609]
```

### Your OS

Ubuntu 16.04; 4.10.0-27-generic 

### Full console output

my sys log: https://gist.github.com/skorasaurus/1f9cea059d6e20dac3c2249d6945b09a

my Vbox log: https://gist.github.com/skorasaurus/e3117c413cb1c4d48520dedb6c64be86

my hardware
https://gist.github.com/skorasaurus/59c747f953182eb7ed98bd36947389eb

### Summary

to reproduce on my system: 
clone drupal-vm 
git checkout 7c05f1ffb1b4c5a794e332a79c6b7d70e4c393bd
vagrant plugin install vagrant-vbguest
cd drupal-vm 
vagrant up

while issuing vagrant up for the first time after cloning the repository; there's no movement for more than 30 minutes at: 
```geerlingguy.drupal : Generate Drupal project with composer package```
Compared to #1460 ; 
My system does not crash and is still very responsive; I can use it for routine tasks, web browsing). 

If I edit my [config.yml](https://gist.github.com/skorasaurus/b804afb287465163932409f269c5356b) to as cited at [NFS shares are slow or hang due to UDP protocol](http://docs.drupalvm.com/en/latest/getting-started/installation-linux/#nfs-shares-are-slow-or-hang-due-to-udp-protocol), I am able to provision an environment after about 20 minutes. 

Possibly related to #1468; there's no package named nfs-utils on my system; I do have nfs-common and nfs-kernel-server both at  ```1:1.2.8-9ubuntu12.1```

oxyc:
Great that TCP works! This is another + for https://github.com/geerlingguy/drupal-vm/issues/1468

- related vagrant issue of tcp vs udp https://github.com/mitchellh/vagrant/issues/2304

20min is still slightly slow but I guess that's because it's HDD rather than SSD.

oxyc:
[...]

Issue #1309, opened 2017-04-20
Drupal installation and update is extremely slow


Niklan:
### Issue Type
 - Support Request

### Your Environment

```
Vagrant 1.9.3
VirtualBox 5.0.36_Ubuntur114008
ansible 2.3.0.0
  config file = /etc/ansible/ansible.cfg
  configured module search path = Default w/o overrides
  python version = 2.7.12 (default, Nov 19 2016, 06:48:10) [GCC 5.4.0 20160609]

```

### Your OS

 - **Linux** (4.4.0-72-generic)

```
DISTRIB_ID=neon
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="KDE neon Developer Edition"
NAME="KDE neon"
VERSION="Developer Edition"
ID=neon
ID_LIKE="ubuntu debian"
PRETTY_NAME="KDE neon Developer Edition"
VERSION_ID="16.04"
HOME_URL="http://neon.kde.org/"
SUPPORT_URL="http://neon.kde.org/"
BUG_REPORT_URL="http://bugs.kde.org/"
VERSION_CODENAME=xenial
UBUNTU_CODENAME=xenial
```

### Summary

When I try to `vagrant up` with drush make or compser build and installation, this process take a long time. Drush make is complete around 30-40 minutes, composer throw error about 1 hour working. This is very annying. But not so critical. If i disable build and installation Drupal by DrupalVM, I can copy and paste clean core and install site in 1-2 minutes.

But `drush up -y` executes around 60 minutes on Drupal 8. Holy... why is it so long? I bet this is problem with folder syncing, I uses default settings almost at all, NFS for synced folders and so on.

I noticed that when this commands (drush up, compser create-project) executes so long, they create (i think from archive which they download it)  file by file very slow. So, it's looks like sync problem. I tried to increase memory limit and cpu's cores, but nothing changes. Everything else works very fast.

By the now it's again hanging on such operation:

![screenshot_20170420_170003](https://cloud.githubusercontent.com/assets/2356744/25229725/d54dd9a0-25ea-11e7-8627-7e069d8dcc00.png)

My $10 DigitalOcean droplet with other projects on it do same operation for 10-20 seconds!

P.s. Example above finally finished in 90 minutes! And with error: `2006 MySQL server has gone away`

thom8:
[...]

PR #1477, opened 2017-07-18
[test] Improve shared folder performance by generating drupal project in /tmp

oxyc:
Issue #1476 #1309

_Do not merge_

### DIFF
--- a/provisioning/roles/geerlingguy.drupal/tasks/build-composer-project.yml
+++ b/provisioning/roles/geerlingguy.drupal/tasks/build-composer-project.yml
@@ -25,9 +25,9 @@
   when: not drupal_site_exists
   failed_when: false
 
-- name: Move Drupal project files to drupal_composer_install_dir
+- name: Move Drupal project files to drupal_composer_install_dir.
   command: >
-    cp -r /tmp/drupal-project/ {{ drupal_composer_install_dir }}/
+    cp -r /tmp/drupal-project/. {{ drupal_composer_install_dir }}/
     creates={{ drupal_core_path }}/index.php
   become: no
   when: not drupal_site_exists
--- a/provisioning/roles/geerlingguy.drupal/tasks/build-composer-project.yml
+++ b/provisioning/roles/geerlingguy.drupal/tasks/build-composer-project.yml
@@ -1,4 +1,20 @@
 ---
+- name: Delete old /tmp/drupal-project
+  file:
+    path: "/tmp/drupal-project"
+    state: absent
+  when: not drupal_site_exists
+
+- name: Generate Drupal project with composer package in /tmp/drupal-project.
+  command: >
+    {{ composer_path }} create-project
+    {{ drupal_composer_project_package }} /tmp/drupal-project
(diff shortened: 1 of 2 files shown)


provisioning/roles/geerlingguy.drupal/tasks/build-composer-project.yml
```diff
@@ -1,4 +1,20 @@
 ---
+- name: Delete old /tmp/drupal-project
+  file:
+    path: "/tmp/drupal-project"
+    state: absent
+  when: not drupal_site_exists
```
oxyc: I run this before the other tasks so that it's always purged even if a followup task aborts the process.

provisioning/roles/geerlingguy.drupal/tasks/build-composer-project.yml
```diff
@@ -9,15 +25,12 @@
...
-    {{ composer_path }} create-project
-    {{ drupal_composer_project_package }} {{ drupal_composer_install_dir }}
-    {{ drupal_composer_project_options|default('--prefer-dist --no-interaction') }}
-  when: not drupal_site_exists
+    cp -r /tmp/drupal-project/ {{ drupal_composer_install_dir }}/
+    creates={{ drupal_core_path }}/index.php
   become: no
-  environment:
-      COMPOSER_PROCESS_TIMEOUT: 1200
+  when: not drupal_site_exists
```
oxyc: As the destination is a NFS shared directory `mv` would not be faster according to [this](http://rothwerx.com/content/simple-speed-comparison-between-cp-mv-and-rsync). Makes sense also. 

`synchronize` module is just too broken, I can never get it working when it's remote -> remote :/

provisioning/roles/geerlingguy.drupal/tasks/build-composer-project.yml
oxyc: If we want to avoid being destructive at all cost we would need an additional `fail` task that checks if `drupal_site_exists` is false and `drupal_composer_install_dir` is not empty. Personally I feel `cp` is safe enough as eg `.git` directory wont be removed, nor any new files.

geerlingguy: As long as the .git directory's not removed, I think this is safe enough. Worst case someone deletes index.php, runs this, finds site in a weird state, then has to go in and `git reset --hard` then probably a `composer install` by hand.

Maybe I'll regret these words someday... but I find that a pretty edge-casey scenario.

provisioning/roles/geerlingguy.drupal/tasks/build-composer-project.yml
geerlingguy: Yeah, `synchronize` is a little iffy.

geerlingguy:
Since there does seem to be some performance benefit, I'm good with this; would need to stick a PR upstream of course.

geerlingguy:
Closing this PR—redirecting efforts upstream in https://github.com/geerlingguy/ansible-role-drupal/pull/51

PR #1485, opened 2017-07-21
Use absolute paths for example post_provision_scripts

oxyc:
I think knowing about `{{ playbook_dir }}` and `{{ config_dir }}` makes it easier for users to reason about the paths. Counting parent directories is a headache.


geerlingguy:
Sounds good to me. And yeah, once you go more than one level of `..`, it gets a bit annoying.

PR #1505, opened 2017-08-06
Updates behat.yml parameters in selenium.md for javascript scenarios.

bt-eric:
I needed to add these parameters to behat.yml to use the chromedriver for behat scenarios tagged with `@javascript`.

### DIFF
--- a/docs/extras/selenium.md
+++ b/docs/extras/selenium.md
@@ -47,6 +47,10 @@ Using the default Drupal site as an example (it's installed in `/var/www/drupalv
               javascript_session: selenium2
               selenium2:
                 wd_host: http://drupalvm.dev:4444/wd/hub
+                capabilities:
+                  chrome:
+                    switches: ['no-sandbox']
+              browser_name: chrome
               base_url: http://drupalvm.dev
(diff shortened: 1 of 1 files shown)


geerlingguy:
@bt-eric - Thanks for the PR! I am wondering, does this make the tests work out of the box with Headless Chrome? Also cc/ @grasmash in case he's interested.

bt-eric:
The `switches: ['no-sandbox']` option still needs xvfb and, AFAIK, is not using Chrome's headless feature. Without the no-sandbox argument my `@javasript` tests would hang and eventually time out. I got the idea to try the no-sandbox option from [here](https://sites.google.com/a/chromium.org/chromedriver/help/chrome-doesn-t-start). I know that the option prevents Chrome from sandboxing processes that are normally sandboxed, but I'm not sure why it is actually needed here for the tests to work.

I find that if I use `switches: ['headless', 'disable-gpu', 'window-size=1920,1080']` I don't need xvfb (or no-sandbox). However, Selenium2Driver's implementation of resizeWindow doesn't seem to work with headless Chrome.

geerlingguy:

Issue #1522, opened 2017-08-20
Skip need for vault pass on local dev provision


iainhouston:
### Issue Type

 - Bug Report / Support Request
 

### Your Environment

```
Vagrant 1.9.7
VirtualBox 5.1.26r117224
ansible 2.3.2.0
  config file = 
  configured module search path = Default w/o overrides
  python version = 2.7.13 (default, Feb 25 2017, 15:14:13) [GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.42.1)]
```

### Your OS

 - **macOS** (version)

### Full console output

```
TASK [Include optional configuration files.] ***********************************
ok: [local-misfits] => (item=/Users/iainhouston/WebDev/misfitstheatre.com/vm/config.yml)
failed: [local-misfits] (item=/Users/iainhouston/WebDev/misfitstheatre.com/vm/secrets.yml) => {"ansible_facts": {}, "failed": true, "item": "/Users/iainhouston/WebDev/misfitstheatre.com/vm/secrets.yml", "message": "Decryption failed on /Users/iainhouston/WebDev/misfitstheatre.com/vm/secrets.yml"}
ok: [local-misfits] => (item=/Users/iainhouston/WebDev/misfitstheatre.com/vm/vagrant.config.yml)
	to retry, use: --limit @/Users/iainhouston/WebDev/misfitstheatre.com/vendor/geerlingguy/drupal-vm/provisioning/playbook.retry
```
### Summary

Following "Soup to Nuts" and having added `vm/secrets.yml` deploys to prod just great. But now, doing `vagrant provision` on the local dev site we fail because we have not been given the opportunity to provide vault password as we were when invoking `ansible-playbook` with `--ask-vault-pass` .

But we really need to skip the need for  `secrets.yml` when we're provisioning local dev?

iainhouston:
In `vendor/geerlingguy/drupal-vm/provisioning/vars/main.yml` I changed the `secrets.yml` line as follows: 

```
- name: Include optional configuration files.
  include_vars: "{{ item }}"
  with_fileglob:
    - "{{ config_dir }}/config.yml"
    - "{{ config_dir }}/local.config.yml"
    - "{{ config_dir }}/{{ lookup('env', 'DRUPALVM_ENV')|default(drupalvm_env, true)|default(ansible_env.DRUPALVM_ENV)|default(omit) }}secrets.yml"
[...]

PR #1523, opened 2017-08-20
distinguish between local and prod secrets

iainhouston:


provisioning/tasks/config.yml
```diff
@@ -12,5 +12,5 @@
   with_fileglob:
     - "{{ config_dir }}/config.yml"
     - "{{ config_dir }}/local.config.yml"
-    - "{{ config_dir }}/secrets.yml"
+    - "{{ config_dir }}/{{ lookup('env', 'DRUPALVM_ENV')|default(drupalvm_env, true)|default(ansible_env.DRUPALVM_ENV)|default(omit) }}.secrets.yml"
```
oxyc: Could you keep the old behavior as well as we talked about in https://github.com/geerlingguy/drupal-vm/issues/1522#issuecomment-333314535

oxyc: True, that sounds like a good solution but I do see a use case for separate `<env>.secrets.yml` files, eg separate staging/prod environments. But maybe we shouldn't add more features unless someone actually asks for it.

geerlingguy: It would be nice to have... but if it makes things harder, let's keep the current level of support only.

geerlingguy:
See comments on #1522

PR #1527, opened 2017-08-22
Reorganization of the dashboard with small styling update.

petrillek:
Hi Jeff,
firstly thank you for all your energy you put into this project. 
We have recently switched to it and so far it is really great experience.

As time goes I've added multiple sites with db's to my VM and the dashboard start to look full and confusing. I've also realized that parts of the dashboard are used more often then others.

This pull request is only changing the dashboard template (provisioning/templates/dashboard.html.j2).

## Provided changes (see attached comparison screenshots):
- Make all tables responsive, to prevent overlapping of content even on desktop.
- Add stripes to some tables.
- Fixed logo position/size on small screen.
- Cleaner header.
- First row – Site and DB list's. 
- DB list is much simpler.
- Second row – Technical info (DB Info, PHP info, Development tools).
- Third row – Host file content.
- All new styling is done only with addition Bootstrap classes.

Hope you will find this at least partially useful.

Petr

drupal-vm-dashboard-desktop

drupal-vm-dashboard-small-screen

### DIFF
--- a/provisioning/templates/dashboard.html.j2
+++ b/provisioning/templates/dashboard.html.j2
@@ -279,6 +279,15 @@
       </div>
     {%- endmacro -%}
 
+    {%- macro SectionDatabaseListLabel() -%}
+      {% if 'adminer' in installed_extras -%}
+        Databases using Adminer
+      {% else -%}
+        Databases
+      {%- endif %}
+    {%- endmacro -%}
+
+
     {%- macro sectionDatabaseUserList() -%}
       <div class="table-responsive">
         <table class="table table-striped table-hover">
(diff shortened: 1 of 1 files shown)


geerlingguy:
A cursory glance looks good to me. I'd like to run it myself once just to be sure everything functions the same, but assuming it does, I'll merge into next minor release.

petrillek:
I just realize that it is probably better to move the condition for Database List label between the macros leaving the actual template cleaner.

geerlingguy:
Looks good to me, thank you for your work on sprucing up the template!

PR #1528, opened 2017-08-23
Enable linked_clone on versions greater than 1.8.

thom8:
Linked clones are only enabled on vagrant versions `1.8.x`, updates to enable of all versions greater than `1.8`

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -124,7 +124,7 @@ Vagrant.configure('2') do |config|
 
   # VirtualBox.
   config.vm.provider :virtualbox do |v|
-    v.linked_clone = true if Vagrant::VERSION =~ /^1.8/
+    v.linked_clone = true
     v.name = vconfig['vagrant_hostname']
     v.memory = vconfig['vagrant_memory']
     v.cpus = vconfig['vagrant_cpus']


geerlingguy:
Just because it's late and I don't have the proper context—what wasn't working with the current syntax (`v.linked_clone = true if Vagrant::VERSION =~ /^1.8/`)?

geerlingguy:
I mean, technically I think Drupal VM now requires 1.8.x+... so we could just drop the entire conditional and make it true in all cases :)

thom8:
It wasn't matching on `1.9.0` and above but happy to update to always `true`

geerlingguy:
@thom8 ah, I see. Yeah, let's just go with `true` now, since anyone running < 1.8 would hit some weird issues otherwise.

thom8:
All done and passing.

oxyc:
ping @geerlingguy please merge this PR if you're planning on a release 😄 

```
~: df -h
Filesystem      Size   Used  Avail Capacity iused      ifree %iused  Mounted on
/dev/disk1     233Gi  226Gi  6.3Gi    98% 3317028 4291650251    0%   /
```

geerlingguy:
/me promises to tag a release soon. Finally had a couple days with nothing in prod environments blowing up, nor DrupalCamps that needed tending to :)

Issue #1548, opened 2017-09-17
Support alternatives when auto-installing plugins


hussainweb:
### Issue Type

 - Feature Idea

### Your Environment

```
Vagrant 2.0.0
VirtualBox 5.1.24r117012
ansible 2.2.0.0
  config file =
  configured module search path = Default w/o overrides
```

### Your OS

 - **macOS** Sierra 10.12.6

### Summary

DrupalVM now auto-installs plugins by default. The problem I face is with `vagrant-hostsupdate` which I don't really need (I use `vagrant-hostmanager` which works well). But since it auto-installs, I need to remember to either remove it from default.config.yml or override it if I am using a composer setup. This is quite irritating.

What is ideal is to allow specifying alternative plugins. I have a PR ready for that. Since Ruby isn't my first language, I would appreciate some code review. I believe it can be done in a much better way than what I have tried. I have tested my changes and they work.

geerlingguy:
This is already supported by the simple case of overriding in your `config.yml` (or `local.config.yml`):

```yaml
vagrant_plugins:
  - name: vagrant-vbguest
  # - name: vagrant-hostsupdater
  - name: vagrant-hostmanager
```

I'd rather not add additional complexity to this part of the configuration, especially since it should be fairly easy to override per-project and/or per-developer.

oxyc:
Hmm just saw this https://github.com/roots/trellis/pull/892 which essentially makes `vagrant-auto_network` unnecessary and multiple boxes work out of the box. It requires `vagrant-hostmanager` though.

hussainweb:
Thanks @geerlingguy. I understand what you mean about complexity. Alternatively, would you consider removing `vagrant-hostsupdate` from the default config? I have to remember to remove this line from every new project or else deal with uninstalling the hostsupdater. It is just really annoying that it checks for specific plugins when DrupalVM actually supports the alternative plugin as well.

geerlingguy:
If there's a compelling technical reason to switch from hostsupdater to hostsmanager, I'm willing to make that change. But I'm 100% behind keeping at least one or the other in the default config. Huge help for first-time users to not have to get all gooky managing hosts file.

hussainweb:
I see your point. The reason I use hostmanager is that it is more feature complete. It manages hosts entries on both host and guest (all guests), and I like how it adds entries in blocks in the hosts file which makes it easier to manage manually if I have to.

[...]

PR #1549, opened 2017-09-17
Support alternatives when auto-installing plugins

hussainweb:
Fixes issue #1548.

DrupalVM auto-installs plugins by default. The problem I face is with `vagrant-hostsupdate` which I don't really need (I use `vagrant-hostmanager` which works well). But since it auto-installs, I need to remember to either remove it from default.config.yml or override it if I am using a composer setup. This is quite irritating.

What is ideal is to allow specifying alternative plugins. I have a PR ready for that. Since Ruby isn't my first language, I would appreciate some code review. I believe it can be done in a much better way than what I have tried. I have tested my changes and they work.

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -43,7 +43,9 @@ vagrant_cpus: 1
 # Ensure vagrant plugins are installed.
 vagrant_plugins:
   - name: vagrant-vbguest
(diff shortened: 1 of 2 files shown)


geerlingguy:
Closing this PR—please see #1549.

Issue #1468, opened 2017-07-06
Latest `nfs-utils` doesn't work out of the box with Vagrant


oxyc:
### Issue Type

 - Bug Report / Support Request

### Summary

As of [`nfs-utils-2-1-2-rc2`](http://git.linux-nfs.org/?p=steved/nfs-utils.git;a=commit;h=fbd7623dd8d5e418e7cb369d4026d5368f7c46a6) ([discussion](https://patchwork.kernel.org/patch/9591703/)) UDP is disabled by default thus requiring the user to either:

1.  enable udp in `/etc/nfs.conf`
2. change the nfs options in Drupal VM's `config.yml`

At the moment this only affect Fedora 25+ but we _have had_ issues regarding this before. Not sure what prompted them then as the the repo sources do specify versions that should have udp and nfsv3 support.

Some `nfs-utils` versions across distros.

- [x] Ubuntu 16.04: UDP enabled `nfs-utils v1.2.8` [ref](https://launchpad.net/ubuntu/xenial/+source/nfs-utils)
- [ ] Fedora 25, 26: UDP disabled `nfs-utils v2.1.1-5.rc4+` [ref](https://apps.fedoraproject.org/packages/nfs-utils)
- [x] Arch Linux: UDP enabled `nfs-utils v2.1.1-4` [ref](https://www.archlinux.org/packages/core/i686/nfs-utils/)
- [x] Debian 9: UDP enabled `nfs-utils v 1.3.4-2.1` [ref](https://packages.debian.org/search?searchon=sourcenames&keywords=nfs-utils)

I think it's still fairly safe to default to nfsv4 on Linux hosts as they mention in the commit that it's been supported for a decade. A possible fix:

```rb
   if options[:type] == 'nfs' && !options.key?(:nfs_version) && Vagrant::Util::Platform.linux?
     options[:nfs_version] = 4
     options[:nfs_udp] = false
   end
```

Related:
- https://github.com/geerlingguy/drupal-vm/issues/1040#issuecomment-313215521
- https://github.com/geerlingguy/drupal-vm/issues/245
- https://github.com/geerlingguy/drupal-vm/issues/1425#issuecomment-310796259
- https://github.com/geerlingguy/drupal-vm/issues/1375
- https://github.com/geerlingguy/drupal-vm/issues/893

oxyc:
@yareckon mind sharing why you changed the nfs version? Was it because it didn't work out of the box, or something else?

[...]

Issue #1381, opened 2017-05-24
Consider allowing use of Vagrant with docker provider to run and provision Drupal VM


geerlingguy:
### Issue Type

 - Feature Idea

### Summary

In #1366, we added the ability to use Drupal VM in a monolithic Docker container, completely independent of Vagrant. Vagrant itself allows the use of the [Docker provider](https://www.vagrantup.com/docs/docker/configuration.html) in lieu of other providers (the most commonly used for Drupal VM is currently VirtualBox...).

After we solidify the experimental Docker support, it might be a nice improvement to let Vagrant manage the building/provisioning of Docker containers, especially if it solves some of the more annoying/difficult bits, like managing the hosts file using `vagrant-hostsupdater`.

Tagging @thom8, who was the original suggester of such an integration.

oxyc:
Played around with this a bit during the weekend. A few notes:

- Vagrants synced folders do not work (https://github.com/hashicorp/vagrant/issues/8021). Dockers is the way to go anyways.
- `vagrant-hostsupdater` correctly edits the hostfile while `vagrant-hostmanager` doesn't (https://github.com/devopsgroup-io/vagrant-hostmanager/issues/243)
- Pretty flexible as we can map the current config to docker params ala `d.ports = vconfig['firewall_allowed_tcp_ports'].map { |p| "#{p}:#{p}" }`
- Networking is a pain on macos ofc. Really hoping for an upstream fix here as it's still sort of deal breaker for me.

geerlingguy:
Closing this out; if I go further down the Docker path, I go further down Docker alone since it's faster/easier to manage than Vagrant with Docker inside at this point, for all things but the networking setup.

Issue #1126, opened 2017-01-15
Add docs on how to build your own pre-provisioned base box


oxyc:
### Issue Type

 - Documentation Report

### Summary

My laptop just ran out of space (`71 GiB	~/VirtualBox VMs`) so I thought it might be a good idea to build a pre-provisioned box and leverage linked clones. As Drupal VM is so awesome for teams, might this be something worth adding to the docs?

I've never used packer or looked into building boxes, but it looks easy enough and quite useful. I know other co-workers have complained about the VM disk usage..

geerlingguy:
"But 512 K is enough for everybody, right?" /s

In all seriousness, though... basically something like "how to build and host your own derivative box" would be a decent addition. Or if not that, at least a blog post, akin to one I did a while back about Packer + Ansible: [Building VM images with Ansible and Packer](http://www.jeffgeerling.com/blog/server-vm-images-ansible-and-packer).

Or more specifically, how to optimize a project for your team even _further_ than something like [Beetbox](https://github.com/beetboxvm/beetbox).

oxyc:
Thanks for the link! Currently I'm just testing it out by building our boilerplate project together with a post-provision task that runs:

```sh
#!/bin/bash

rm -f /home/vagrant/.ssh/known_hosts
apt-get clean
dd if=/dev/zero of=/EMPTY bs=1M
rm -f /EMPTY
cat /dev/null > ~/.bash_history
```

And then I was planning to run:

```sh
vagrant package --output mybox.box
```

oxyc:
The simple solution works great. To make it maintainable I guess I do need packer though.

oxyc:
For packer I guess the easiest is to fork on your box repositories and configuring it to run Drupal VM's playbook with the instructions from _Deploying Drupal VM to a production environment_.

@geerlingguy feel free to close this if you think it's better suited as a blog post or in general too far from the common use case.

geerlingguy:
Not necessarily... it might still be suited for a blog post, but we can explore the topic in this issue a bit more before closing it out.

For months, I've been wanting to do basically what you've suggested:

> fork one of your box repositories and configuring it to run Drupal VM's playbook with the instructions from Deploying Drupal VM to a production environment

But make it into a digestible and as-simple-as-possible process so it's really easy for people to do it on their own.

It would also alleviate some of my bandwidth, for which I'm now up to about $25/month just to host my boxes 😳

thom8:
[...]

PR #1559, opened 2017-09-22
Support vagrant DHCP IP assignment with `vagrant-hostmanager`

oxyc:
See https://github.com/geerlingguy/drupal-vm/issues/1548#issuecomment-330365797.

This needs to be thoroughly tested but in my opinion this classifies as a _"compelling technical reason"_ for a switch to `vagrant-hostmanager`.

- If `vagrant_ip` is set to the new default `dhcp`, the VM will be assigned an IP with DHCP and configured using `vagrant-hostmanager`.
- If a static IP is set, it will be kept but `vagrant-hostmanager` will be installed and used instead.
- If `0.0.0.0` is used, `vagrant-auto_network` will continue to be used.

This is what my hosts file looks like after bringing up two new VMs:

```
192.168.88.88  drupalvm.test  # VAGRANT: a97adfc1a53d882e0a42473a42d1ec66 (drupalvm) / ec047028-5e0a-49fd-bd7a-f0472f127d6c
192.168.88.88  www.drupalvm.test  # VAGRANT: a97adfc1a53d882e0a42473a42d1ec66 (drupalvm) / ec047028-5e0a-49fd-bd7a-f0472f127d6c
192.168.88.88  adminer.drupalvm.test  # VAGRANT: a97adfc1a53d882e0a42473a42d1ec66 (drupalvm) / ec047028-5e0a-49fd-bd7a-f0472f127d6c
192.168.88.88  xhprof.drupalvm.test  # VAGRANT: a97adfc1a53d882e0a42473a42d1ec66 (drupalvm) / ec047028-5e0a-49fd-bd7a-f0472f127d6c
192.168.88.88  pimpmylog.drupalvm.test  # VAGRANT: a97adfc1a53d882e0a42473a42d1ec66 (drupalvm) / ec047028-5e0a-49fd-bd7a-f0472f127d6c
192.168.88.88  dashboard.drupalvm.test  # VAGRANT: a97adfc1a53d882e0a42473a42d1ec66 (drupalvm) / ec047028-5e0a-49fd-bd7a-f0472f127d6c

## vagrant-hostmanager-start id: 507bfe16-8e86-4df9-b21a-fcac1af63073
172.28.128.3	drupalvm-two.test
172.28.128.3	www.drupalvm-two.test
172.28.128.3	adminer.drupalvm-two.test
172.28.128.3	xhprof.drupalvm-two.test
172.28.128.3	pimpmylog.drupalvm-two.test
172.28.128.3	dashboard.drupalvm-two.test
## vagrant-hostmanager-end

## vagrant-hostmanager-start id: 80c4c10c-3b29-4c0c-b0f1-91aa3b4e0811
172.28.128.4	drupalvm-foo.test
172.28.128.4	www.drupalvm-foo.test
172.28.128.4	adminer.drupalvm-foo.test
172.28.128.4	xhprof.drupalvm-foo.test
172.28.128.4	pimpmylog.drupalvm-foo.test
172.28.128.4	dashboard.drupalvm-foo.test
## vagrant-hostmanager-end
```

It's noteworthy to mention that `vagrant-hostmanager` doesn't clean up the host entries when you run `vagrant halt`. They're only cleared out on `vagrant destroy` which is a better behavior as the IP wont change constantly.

@hussainweb also mentioned

> The reason I use hostmanager is that it is more feature complete. It manages hosts entries on both host and guest (all guests), and I like how it adds entries in blocks in the hosts file which makes it easier to manage manually if I have to.

If we decide to go with this, [geerlingguy.mac-dev-playbook](https://github.com/geerlingguy/mac-dev-playbook) also needs to be updated for passwordless commands.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -48,13 +48,40 @@ Vagrant.configure('2') do |config|
 
   # Networking configuration.
   config.vm.hostname = vconfig['vagrant_hostname']
-  config.vm.network :private_network,
-    ip: vconfig['vagrant_ip'],
-    auto_network: vconfig['vagrant_ip'] == '0.0.0.0' && Vagrant.has_plugin?('vagrant-auto_network')
 
+  if vconfig['vagrant_ip'] == 'dhcp'
+    # Automatically assign IP
+    config.vm.network :private_network, type: 'dhcp'
+
+    # Use assigned IP with vagrant-hostmanager
+    if Vagrant.has_plugin?('vagrant-hostmanager')
+      cached_addresses = {}
+      config.hostmanager.ip_resolver = proc do |vm, _resolving_vm|
+        if cached_addresses[vm.name].nil? && vm.communicate.ready?
+          vm.communicate.execute("hostname -I | cut -d ' ' -f 2") do |_type, contents|
+            cached_addresses[vm.name] = contents.split("\n").first[/(\d+\.\d+\.\d+\.\d+)/, 1]
+          end
+        end
+        cached_addresses[vm.name]
+      end
+    end
+  elsif vconfig['vagrant_ip'] == '0.0.0.0' && Vagrant.has_plugin?('vagrant-auto_network')
+    # Use vagrant-auto_network to assign IP automatically
+    config.vm.network :private_network, ip: vconfig['vagrant_ip'], auto_network: true
+  else
+    # Static IP
+    config.vm.network :private_network, ip: vconfig['vagrant_ip']
+  end
+
+  # Allow public network access
   unless vconfig['vagrant_public_ip'].empty?
-    config.vm.network :public_network,
-      ip: vconfig['vagrant_public_ip'] != '0.0.0.0' ? vconfig['vagrant_public_ip'] : nil
+    if vconfig['vagrant_public_ip'] == '0.0.0.0' || vconfig['vagrant_public_ip'] == 'dhcp'
+      # Automatically assign IP
+      config.vm.network :public_network
+    else
+      # Static IP
+      config.vm.network :public_network, ip: vconfig['vagrant_public_ip']
+    end
   end
 
   # SSH options.
@@ -69,7 +96,7 @@ Vagrant.configure('2') do |config|
 
   # If a hostsfile manager plugin is installed, add all server names as aliases.
   aliases = get_vhost_aliases(vconfig) - [config.vm.hostname]
-  if Vagrant.has_plugin?('vagrant-hostsupdater')
+  if Vagrant.has_plugin?('vagrant-hostsupdater') && vconfig['vagrant_ip'] != 'dhcp'
     config.hostsupdater.aliases = aliases
   elsif Vagrant.has_plugin?('vagrant-hostmanager')
     config.hostmanager.enabled = true
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -55,15 +55,7 @@ Vagrant.configure('2') do |config|
 
     # Use assigned IP with vagrant-hostmanager
     if Vagrant.has_plugin?('vagrant-hostmanager')
-      cached_addresses = {}
-      config.hostmanager.ip_resolver = proc do |vm, _resolving_vm|
-        if cached_addresses[vm.name].nil? && vm.communicate.ready?
-          vm.communicate.execute("hostname -I | cut -d ' ' -f 2") do |_type, contents|
-            cached_addresses[vm.name] = contents.split("\n").first[/(\d+\.\d+\.\d+\.\d+)/, 1]
-          end
-        end
-        cached_addresses[vm.name]
-      end
+      config.hostmanager.ip_resolver = vagrant_ip_resolver
     end
   elsif vconfig['vagrant_ip'] == '0.0.0.0' && Vagrant.has_plugin?('vagrant-auto_network')
     # Use vagrant-auto_network to assign IP automatically
--- a/lib/drupalvm/vagrant.rb
+++ b/lib/drupalvm/vagrant.rb
@@ -72,6 +72,18 @@ def vagrant_provisioner
   ansible_bin ? :ansible : :ansible_local
 end
 
+def vagrant_ip_resolver
+  cached_addresses = {}
+  proc do |vm, _resolving_vm|
+    if cached_addresses[vm.name].nil? && vm.communicate.ready?
+      vm.communicate.execute("hostname -I | cut -d ' ' -f 2") do |_type, contents|
+        cached_addresses[vm.name] = contents.split("\n").first[/(\d+\.\d+\.\d+\.\d+)/, 1]
+      end
+    end
+    cached_addresses[vm.name]
+  end
+end
+
 def ensure_plugins(plugins)
   logger = Vagrant::UI::Colored.new
   installed = false
--- a/provisioning/templates/dashboard.html.j2
+++ b/provisioning/templates/dashboard.html.j2
@@ -21,7 +21,7 @@
   </head>
   <body>
     {# Sets vagrant IP if equals 0.0.0.0 #}
-    {%- if vagrant_ip == '0.0.0.0' -%}
+    {%- if vagrant_ip == '0.0.0.0' or vagrant_ip == 'dhcp' -%}
       {%- set vagrant_ip = ansible_all_ipv4_addresses[1] -%}
     {%- endif -%}
     {# Returns the hosts server name based on the document root #}


oxyc:
Personally I've managed the IP manually for all my projects at work because at some point `vagrant-auto_network` was causing issues. This was a long time ago though so maybe it isn't an issue anymore.

oxyc:
If interested, we need to verify what happens when this is updated on an existing project.

oxyc:
> If we decide to go with this, geerlingguy.mac-dev-playbook also needs to be updated for passwordless commands.

Thought they used the same commands, but no.

```sh
# Vagrant sudoers config
Cmnd_Alias VAGRANT_EXPORTS_ADD = /usr/bin/tee -a /etc/exports
Cmnd_Alias VAGRANT_NFSD = /sbin/nfsd restart
Cmnd_Alias VAGRANT_EXPORTS_REMOVE = /usr/local/bin/sed -E -e /*/ d -ibak /etc/exports
Cmnd_Alias VAGRANT_HOSTS_ADD = /bin/sh -c echo "*" >> /etc/hosts
Cmnd_Alias VAGRANT_HOSTS_REMOVE = /usr/local/bin/sed -i -e /*/ d /etc/hosts
Cmnd_Alias VAGRANT_HOSTMANAGER_UPDATE = /bin/cp /Users/oxy/.vagrant.d/tmp/hosts.local /etc/hosts
%admin ALL=(root) NOPASSWD: VAGRANT_EXPORTS_ADD, VAGRANT_NFSD, VAGRANT_EXPORTS_REMOVE, VAGRANT_HOSTS_ADD, VAGRANT_HOSTS_REMOVE, VAGRANT_HOSTMANAGER_UPDATE
```

> we need to verify what happens when this is updated on an existing project.

Tested, `vagrant reload` doesn't work and fails with `mount.nfs: access denied by server while mounting 172.28.128.1:/Users/oxy/Projects/Genero/genero`. Fix is to `vagrant halt; vagrant reload`

docs/other/networking.md
```diff
@@ -1,12 +1,15 @@
...
+## `vagrant-auto_network` plugin
 
-  1. Switch the `vagrant_ip` in `config.yml` to a different private IP address, e.g. `172.16.0.88` or `10.0.1.88`.
-  2. Install the `vagrant-auto_network` plugin (`vagrant plugin install vagrant-auto_network`), and set the `vagrant_ip` to `0.0.0.0`.
+If you prefer using the `vagrant-auto_network` plugin (`vagrant plugin install vagrant-auto_network`) to assign your IP, set the `vagrant_ip` to `0.0.0.0`.
+
+## Network Route Collision
 
-Another cause of route collisions is the use of multiple VM providers on your computer. If you have both VirtualBox and VMware Fusion, and you have VMs running in both, and you attempt to use the same IP range in both providers, you'll hit a networking conflict. In this case, the only easy way to restore connectivity is to restart your host machine.
+If you have multiple VM providers running VMs on your computer, and you attempt to use the same IP range in both providers, you'll hit a networking conflict. In this case, the only easy way to restore connectivity is to restart your host machine.
```
oxyc: I think this can be removed. I got this when I just attempted to launch two providers:

> An active machine was found with a different provider. Vagrant
currently allows each machine to be brought up with only a single
provider at a time. A future version will remove this limitation.
Until then, please destroy the existing machine to up with a new
provider.

oxyc: Maybe skip the entire file?

oxyc:
One issue to consider is https://github.com/geerlingguy/drupal-vm/issues/1381. Not sure if one or the other will work with Docker. 

Some initial testing actually shows that `vagrant-hostsupdater` works while `vagrant-hostmanager` doesn't.

I believe this is because hostmanager adds a provisioner, which will be disabled by vagrant when SSH isn't available.

oxyc:
Tried to track down the problem but couldn't. My Vagrant internal skills are not quite enough. I came as far as to figure out that even if the block runs, the `use` is never initialized https://github.com/devopsgroup-io/vagrant-hostmanager/blob/master/lib/vagrant-hostmanager/action.rb#L13

oxyc:
I tracked down the problem and filed an issue https://github.com/devopsgroup-io/vagrant-hostmanager/issues/243. This PR should probably not be merged if it's never going to work with Docker.

I rebased this PR into two separate commits:

- First one adds support to DHCP and defaults to `vagrant-hostmanager` if set.
- Second one defaults Drupal VM to DHCP and `vagrant-hostmanager`.

The latter commit should be dropped from this PR but I'm leaving it here for input. If the commit is dropped, this PR basically adds a new DHCP feature which requires `vagrant-hostmanager`. It's only triggered if `vagrant_ip: dhcp`

oxyc:
Let me know if you want me top drop the second commit so this PR is testable.

geerlingguy:
In general, I'm thumbs up for this, and if we want it in soon, I'd be willing to put it in what will be a minor version bump, the next release, since I can note that you may need to vagrant halt and vagrant up...—but is the failure only due to the mailhog bug?

oxyc:
No I think the failure will appear when you switch over to DHCP on an existing VM. But as it wont be the default a note in the release sound enough to me. It's basically just a feature addition with a slight quirk if you switch over on an existing project.

I don't have time at the moment but tomorrow morning I'll revert the second commit.

oxyc:
For reference here's the commit I removed https://github.com/oxyc/drupal-vm/commit/8fe94cee4d177ed8c106c422bc6367fa738cf537

swalkinshaw:
Does this work?

We [tried it](https://github.com/roots/trellis/issues/902) on Trellis and ran into some issues.

oxyc:
Thanks for the heads up @swalkinshaw.

I admit I only tested this on an already existing VM, but I just tested it and with Drupal VM it does work. `vagrant up` works with NFS and `vagrant-hostmanager` adds the host entries correctly. Unfortunately as of today I'm going offline for the next 3 months so I won't have time to look into what exactly makes it work in my case.

Some differences that come to mind:
- My VirtualBox version matches the Guest Additions installed on the box. We also use `vagrant-vbguest` to enforce this.
- We use NFS over TCP rather than UDP (https://github.com/geerlingguy/drupal-vm/issues/1468)
- I'm using linked clones

I'm running Vagrant v2.0.0

oxyc:
I just updated VirtualBox to 5.1.28r117968 to see what happens when the guest additions are upgraded and I was able to reproduce the error trellis gets.

```
Unmounting Virtualbox Guest Additions ISO from: /mnt
==> drupalvmoro: Checking for guest additions in VM...
==> drupalvmoro: [vagrant-hostsupdater] Checking for host entries
==> drupalvmoro: Setting hostname...
==> drupalvmoro: Configuring and enabling network interfaces...
No guest IP was given to the Vagrant core NFS helper. This is an
internal error that should be reported as a bug.
```

Seems `vagrant-vbguest` is the culprit.

oxyc:
I'm currently testing this again and it works fine so far with:

```rb
config.vbguest.auto_update = false
```

Vagrantfile
```diff
@@ -48,13 +48,40 @@ Vagrant.configure('2') do |config|
...
-  config.vm.network :private_network,
-    ip: vconfig['vagrant_ip'],
-    auto_network: vconfig['vagrant_ip'] == '0.0.0.0' && Vagrant.has_plugin?('vagrant-auto_network')
 
+  if vconfig['vagrant_ip'] == 'dhcp'
+    # Automatically assign IP
+    config.vm.network :private_network, type: 'dhcp'
+
+    # Use assigned IP with vagrant-hostmanager
+    if Vagrant.has_plugin?('vagrant-hostmanager')
```
geerlingguy: This section feels so complicated though... basically going from a few lines to 22 lines with some regex and shell commands thrown in (if I'm reading it correctly)?

geerlingguy: Is the only issue if you use `vagrant-hostmanager`? E.g. does it work currently if you have `vagrant-hostsupdater` instead?

geerlingguy:
@oxyc Doesn't:

```
config.vbguest.auto_update = false
```

Kind of defeat the purpose, and could cause a lot more problems with people with versions of VirtualBox different than the one the box is for?

oxyc:
> Kind of defeat the purpose, and could cause a lot more problems with people with versions of VirtualBox different than the one the box is for?

Yeah it's not something we should add to Drupal VM. I decided to try it and see if it's in fact was `vagrant-vbguest` that was causing the issue. So far my 3 work projects running with DHCP assigned IP are working great.

I also chose to patch in my PR for DHCP to our boilerplate at work because:

1. @thom8 mentioned he's been fine without updating vbguest for a while https://github.com/geerlingguy/drupal-vm/issues/1126#issuecomment-331681950
2. We're a small team and everyone is aware of my patch
3. We have so many projects that we often have IP collisions between VM's (doh)

Vagrantfile
oxyc: The scary looking `config.hostmanager.ip_resolver` proc is what makes `vagrant-hostmanager` pick up the DHCP assigned IPs from within the VM and then adds them to `/etc/hosts`.

oxyc: No, vagrant-hostsupdater doesn't work unless they add support for the same `ip_resolver` functionality. In which case this logic would probably be the same anyways. https://github.com/cogitatio/vagrant-hostsupdater/issues/56

oxyc:
I'd want to run a few more tests on this still but my vote goes for including it and adding some docs that `vagrant-vbguest` has compatibility issues. At least for me it's a great feature to have even if we can't switch to it as a default.

Also note that either way we probably shouldn't switch to this as a default until https://github.com/devopsgroup-io/vagrant-hostmanager/issues/243 is solved (if it's solved). It won't work with docker as a provider, which we might eventually want to support someday.

Vagrantfile
oxyc: @geerlingguy if you prefer not having this logic how would you feel about leaving this section as:

```rb
if vconfig['vagrant_ip'] == 'dhcp'
  # Automatically assign IP. See docs for integration with `vagrant-hostmanager`
  config.vm.network :private_network, type: 'dhcp'
elsif ...
```

And moving the snippet there and users can include it in their `Vagrantfile.local`.

oxyc: Since September we've been using this at work on 10+ VM's without `vagrant-vbguest`. No issues so far.

oxyc:
It works better now. We should definitely keep the old behavior until a major release but would be nice to allow users to use:

```yaml
vagrant_ip: 'dhcp'
```

- [x] Vagrant up first time with wrong vbguest version
- [x] Change to DHCP on an existing VM with wrong vbguest version: `vagrant halt; vagrant up`
- [ ] Change to DHCP on an existing VM with wrong vbguest version: `vagrant reload`
  _Produces the following error (fixed with `vagrant halt; vagrant up`):_
  > NFS requires a host-only network to be created.
  > Please add a host-only network to the machine (with either DHCP or a static IP) for NFS to work.  
- [ ] Change from DHCP back to IP on an existing VM:
  _This breaks pretty badly and I haven't found a way to switch back to IP._
- [x] Upgrade Virtualbox
- [x] Vagrant up first time with correct vbguest version

geerlingguy:

oxyc:
I'll still add that we've been using this on 20+ VM's with 4 developers since September 2017 and have had no issues so far.  The conflict with `vagrant-vbguest` should have been fixed but we opted for disabling it anyways due to the increased build time of VM's. No issues since 2017 with having it turned off either.

My vote still goes for including this but not as a default because of the two outstanding issues here https://github.com/geerlingguy/drupal-vm/pull/1559#issuecomment-387200433

PR #1560, opened 2017-09-23
Improve debug output in travis

oxyc:
Print body and headers if a test fails. Let's see what mailhog says.

- `-i` show headers in output
- `-I` fetch headers only
- `-s` hide progress bar (and error messages unless `-S`)
- `-S` show error messages.

### DIFF
--- a/tests/run-tests.sh
+++ b/tests/run-tests.sh
@@ -51,10 +51,15 @@ fi
 # Set OS-specific options.
 if [[ "$OSTYPE" == "darwin"* ]]; then
   volume_opts='rw,cached'
-  xargs_command='xargs'
 else
   volume_opts='rw'
+fi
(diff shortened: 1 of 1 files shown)


tests/run-tests.sh
```diff
@@ -114,15 +114,15 @@ fi
```
oxyc: derp something slipped. cron == curl

oxyc:
Finally trying this out locally to try and debug the mailhog issue.

Added an option to skip cleanup and fixed an issue when BSD xargs params were used even if I had GNU xargs installed (first time I'm actually hit by a bug because of having gnu tools, it's usually the other way around with the scripts i use).

geerlingguy:
LGTM

PR #1562, opened 2017-09-23
Fixes #1520: Let local.config.yml override environment.config.yml

### LINKED ISSUE #1520: local.config.yml should be able to override environment.config.yml -- for instance vagrant.config.yml
### Issue Type
 - Feature Idea
### Summary
I have a team of folks using drupalvm inside vagrant with very different host OS.  We need to share a vagrant.config.yml in our repo to keep our boxes similar.  We also want to allow devs to override settings.  unfortunately, vagrant.config.yml overrides anything we put in local.config.yml because it is read afterwards.  

Our workaround is to move shared vagrant config into config.yml, which we can do because we don't use config.yml to configure a production instance at this time.  Others may not be able to do that if they have conflicting per environment settings that need to be overridable by devs.

The solution would be to load a local.environment.config.yml if it exists to allow futher overriding.  Or if we don't mind sneakily changing behavior on people we could just change the order of loading in the vagrantfile to load local.config.yml last -- this is likely to make bugs of the worst kind though.

oxyc:


geerlingguy:
This is a slightly-breaking change (since I know only a few people use an environment.config.yml file right now, myself include), but I'm happy to add it because local config should always take precedence (IMO).

I'm going to merge it and we'll just add a warning to the minor version bump's release notes. Thanks!

Issue #1553, opened 2017-09-19
Deprecation warnings from Ansible if running Ansible 2.4+


geerlingguy:
### Issue Type

 - Bug Report

### Full console output

```
[DEPRECATION WARNING]: The use of 'include' for tasks has been deprecated. Use 
'import_tasks' for static inclusions or 'include_tasks' for dynamic inclusions.
 This feature will be removed in a future release. Deprecation warnings can be 
disabled by setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: include is kept for backwards compatibility but usage is
 discouraged. The module documentation details page may explain more about this
 rationale.. This feature will be removed in a future release. Deprecation 
warnings can be disabled by setting deprecation_warnings=False in ansible.cfg.
[DEPRECATION WARNING]: The use of 'static' has been deprecated. Use 
'import_tasks' for static inclusion, or 'include_tasks' for dynamic inclusion. 
This feature will be removed in a future release. Deprecation warnings can be 
disabled by setting deprecation_warnings=False in ansible.cfg.
 [WARNING]: file /Users/jeff.geerling/Dropbox/Development/GitHub/drupal-
vm/provisioning/roles/geerlingguy.apache-php-fpm/tasks/configure-RedHat.yml is
empty and had no tasks to include
```

### Summary

There are a few warnings that occur when running Drupal VM's playbook under Ansible 2.4.0+ (released today):

  - <s>[defaults]hostfile option, The key is misleading as it can also be a list of hosts, a directory or a list of paths . This feature will be removed in version 2.8.</s> (This was from my own ansible.cfg file, not Drupal VM's—fix is to switch `hostfile` to `inventory` under the `[defaults]` section in `ansible.cfg`.)
  - The use of 'include' for tasks has been deprecated. Use 'import_tasks' for static inclusions or 'include_tasks' for dynamic inclusions. This feature will be removed in a future release. (same with 'static)
  - <s>(And in Ansible 2.5+) [DEPRECATION WARNING]: State 'installed' is deprecated. Using state 'present' instead.. This feature will be removed in version 2.9.</s> Fixed.

[...]

PR #1563, opened 2017-09-23
Add debug mode and hide deprecation warnings by default

oxyc:
Issue #1553. I'm assuming you're not going to bump the version requirements of all your roles to 2.4.0. If you are we could drop the conditional `ansible.cfg` config.

- `ansible.cfg` was never loaded with Vagrant. It would need to reside next to the `Vagrantfile` to do so.
- `control_path` no longer needs to be set in `ansible.cfg` (read https://github.com/geerlingguy/drupal-vm/commit/633810ae76f19cedc3e204d56cd8d1c866c13926)
- Hide Vagrant 2.0+ messages about ansible auto compatibility mode. @geerlingguy could you test this on  vagrant < 2.0 to make sure it works? (read [docs](https://www.vagrantup.com/docs/provisioning/ansible_common.html#compatibility_mode))
- Add support for `DRUPALVM_DEBUG` environment variable to easily debug ansible tasks (see examples in https://github.com/geerlingguy/drupal-vm/commit/93abae9d3f05b1147a9071e256820887e10fe338)
- Hide deprecation warnings by default, unless DRUPALVM_DEBUG is truthy. I don't think users need to care about such things. This puts a bit more pressure on developers to stay up-to-date with ansible releases though but that shouldn't be a problem for @geerlingguy :) He has 1 or 2 other roles he maintains and learns about these things from. If you prefer I can drop this commit though.


oxyc:
I especially like the ability to tell users in the issue queue to run `DRUPALVM_DEBUG=v vagrant up`

geerlingguy:
My main question is what does the debug .cfg file provide over the normal one? It seems like it would just be the first few messages at the top of the Ansible playbook output—but is there any more? Maybe it's just my late-night read of the code...

oxyc:
> My main question is what does the debug .cfg file provide over the normal one? It seems like it would just be the first few messages at the top of the Ansible playbook output—but is there any more? 

No that's it but I wasn't aware of disabling depreciation notices any other way than with a config file. Is it possible somehow?

oxyc:
Also as I'm fixing the _bug_ where the `ansible.cfg` isn't taken into account at all currently, I thought it was better to introduce a new `ansible.debug.cfg` vs `ansible.cfg` setup so it's more clear. Let me know if you come up with something better.

geerlingguy:
I'll take a second look soon since I finally had some time to finish up the 4.7.0 release.

geerlingguy:
@oxyc - This looks great! One last suggestion—can we add something in the docs about setting `DRUPALVM_DEBUG=true` to enable debug mode?

geerlingguy:

geerlingguy:
(Note that I'm still for these changes, with the small addition of something about debugging with `DRUPALVM_DEBUG=v`.)

oxyc:
I'll reopen this and work on the docs (soon!)

PR #1564, opened 2017-09-24
Use shellwords hack to allow spaces in DRUPALVM_ANSIBLE_ARGS

oxyc:
https://gist.github.com/phantomwhale/9657134

This always bothered me.


geerlingguy:
I thought I had fixed that in Drupal VM (since I'm using it in some of my other playbooks), but apparently not. Thanks!

PR #1566, opened 2017-09-26
php_xdebug_cli_enable is unused

danepowell:
This is a follow up to https://github.com/geerlingguy/ansible-role-php-xdebug/pull/47

As indicated in that issue, `php_xdebug_cli_enable` is no longer used and references to it should be removed.

However, it's being replaced by a new option `php_xdebug_cli_disable` in https://github.com/geerlingguy/ansible-role-php-xdebug/pull/54. So it might make sense to just switch one for the other.

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -319,7 +319,7 @@ pimpmylog_grant_all_privs: true
 # XDebug configuration. XDebug is disabled by default for better performance.
 php_xdebug_default_enable: 0
 php_xdebug_coverage_enable: 0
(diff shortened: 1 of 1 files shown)


geerlingguy:
Thanks!

PieterDC:
Thanks for your contributions.
I just wanted to note that (looking at https://github.com/geerlingguy/drupal-vm/pull/1566/commits/7c3d210bc2296ee5b9fa1eff45475a9f5fd2cd12) it seems the default changed from enabled by default to disabled by default.

PR #1602, opened 2017-11-04
Dashboard header can be different if running multiple VM's.

petrillek:
I'm running multiple VM's at one time and have found out that I'm missing clear info about the VM on the dashboard. I've add one line to the dashboard.html.j2 which adds the Machine name in header. 

Originally I was thinking about creating a new key "vagrant_name" to get a "nice" name, but the machine name is better solution as it is already there and it should be always unique.

obrazek


geerlingguy:
LGTM!

PR #1619, opened 2017-11-24
Add an example of what to put in secrets.yml

brockfanning:
I grabbed the text for this change from: https://www.jeffgeerling.com/blog/2017/soup-nuts-using-drupal-vm-build-local-and-prod


geerlingguy:
Thanks for this nice little addition!

PR #1622, opened 2017-11-28
Move SSH tasks in pre_tasks session

denics:
When you develop your own modules and they are hosted in a private repository, you want to add your known_hosts before launching the drush make, or the vagrant process simply goes in timeout

### DIFF
--- a/provisioning/playbook.yml
+++ b/provisioning/playbook.yml
@@ -28,6 +28,7 @@
       with_fileglob: "{{ pre_provision_tasks_dir|default(omit) }}"
 
     - include: tasks/php.yml
+    - include: tasks/sshd.yml
 
   roles:
     # Essential roles.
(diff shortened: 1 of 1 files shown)


thom8:
@denics as a workaround you can duplicate this task in the pre-provision section -- http://docs.drupalvm.com/en/latest/extending/scripts/#ansible-task-files

It shouldn't matter if the task runs multiple times.

denics:
Thanks @thom8 , I just found this solution more elegant and inoffensive.

thom8:
Well if there's no dependencies in this tasks with variables set by the roles there's no reason it can't be run before the roles.

oxyc:
This sounds like a good addition. @geerlingguy?

geerlingguy:
It seems like a harmless enough change. Will wait until after the major Drush integration changes before merging though.

denics:
in the meantime this pull became updated should I open a new one?

geerlingguy:

PR #1624, opened 2017-11-29
Add beet/box as an available base box.

thom8:
The base box has been adjusted to work with Drupal VM so provisioning tasks will skip is they've been preprovisioned..

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -1,6 +1,6 @@
 ---
 # Available `vagrant_box` values include:
-#   - beet/box
+#   - beet/box (pre-provisioned, based on Ubuntu 16.04)
 #   - geerlingguy/centos7
 #   - geerlingguy/centos6
 #   - geerlingguy/debian9
--- a/default.config.yml
+++ b/default.config.yml
@@ -1,5 +1,6 @@
 ---
 # Available `vagrant_box` values include:
+#   - beet/box
 #   - geerlingguy/centos7
 #   - geerlingguy/centos6
 #   - geerlingguy/debian9


geerlingguy:
Just for my own sake—is `beet/box` based on the ubuntu 16.04 base, or on another distro? Haven't used beetbox in over a year, so I'm slightly out of date in my memory.

thom8:
Yep `16.04` -- https://github.com/beetboxvm/beetbox/blob/master/provisioning/ubuntu-16.04.json#L49

Did a bit of work a few month ago to add support for BLT so it should come with everything preprovisioned and supports Drupal VM provisioning on top of the base box.

danepowell:
If downstream projects are using Beetbox as the base box, should they still use Drupal VM's Vagrantfile and provisioning roles? Or should they include Beetbox's?

For instance, this is BLT's latest Vagrantfile. BLT now supports Beetbox as a base box, but it includes DrupalVM's Vagrantfile no matter what. Not sure if that's correct: https://github.com/acquia/blt/blob/95042f8cc72d729704b3f94c1af68e2a51a233f2/scripts/drupal-vm/Vagrantfile#L9

thom8:
@danepowell correct, the base box works for both projects. 

The main difference is the first provision time, which is useful if you destroy and rebuild frequently. The other major benefit is linked clones which reduces the size of each VM since all projects share a preprovisioned master VM and only need to track changes made, useful if you have many projects and limited disk.

However, as provisioning time can be only a couple of mins I'll often destroy projects when not in use a rebuild them on demand.

geerlingguy:
Before I merge, can you add into the comment something like:

```
#   - beet/box (pre-provisioned, based on Ubuntu 16.04)
```

Just because at least for _me_, it's important to know what base OS I'm targeting...

geerlingguy:
Looks good!

PR #1631, opened 2017-12-01
Added php_xdebug_idekey for Atom with PHP Debug

tormi:


oxyc:
I think it's better to add it to the docs instead of the config. Could you mention it in https://github.com/geerlingguy/drupal-vm/blob/master/docs/extras/xdebug.md instead?

tormi:
Good idea, @oxyc !

Issue #1469, opened 2017-07-07
Installing GoAccess?


kevinquillen:
### Issue Type
 - Feature Idea

### Summary

I would like to use [GoAccess](https://goaccess.io/) as a dev tool instead of PimpMyLog. How can I go about setting it up automatically when I create a new VM?

geerlingguy:
@kevinquillen - The best way currently would be to add a task include file (to use Ansible tasks) or a shell script to run post-provision: http://docs.drupalvm.com/en/latest/extending/scripts/

If you end up finding GoAccess to be a good drop-in replacement for PML, then I'm definitely not opposed to the idea of switching. I'm not married to PML, as I never use it myself!

But does GoAccess display log entries in a centralized way, or is it just a stats dashboard?

fubarhouse:
@geerlingguy GoAccess a dashboard of stats collected from input access logs.

Back when I was complaining about PML I switched over to GoAccess and found it incredibly valuable. It's got a more concise scope dealing strictly with access logs, and there is (minimal) configuration to happen - but it's a very useful tool.

I plugged 6 months worth of production logs into it 👍

bar0191:
@geerlingguy Where do I sign the petition to get PML switched out with GoAccess? That would be awesome if you're still considering it!

oxyc:
@fubarhouse do you have any example of your setup that you could post?

fubarhouse:
@oxyc I'll add some information about it on Tuesday morning AEST.

fubarhouse:
#Goaccess

## Install

I usually build mine from source, however 1.2 is available via the system package management utility for all linux-based systems.

Building from source. Note that the folder created must not be moved or changed after building.
````
sudo apt-get -y install libncursesw5-dev gcc make
sudo apt-get -y install libgeoip-dev libtokyocabinet-dev
wget http://tar.goaccess.io/goaccess-1.2.tar.gz
tar -xzvf goaccess-1.2.tar.gz
cd goaccess-1.2
sudo ./configure --enable-utf8 --enable-geoip
sudo make
sudo make install
sudo ln -s /usr/local/bin/goaccess /usr/bin/goaccess
````

It can now be used from the console, but it also features some really cool HTML generation tools...

For the html work, you'd need to point a port to a location, or a virtual host to a location to serve it.

## Config

The following three requirements must be specified in the goaccess.conf file.

[...]

PR #1647, opened 2018-01-03
Support for Goaccess

fubarhouse:
PR for #1469, which still needs work.

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -139,6 +139,10 @@ apache_vhosts:
     documentroot: "{{ pimpmylog_install_dir }}"
     extra_parameters: "{{ apache_vhost_php_fpm_parameters }}"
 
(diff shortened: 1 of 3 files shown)


fubarhouse:
I've updated the fork to continue work, and it's reaching the point it can work out of the box.

I've replaced the non-working daemon for a cron task run every minute.

The only thing left before you look at it would be a test and slight change for fedora/centos apache log file path or log format configuration. I think Fedora will be the only system affected, but I'll see to it.

geerlingguy:
Thanks! I think what I'd like to do here is create a separate goaccess role, and stash all the OS-specific bits in it. If you're okay with it, I can lift the config you've done here and attribute the initial commit in that new role to you.

geerlingguy:

geerlingguy:
(Note that I'd still like to switch to GoAccess at some point.)

PR #1648, opened 2018-01-04
1579 cableconnect with provisioning

rpsu:


rpsu:
More info on this over here: @see https://github.com/hashicorp/vagrant/issues/8157#issuecomment-304492943

oxyc:
Would you mind summarizing the issue and how this fixes it? Some steps to reproduce an issue would be great.

geerlingguy:
@rpsu - Same comment as oxyc—I haven't seen this issue crop up on any of my builds; is there an open issue related, or do you have any more details as to what problem you were having that this fixed? Also, would need to know version and OS details to reproduce the issue.

geerlingguy:
Closing this for now—please let us know if you can find a way to reproduce the issue or have a better description of why this change would be helpful to the project.

PR #1651, opened 2018-01-05
Update NPM by default

danepowell:
Normally, the NPM version installed by Node out of the box is pretty out of date.

This PR will cause it to automatically be updated to the latest available version (respective to whatever version of Node you are using). This is recommended and considered best practice by Node/NPM.

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -298,7 +298,8 @@ adminer_install_filename: index.php
 # Node.js configuration (if enabled above).
 # Valid examples: "0.10", "0.12", "4.x", "5.x", "6.x".
 nodejs_version: "6.x"
-nodejs_npm_global_packages: []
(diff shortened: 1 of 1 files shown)


oxyc:
Sounds good to me.

Reference:

> Node comes with npm installed so you should have a version of npm. However, npm gets updated more frequently than Node does, so you'll want to make sure it's the latest version.
> If the version is not the latest version, run:
> npm install npm@latest -g

https://docs.npmjs.com/getting-started/installing-node#2-update-npm

geerlingguy:
This makes sense (they didn't used to release new versions on practically a daily basis like they do nowadays), but why not also propose it as a default upstream? https://github.com/geerlingguy/ansible-role-nodejs

PR #1659, opened 2018-01-17
Caveat about group perms for web server with bindfs.

siliconmeadow:


geerlingguy:
Good catch! Sometimes we're a bit too Debian-centric in the examples :(

siliconmeadow:
\o/

PR #1708, opened 2018-02-27
Selenium paths value needs to be quoted

kmbremner:
With behat 3.4.3, the example config gives the following error on behat --init:

The reserved indicator "%" cannot start a plain scalar; you need to quote the scalar at line 4 (near "paths: [ %paths.base%/features/web ]").

This is fixed by quoting the paths value.


geerlingguy:
Sounds reasonable!

PR #1726, opened 2018-03-13
Make ansible version parsing compatible with releases from git repository

rsanzante:
When ansible is installed from repository version string is like: 

`ansible 2.4.3.0 (stable-2.4 8708105616) `

Instead of:

`ansible 2.4.3.0`

Regexp fails because it includes the part in parenthesis. Added new regexp that ignores that part. See http://rubular.com/r/J4TBshh16b to see it working.


geerlingguy:
Works for me.

PR #1734, opened 2018-03-19
create acquia.config.yml

hoff331:
See PF-1064 - To provide Acquia customers with a better local Drupal development experience, we should provide and (more importantly) maintain an acquia.config.yml file to emulate Acquia Cloud.

### DIFF
--- a/acquia.config.yml
+++ b/acquia.config.yml
@@ -1,0 +1,353 @@
+---
+# This configuration file is meant to emulate Acquia Cloud
+# To use, create a copy of this file and rename it to config.yml
+
+# Available `vagrant_box` values include:
+#   - beet/box (pre-provisioned, based on Ubuntu 16.04)
+#   - geerlingguy/centos7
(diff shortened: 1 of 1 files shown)


geerlingguy:
@hoff331 - Please see:

  - http://docs.drupalvm.com/en/latest/other/acquia/
  - https://github.com/geerlingguy/drupal-vm/commit/f4af1fa92994aeec28c3e77bb6bd606ac0745887

Basically, Drupal VM _used_ to include a special `acquia.config.yml` example, but since nobody kept it updated over time it got very much out of date (it was still on ubuntu1204 I think, when Acquia Cloud upgraded to 16.04).

The current recommendation is to either us Acquia BLT entirely to manage Drupal VM, or to grab the config.yml file that is included with BLT and use it with your own custom copy of Drupal VM: https://github.com/acquia/blt/blob/9.x/scripts/drupal-vm/config.yml

Issue #1736, opened 2018-03-20
Updated ansible version parsing fails with releases not from git repo


rootwork:
### Issue Type

 - Bug Report / Support Request

### Your Environment

```
Vagrant 2.0.3
VirtualBox 5.2.8r121009
ansible 2.4.3.0
  config file = None
  configured module search path = [u'/Users/foo/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/Cellar/ansible/2.4.3.0_4/libexec/lib/python2.7/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 2.7.14 (default, Mar 10 2018, 00:01:04) [GCC 4.2.1 Compatible Apple LLVM 9.0.0 (clang-900.0.39.2)]
```

### Your OS

 - **macOS** (10.12.6)

### Summary

#1726 changed the way the ansible version is parsed to allow for git repo releases, but it now includes extraneous strings when ansible is installed via pip or homebrew, then fails. You'll notice in "your environment" above that `ansible --version` returns not only the version but some other settings. That entire string is then passed into the version parsing, which fails. You can verify it here, using the regex from that change:

http://rubular.com/r/aFs8eiAxV7

A similar string is returned when ansible is installed via pip.

Prior to #1726, the version number was successfully parsed.

geerlingguy:
Gah! It will never be working 100% I guess :(

oxyc:
https://github.com/geerlingguy/drupal-vm/issues/1725#issuecomment-374705142

geerlingguy:
Fixed in PR #1740.

Issue #1725, opened 2018-03-13
I cannot initialize a new vm with ansible installed via git as normal user (with root it works).


dlopez-akalam:
### Issue Type
 - Bug Report / Support Request

### Your Environment

```
Vagrant 2.0.2
VirtualBox 5.1.30_Ubuntur118389
ansible 2.4.4 (stable-2.4 92cea82a4e) last updated 2018/02/28 20:16:43 (GMT +200)
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/home/david/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /opt/ansible/lib/ansible
  executable location = /opt/ansible/bin/ansible
  python version = 2.7.14 (default, Sep 23 2017, 22:06:14) [GCC 7.2.0]
```

### Your OS

 - **Linux** (Ubuntu 17.10)

### Full console output

```
vagrant up --debug
 INFO global: Vagrant version: 1.9.8
 INFO global: Ruby version: 2.3.4
 INFO global: RubyGems version: 2.5.2
 INFO global: VAGRANT_OLD_ENV_XDG_SESSION_DESKTOP="KDE"
 INFO global: VAGRANT_OLD_ENV_SHELL="/bin/bash"
 INFO global: VAGRANT_OLD_ENV_WINDOWID="83886086"
 INFO global: VAGRANT_OLD_ENV_PATH="/opt/ansible/bin:/opt/ansible/test/runner:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games:/snap/bin"
 INFO global: VAGRANT_OLD_ENV_LC_TIME="es_ES.UTF-8"
 INFO global: VAGRANT_EXECUTABLE="/opt/vagrant/embedded/gems/gems/vagrant-1.9.8/bin/vagrant"
 INFO global: VAGRANT_OLD_ENV_DISPLAY=":0"
 INFO global: VAGRANT_OLD_ENV_LC_ADDRESS="es_ES.UTF-8"
 INFO global: VAGRANT_OLD_ENV_SESSION_MANAGER="local/bree:@/tmp/.ICE-unix/1794,unix/bree:/tmp/.ICE-unix/1794"
 INFO global: VAGRANT_OLD_ENV_OLDPWD="/home/david/vagrant/blt-test/blt"
[...]

PR #1740, opened 2018-03-21
Issue #1736, Followup #1725: Fix ansible version parsing of non-git releases

oxyc:


geerlingguy:
Thank you! I just jumped in to work on this... glad I don't have to do anything besides merge the PR :)

PR #1744, opened 2018-03-27
XDebug IDE Key for VS Code

gobinathm:
Updated Xdebug comment for Visual Studio Code Usage


geerlingguy:
Let's delete that line entirely, then make sure that all the major editors are covered in the docs page; see source at https://github.com/geerlingguy/drupal-vm/blob/master/docs/extras/xdebug.md, and the docs page rendered at http://docs.drupalvm.com/en/latest/extras/xdebug/

gobinathm:
@geerlingguy update XDebug config documentations.

PR #1800, opened 2018-07-09
Allow Solr version argument to configure script

joestewart:
Allowing the version argument will reduce the need to copy and modify the script.


geerlingguy:
Are there any docs that would go along with this change to make it easier to use different Solr versions?

joestewart:
Probably can decline and close this PR since https://www.drupal.org/project/search_api_solr shows that the 8.x-1.x used by default in the script does not support 6.6.x nor 7.x.x.  So there would need to be more changes to support 8.x-2.x or 8.x-3-x.

geerlingguy:
Sounds good, thanks for the update!

PR #1808, opened 2018-07-29
Allow choosing the cron user per task

grahl:
Hi

The change itself should be self-explanatory. 

At the moment (as far as I can tell), it's not possible to change the user's crontab in drupalvm_cron_jobs. This would make it possible to run a job for example as www-data or another user.

Thanks for your review.

### DIFF
--- a/provisioning/tasks/cron.yml
+++ b/provisioning/tasks/cron.yml
@@ -9,5 +9,6 @@
     month: "{{ item.month | default('*') }}"
     job: "{{ item.job }}"
     state: "{{ item.state | default('present') }}"
+    user: "{{ item.user | default(ansible_user) }} "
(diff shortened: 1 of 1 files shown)


provisioning/tasks/cron.yml
```diff
@@ -9,5 +9,6 @@
     month: "{{ item.month | default('*') }}"
     job: "{{ item.job }}"
     state: "{{ item.state | default('present') }}"
+    user: "{{ item.user | default(ansible_user) }} "
```
geerlingguy: Can you set this to:

```yaml
    user: "{{ item.user | default(omit) }}"
```

So it just doesn't set it if none is provided? `ansible_user` can do... weird things sometimes.

geerlingguy:

Issue #1802, opened 2018-07-13
How to avoid writing host configuration to WINDOWS/system32/drivers/etc/hosts?


webprogrammierer:
### Issue Type

 - Bug Report / Support Request

### Your Environment

```
Vagrant 1.9.4
drupal-vm-4.9.0
```

### Your OS

 - **Windows 10**

### Full console output

**==> drupalvm: [vagrant-hostsupdater] Removing hosts**
vagrant.d/gems/2.2.5/gems/vagrant-hostsupdater-1.1.1.160/lib/vagrant-hostsupdater/HostsUpdater.rb:117:in `initialize': Permission denied @ rb_sysopen - C:/WINDOWS/system32/drivers/etc/hosts (Errno::EACCES)

### Summary

hostname_configure: false 

I set hostname_configure to false but it still writes the host coniguration to my hosts file.
Because the host file is blocked by Avira the Drupal VM aborts with an error on every vagrant up or vagrant halt.

How can I avoid that the host configuration is written to the windows host file (I will do this manually)???

geerlingguy:
The `hostname_configure` option actually just controls whether the `drupalvm.hostname` role is run or not, for now. It doesn't control whether the Vagrant hostsupdater plugin is used—it is always used at this point.

Before considering changing this default behavior, which would be a breaking change... what is the purpose of Avira? I've never heard of it and haven't seen it come up in any other issues before.

webprogrammierer:
The purpose of Avira is to protect the file "hosts" in Windows 10: No authorized access to hosts.
When I manually want to change the contents of hosts file I have to turn off Avira Live Protect before and turn on again afterwards. This is a great and important protection.

I do not understand what the hostname_configure is for? 

We all had to write the IP addresses into hosts file manually before hostname_configure exists. But we needed to do this only once! And there was no need to remove this entries even if one project was not used anymore ...

But now on every vagrant up or on vagrant halt the contents of hosts file is re-written. This is not necessary, it should be possible to turn this off. 

Or this solution would also be good: 'vagrant halt' should not remove anything from hosts file if this project still exists in config.yml of Drupal VM. If it there still exists no removal is necessary.
vagrant/drupal vm should only write into hosts file if the entries do not exists!

And it should be possible to turn off every writing to host file.

webprogrammierer:
When will this problem be solved and how?

geerlingguy:
[...]

PR #1821, opened 2018-09-05
Issue #1802, by Michel Settembrino: Make possible to avoid updating local hosts file.

Michel-Settembrino:
I provided a new variable called: 'disable_hostsupdater' with as default value false.
If you don't want to update your hosts file, then you should set this to true.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -44,19 +44,39 @@ Vagrant.require_version ">= #{vconfig['drupalvm_vagrant_version_min']}"
 
 ensure_plugins(vconfig['vagrant_plugins'])
 
+# Local host updater settings
+begin
+  disable_hostsupdater = vconfig.fetch('disable_hostsupdater')
+rescue KeyError
+  disable_hostsupdater = false
+end
+
 Vagrant.configure('2') do |config|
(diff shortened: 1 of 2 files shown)


Vagrantfile
```diff
@@ -44,19 +44,39 @@ Vagrant.require_version ">= #{vconfig['drupalvm_vagrant_version_min']}"
 
 ensure_plugins(vconfig['vagrant_plugins'])
 
+# Local host updater settings
+begin
+  disable_hostsupdater = vconfig.fetch('disable_hostsupdater')
+rescue KeyError
```
geerlingguy: Since this is in the defaults, it will always be defined—we can just use the line:

```
disable_hostsupdater = vconfig.fetch('disable_hostsupdater')
```

Michel-Settembrino: Thx for the suggestion. It's changed.

geerlingguy:
I'm going to go with the comment from Oscar here: https://github.com/geerlingguy/drupal-vm/issues/1802#issuecomment-421494119 — easier to just have anyone who wants to disable to behavior do it in a `Vagrantfile.local` instead of adding more duplicate code to our already-kinda-complicated ruby code.

PR #1827, opened 2018-09-27
Clean up Tideways documentation

danepowell:
I was trying to set up Tideways recently and got pretty confused because it wasn't clear what the relationship was between the Tideways extension and the native XHProf extension, especially given that Tideways recently split and now is called the "Tideways XHProf extension" :roll_eyes: 

Hopefully this clears things up a little.


geerlingguy:
LGTM, thanks!

Issue #1835, opened 2018-10-11
Should ` include_tasks` that are tagged `always` be `import_tasks` in `provisioning/playbook.yml`?


lo5an:
### Issue Type

 - Bug Report / Support Request

### Your Environment

```
Vagrant 2.1.2
VirtualBox 5.2.18r124319
ansible 2.5.5
  config file = None
  configured module search path = [u'/home/lmc/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /home/lmc/.local/lib/python2.7/site-packages/ansible
  executable location = /home/lmc/.local/bin/ansible
  python version = 2.7.12 (default, Dec  4 2017, 14:50:18) [GCC 5.4.0 20160609]
```

### Your OS

 - **Windows** 10 Enterprise 1803
 - **Linux** Centos 7

#### Running without tags works as expected

When I run the provisioning playbook on a remote machine without specifying any tags, I see the expected behavior:
```
$ DRUPALVM_ENV=dev ansible-playbook -i vm/inventory vendor/geerlingguy/drupal-vm/provisioning/playbook.yml -e "config_dir=$(pwd)/vm"  --ask-vault-pass --check
Vault password:

PLAY [all] *******************************************************************************************************************************************************************************************************

TASK [Gathering Facts] *******************************************************************************************************************************************************************************************
ok: [dvm-explore.dev.ec2.internal]

TASK [include_tasks] *********************************************************************************************************************************************************************************************
included: /mnt/c/DrupalVM/drupal-vm-explore/vendor/geerlingguy/drupal-vm/provisioning/tasks/config.yml for dvm-explore.dev.ec2.internal

TASK [Include OS-specific variables.] ****************************************************************************************************************************************************************************
ok: [dvm-explore.dev.ec2.internal]

TASK [Define config_dir.] ****************************************************************************************************************************************************************************************
skipping: [dvm-explore.dev.ec2.internal]

TASK [Include optional configuration files.] *********************************************************************************************************************************************************************
ok: [dvm-explore.dev.ec2.internal] => (item=/mnt/c/DrupalVM/drupal-vm-explore/vm/config.yml)
ok: [dvm-explore.dev.ec2.internal] => (item=/mnt/c/DrupalVM/drupal-vm-explore/vm/secrets.yml)
ok: [dvm-explore.dev.ec2.internal] => (item=/mnt/c/DrupalVM/drupal-vm-explore/vm/dev.config.yml)

[...]

PR #1836, opened 2018-10-11
Change include_tasks to import_tasks for always tasks.

lo5an:
Tags don't seem to be inherited with `include_tasks` so switch to `import_tasks` to work on #1835


geerlingguy:
Manually ran through those task files, and they should be safe to import instead of include, as all the vars would be locked in by runtime. Thanks for the PR!

PR #1846, opened 2018-10-17
Fixed typo in build-composer.yml that caused composer install to fail

clemit:
A bug was introduced in commit e8664798e08421db9fc85edfadaf7958750a8907 that causes composer install to fail when drupalVM was added as a dependency to an existing site.


geerlingguy:
falset == portmanteau of false and not?

Don't know what I was thinking 🤪

Issue #1894, opened 2019-02-06
Error in documentation


hepabolu:
### Issue Type

 - Documentation Report

### Summary

[http://docs.drupalvm.com/en/latest/extras/drush/](http://docs.drupalvm.com/en/latest/extras/drush/) contains an error:

> Drupal VM automatically generates a drush alias file in ~/.drush/drupalvm.aliases.drushrc.php (for Drush < 9.0.0) and ~/.drush/sites/drupalvm.site.yml (for Drupal 9.0.0+) with an alias for every site you have defined in the apache_vhosts or nginx_vhosts variable.

I assume _Drupal 9.0.0+_ needs to be _Drush 9.0.0+_

Michel-Settembrino:
Indeed it should be Drush.
I created a PR to update the documentation. See https://github.com/geerlingguy/drupal-vm/pull/1895

geerlingguy:
Good catch! And thanks @Michel-Settembrino for the PR. I've now merged it in.

PR #1895, opened 2019-02-07
Drush instead of Drupal

Michel-Settembrino:
Related to https://github.com/geerlingguy/drupal-vm/issues/1894.


hepabolu:
FYI the error is not yet fixed in http://docs.drupalvm.com/en/latest/extras/drush/

PR #1909, opened 2019-03-08
Port Collisions and Correction

markus-drupal:
This change allows multiple vms to run at the same time. Using auto_network plugin was not enough, as it seems to deal only with ip conflicts.


markus-drupal:
Not really sure why the test failed. :(

geerlingguy:
This seems like it would only work with forwarded ports...? https://www.vagrantup.com/docs/networking/forwarded_ports.html

Is the idea that you could have two Drupal VM instances, both with the same hostname (e.g. `mysite.test` and `mysite.test`), but one be on port 80 and another on port 81 or something?

PR #1968, opened 2019-09-09
Install pip for compatibility with Vagrant ^2.2.5

q0rban:
Initially running `vagrant up` gave me the following error:

```
    drupal-cms: Installing pip... (for Ansible installation)
The following SSH command responded with a non-zero exit status.
Vagrant assumes that this means the command failed!

pip install  --upgrade ansible

Stdout from the command:



Stderr from the command:

bash: line 5: pip: command not found
```

This led me to https://github.com/hashicorp/vagrant/issues/9584 and https://github.com/hashicorp/vagrant/issues/10950#issuecomment-514023810. The following change fixed it for me.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -109,6 +109,7 @@ Vagrant.configure('2') do |config|
   end
 
   config.vm.provision 'drupalvm', type: provisioner do |ansible|
+    ansible.pip_install_cmd = "curl https://bootstrap.pypa.io/get-pip.py | sudo python"
     ansible.compatibility_mode = '2.0'
     ansible.playbook = playbook
     ansible.extra_vars = {


AlannaBurke:
I had the same issue and this change fixed it for me!

q0rban:
Thanks @geerlingguy! I tried to do something similar in Vagrantfile.local and couldn't figure it out. Could you elaborate?

q0rban:
Also, regarding:

> especially one that sudo bash's a downloaded file

Giving this a closer read hashicorp/vagrant#9584:

1. Vagrant's ansible does the sudo bash to install pip. The "bug" in this case seems to be that  "ansible_local installs pip before it installs ansible"
2. @gildegoma says the reporter did, "the right thing [by] implementing your own ansible setup via a shell provisioner"
3. @gildegoma then proposes as a fix to, "introduce a new pip_install_command option, with
"curl https://bootstrap.pypa.io/get-pip.py | sudo python" as default."

To be clear, I'm all in favor of avoiding scripts piped to bash, but if I'm understanding, it seems avoiding that would mean writing your own ansible installer.

Perhaps a route forward for DrupalVM is [here](https://github.com/hashicorp/vagrant/issues/9584#issuecomment-374127034):

> you can extend your shell provisioner to also install ansible, and then configure the ansible_local provisioner to not install ansible by setting the install option to `false`.

q0rban:
Thoughts on the above @geerlingguy ?

PR #1997, opened 2019-12-05
Include php_xdebug_version in default.config.yml

joel-travieso:
Include php_xdebug_version in default.config.yml to provide visibility to this option.

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -319,6 +319,7 @@ pimpmylog_install_dir: /usr/share/php/pimpmylog
 pimpmylog_grant_all_privs: true
 
 # XDebug configuration. XDebug is disabled by default for better performance.
+# php_xdebug_version: 2.6.0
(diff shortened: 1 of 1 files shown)


geerlingguy:
Just wondering, what's the impetus here (why did you need to change the version from the default)?

joel-travieso:
> Just wondering, what's the impetus here (why did you need to change the version from the default)?

@geerlingguy  I found compatibility problems between the default version (2.6.0) and PHP 7.3. I struggled a bit to find a way to change that, so I thought adding it here would save others the time.

geerlingguy:
I'll merge this in but I'll also set the default in the playbook since the version of PHP Drupal VM defaults to may shift from what the upstream role defaults to over time, and it's good to test the versions in lockstep.

Note that the current upstream default is 2.9.0...

Issue #1897, opened 2019-02-11
Help installing PHP PECL "jsmin" extension? (redux)


greylabel:
### Issue Type

 - Bug Report / Support Request

### Your Environment
```
Vagrant 2.2.3
VirtualBox 6.0.4r128413
ansible 2.7.6
  config file = None
  configured module search path = ['/Users/username/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/Cellar/ansible/2.7.6/libexec/lib/python3.7/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.7.2 (default, Jan 13 2019, 12:50:01) [Clang 10.0.0 (clang-1000.11.45.5)]
```

### Your OS
 - **macOS** (10.14.3)

### Summary

I basically have the same support request as #888 -- help installing PHP PECL "jsmin" extension --which, IIUC from that issue, should be achievable. I feel I might just be having a head-desk moment here, but any guidance would be appreciated.

Setup steps
1. Clone a fresh copy of DrupalVM
2. `cp default.config.yml config.yml`
3. customize `config.yml` (see below)
4. run `vagrant up --provision`

Adding the below to `config.yml`:
```
php_pecl_extensions:
  - jsmin
```

Fails with the below error:
```
TASK [geerlingguy.php-pecl : Include OS-Specific variables.] *******************
ok: [drupalvm] => (item=/Users/username/Projects/tmp/drupal-vm/provisioning/roles/geerlingguy.php-pecl/vars/Debian.yml)

TASK [geerlingguy.php-pecl : Ensure pecl is installed (if configured).] ********
skipping: [drupalvm]

TASK [geerlingguy.php-pecl : Install PECL libaries.] ***************************
failed: [drupalvm] (item=jsmin) => {"changed": false, "cmd": "yes '' | pecl install jsmin", "delta": "0:00:00.001933", "end": "2019-02-11 03:55:12.605043", "failed_when_result": true, "item": "jsmin", "msg": "non-zero return code", "rc": 127, "start": "2019-02-11 03:55:12.603110", "stderr": "/bin/sh: 1: pecl: not found", "stderr_lines": ["/bin/sh: 1: pecl: not found"], "stdout": "", "stdout_lines": []}
	to retry, use: --limit @/Users/username/Projects/tmp/drupal-vm/provisioning/playbook.retry

PLAY RECAP *********************************************************************
[...]

PR #2018, opened 2020-02-29
Add Debian10 for pecl.

grahl:
When I use the base image geerlingguy/drupal-vm and want to install pecl the installation fails since it looks for the php-pecl package which should be php-pear.

Just from looking at other roles I presume that this change is sufficient but at least on the site I initially tested with, adding that file did not produce the desired change and I edited Debian.yml there, too.

This should fix some of the issues originally reported in #1897.

### DIFF
--- a/provisioning/roles/geerlingguy.php-pecl/.travis.yml
+++ b/provisioning/roles/geerlingguy.php-pecl/.travis.yml
@@ -9,6 +9,7 @@ env:
     - MOLECULE_DISTRO: centos7
     - MOLECULE_DISTRO: ubuntu1804
     - MOLECULE_DISTRO: debian9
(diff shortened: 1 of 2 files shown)


geerlingguy:
@grahl - Note that any changes to the roles themselves should go against the actual role repositories—can you look there: https://github.com/geerlingguy/ansible-role-php-pecl

geerlingguy:
And note that this _should_ be fixed by https://github.com/geerlingguy/ansible-role-php-pecl/commit/730d53c049188d0205fae25b185f0163b5cd1240 — I'll try to update that role in Drupal VM soon (it will be in master until the next Drupal VM release).

grahl:
Thanks for the info and the fix there, didn't think to go looking for that repo.

PR #2041, opened 2020-04-25
🐍 point to the Python 3 interpreter in PATH

codextypes:
For Mac developers using [Homebrew](https://brew.sh/), this edit will be helpful as most of them won't have Python 3 in `/usr/bin`

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -45,7 +45,7 @@ vagrant_plugins:
   - name: vagrant-vbguest
   - name: vagrant-hostsupdater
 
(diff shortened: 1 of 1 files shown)


geerlingguy:
Good point—As long as it works inside the VM as well, it's fine to leave out the path. I think I just did this since that was the path of Python on the system where I was testing. My own laptop shows installs at both paths (but defaults to the Homebrew installation!).

Regardless, Linux users may tend to use venvs for Python, which would be somewhere else altogether.

PR #2060, opened 2020-07-04
Update Linux documentation with Debian 10 instructions

rubenvarela:
Instructions include how to install NFS and increase boot timeout. 

I've been helping someone out with this and they've had both issues. I''ve been recreating from a MacOS host so maybe the timeout I'm using could be lowered but they're also seeing the timeout issue.

More info on my setup, 
- MacOS host (Catalina 10.15.5, VirtualBox 6.1.10) running a Debian 10.4.0 guest. Debian there was installed using the netinstall, gnome desktop on `tasksel`. It then acts as a host which in turns runs vagrant (2.2.9) +virtualbox (6.1.10) with an Ubuntu guest (geerlingguy/drupal-vm) 

If someone wants to test recreate and has similar setup, I'll add that on the Mac host I had to enable the "Enable Nested VT-x/AMD-V" checkbox. It shows as disabled but running `$ VBoxManage modifyvm vm-name --nested-hw-virt on` works, [source](https://stackoverflow.com/a/56973830/1392131). I didn't test if I needed "Enable PAE/NX", but I also enabled it.

### DIFF
--- a/docs/getting-started/installation-linux.md
+++ b/docs/getting-started/installation-linux.md
@@ -14,9 +14,13 @@ Some older Ubuntu Desktop installations don't include NFS support by default, so
 
 ### Debian
 
-A regular desktop install of Debian 10 doesn't include NFS. You'll see a message, `It appers your machine doesn't support nfs, or there is not an adapter to enable NFS on this machine for Vagrant. Please verify that `nfsd`is installed on your machine, and try again.`. You need to install `sudo apt install -y nfs-common nfs-kernel-server`. After that, you should be able to restart the machine
+A regular desktop install of Debian 10 doesn't include NFS. You'll see a message,
 
-If you see `drupalvm: Warning: Connection reset. Retrying...` and it ends with it timing out. You should be able to create a `Vagrantfile.local` and add `config.vm.boot_timeout = 600`. The default value is 300, so we are just doubling it. If it still comes back, you can increase it as needed.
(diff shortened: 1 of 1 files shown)


geerlingguy:
Later on the page there are more [notes about NFS installation and problems with NFS](http://docs.drupalvm.com/en/latest/getting-started/installation-linux/#mounting-nfs-shared-folders-hangs), which also mentions the fact that many distros need it installed. Maybe we can just fix up that section if it is missing some of this information? I'd rather not duplicate the information for ever distro.

rubenvarela:
Sounds good. I can take a look at it. 

I'll do a basic install of Ubuntu, see what's needed there to make it work. See if we can consolidate anything too.

NYCdude2:
Hi, I followed a guide to "secure" my Debian installation so it is likely that as your docs say at: 
http://docs.drupalvm.com/en/latest/getting-started/installation-linux/#mounting-nfs-shared-folders-hangs
that the file/drive was encrypted.  
In the future, if I am using a security style Debian or just a Ubuntu OS with encryption, how do I allow the VM by itself operate in an unencrypted file?  
As a novice, I would try right clicking the file to try to get an unencrypt option.  
Perhaps there is a sandbox type thing that can be run on those security distros which allow for the VM & a specified file/dir to operate without encryption in an otherwise encrypted host OS/drive.  
Please share your thoughts.  Thanks.

PR #2100, opened 2020-10-27
Fixes #2099: Prestissimo not needed with Composer 2.x

### LINKED ISSUE #2099: Provision dies at geeringguy.composer task
### With composer 2, `hirak/prestissimo` is no longer needed and default provisioning fails

### Your Environment

```
Vagrant 2.2.10
VirtualBox 6.1.14r140239
zsh: command not found: ansible
```

### Your OS

 - **macOS** (version)
### Full console output

https://gist.github.com/siliconmeadow/f4b298e5856b644de8e94b81ad53fb57

### Summary

When onboarding an existing project which was built via Acquia's BLT today, I got the error as per the full console output as per the gist. I managed to resolve it by removing `hirak/prestissimo` from the `default.config.yml` file, so thought I'd report that here and submit a PR.

siliconmeadow:
Fixes #2099 

I don't know if there is a more sophisticated way to deal with this, but by default, Composer 2.x is installed on new projects, and this solution got me past the provisioning error. Composer 2 has `hirak/prestissimo` 's parallel download feature built in now.

### DIFF
--- a/default.config.yml
+++ b/default.config.yml
@@ -278,8 +278,7 @@ composer_path: /usr/bin/composer
 composer_home_path: "/home/{{ drupalvm_user }}/.composer"
 composer_home_owner: "{{ drupalvm_user }}"
 composer_home_group: "{{ drupalvm_user }}"
-composer_global_packages:
-  - { name: hirak/prestissimo, release: '^0.3' }
+composer_global_packages: []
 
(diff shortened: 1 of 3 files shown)


geerlingguy:
Quite correct! I'll merge this as CI builds started failing here as well. Thanks!

PR #2108, opened 2020-11-20
Add option to disable nfs_export

joelpittet:
Allow DrupalVM to leave the /etc/exports file alone and let me manage it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ The `hirak/prestissimo` package was removed from `composer_global_packages` (and
 
 ### Improvements
 
-  * N/A
+  * #2108: Default synced folder nfs_export to `vagrant_synced_folder_default_nfs_export` if unset.
 
 ### Bugfixes
 
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -100,6 +100,7 @@ Vagrant.configure('2') do |config|
       id: synced_folder['id'],
       create: synced_folder.fetch('create', false),
       mount_options: synced_folder.fetch('mount_options', []),
+      nfs_export: synced_folder.fetch('nfs_export', vconfig['vagrant_synced_folder_default_nfs_export']),
       nfs_udp: synced_folder.fetch('nfs_udp', false)
     }
     synced_folder.fetch('options_override', {}).each do |key, value|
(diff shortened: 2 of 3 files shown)


joelpittet:
https://www.vagrantup.com/docs/synced-folders/nfs#nfs_export

geerlingguy:
I think the idea with Drupal VM is to provide sensible defaults that are easily overridden, and in this case, there are already two options that can do the same thing as this one option:

Changing [the `type` in any synced folders](https://github.com/geerlingguy/drupal-vm/blob/master/default.config.yml#L33) to anything other than `nfs` disables that export (e.g. `''` uses built-in shared folders, or `smb` uses Samba), and then changing the [default `vagrant_synced_folder_default_type `](https://github.com/geerlingguy/drupal-vm/blob/master/default.config.yml#L10) to `''` would also disable that NFS export.

This ability to disable NFS is not explicitly documented somewhere, but there are references to it in NFS troubleshooting pages in the docs (e.g. the Windows NFS troubleshooting here: http://docs.drupalvm.com/en/latest/other/performance/#nfs).

I'd rather we just make the current behavior and how to change it to something other than NFS more obvious than add yet another variable to the defaults (especially one that could add to more confusion if you had `type` set to `nfs` but this new variable overrode that and you ended up not getting an export added).

joelpittet:
I'm looking for a manual override here, the PR should have the sane defaults intact. I still want to use NFS I just don't want it to write to that file and I'll take care of it.

oxyc:
You can also create a `Vagrantfile.local` with:

```rb
  config.vm.synced_folders.each do |path, synced_folder|
    next unless synced_folder[:type] == :nfs

    config.vm.synced_folders[path][: nfs_export] = true
  end
```

I use something similar to this to automatically swap to vboxsf on windows.

geerlingguy:
@joelpittet - See, if you bring a complicated enough problem up, eventually @oxyc joins in and brings up an even better solution 😂

I'm going to recommend that instead of adding a new var in this case, just because I think the use case of "I want to use NFS, but manage exports manually (or in some other way that's not automated through Vagrant)" is more narrow and can be worked around using @oxyc's custom Vagrantfile.local suggestion.

geerlingguy:
(I also need to get this repo moved over to GitHub Actions stat!)

joelpittet:
Thanks I'll give that a try, didn't know I could do that, thanks @oxyc and @geerlingguy !

PR #2121, opened 2021-01-05
Replace php_opcache_enabled_in_ini with php_opcache_enable

phizev:
The `php_opcache_enabled_in_ini` was removed from [ansible-role-php](https://github.com/geerlingguy/ansible-role-php) in [commit b9f0a61](https://github.com/geerlingguy/ansible-role-php/commit/b9f0a6172000fb834a95fb30f5c46cafe03f4331).

This means that within Drupal VM, the only place I can find `php_opcache_enabled_in_ini` is in default.config.yml. Switching this out for the `php_opcache_enable` will allow toggling the state of the opcache.


geerlingguy:
Good catch—thanks for finding this!

PR #2137, opened 2021-01-26
fix: add ansible install_mode "pip3"

rockymountainhigh1943:
This fixes an issue that was recently created by the release of a new version of the script here: https://bootstrap.pypa.io/get-pip.py.

TL;DR — they dropped support for Python2 and this now requires pip3. Maybe worth noting I may have had to manually install Python3 via SSH into the box, but I may have also already had it and it was a step I just duplicated. Either way — this fixes issues folks should be having as of this last Saturday.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -111,6 +111,7 @@ Vagrant.configure('2') do |config|
   config.vm.provision 'drupalvm', type: provisioner do |ansible|
     ansible.compatibility_mode = '2.0'
     ansible.playbook = playbook
+    ansible.install_mode = "pip3"
     ansible.extra_vars = {
       config_dir: config_dir,
       drupalvm_env: drupalvm_env,
@@ -119,7 +120,7 @@ Vagrant.configure('2') do |config|
     ansible.raw_arguments = Shellwords.shellsplit(ENV['DRUPALVM_ANSIBLE_ARGS']) if ENV['DRUPALVM_ANSIBLE_ARGS']
     ansible.tags = ENV['DRUPALVM_ANSIBLE_TAGS']
     # Use pip to get the latest Ansible version when using ansible_local.
-    provisioner == :ansible_local && ansible.install_mode = 'pip'
+    provisioner == :ansible_local && ansible.install_mode = 'pip3'
   end
 
   # VMware Fusion.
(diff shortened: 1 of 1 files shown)


siliconmeadow:
Fixes #2138 - tried it out and provisioning works tickety-boo.

very-random-man:
I had a quick play and whilst Python3.6.9 is installed, pip3 doesn't seem to exist on the base box by default.

siliconmeadow:
> I had a quick play and whilst Python3.6.9 is installed, pip3 doesn't seem to exist on the base box by default.

I may be misunderstanding what you're saying, but I see the same thing. But my understanding is that `pip3` gets installed by ansible when you bring the box up or do a `vagrant provision`. With this PR anyway. Without the PR, I suspect `pip` probably doesn't exist on the box either, but the provisioning pulls it down.

Vagrantfile
```diff
@@ -111,6 +111,7 @@ Vagrant.configure('2') do |config|
   config.vm.provision 'drupalvm', type: provisioner do |ansible|
     ansible.compatibility_mode = '2.0'
     ansible.playbook = playbook
+    ansible.install_mode = 'pip3'
```
geerlingguy, line 114: Ahhhh... in my previous readings I was thinking this was changing something inside the VM with the pip install that may happen as a result of Drupal VM's playbooks.

This is changing the Vagrantfile so that installs of _ansible_ in the VM if you need it installed inside will work correctly.

Thanks for catching this!

shaikz: Thank you!  Provisioning works 👍

jasloe:
Might've misunderstood something here or in #2138 but with dev-master, I'm still seeing the following on vagrant up and vagrant provision:

```
ansible remote provisioner:
* The following settings shouldn't exist: install_mode
```

My local environment: 

```
Vagrant 2.2.14
VirtualBox 6.1.8r137981
ansible 2.9.1
  config file = None
  configured module search path = ['/Users/jasloe/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.9/site-packages/ansible
  executable location = /usr/local/bin/ansible
  python version = 3.9.1 (default, Jan 30 2021, 15:51:26) [Clang 11.0.0 (clang-1100.0.33.17)]
```

geerlingguy:
@jasloe - Oops! Sorry about that; it should only apply for `ansible_local`, not `ansible`, so I believe the change I just pushed (see commit linked above) should fix it for those of us with Ansible running locally.

PR #2143, opened 2021-02-03
Remove elastic search heap options dependency

sonkkeli:
Remove line `/etc/elasticsearch/jvm.options.d/heap.options`. This line causes the whole build to fail since it doesn't exist.


geerlingguy:
If this change needs to be made, it should be made in the upstream geerlingguy.elasticsearch repository: https://github.com/geerlingguy/ansible-role-elasticsearch

PR #2145, opened 2021-02-08
Fixes #2144: Change docs on native synced folders to use explicit type

### LINKED ISSUE #2144: Setting type: "" for synced folders no longer valid.
### Issue Type

Documentation Report

### Summary

Since https://github.com/hashicorp/vagrant/pull/11359 (latest vagrant) ([docs](https://www.vagrantup.com/docs/synced-folders/basic_usage) changed as well at some point), using `type: ""` no longer works for synced folders and it should instead be omitted entirely or set to eg `:virtualbox`.

I know we mention this at least in http://docs.drupalvm.com/en/latest/getting-started/syncing-folders/#using-native-synced-folders

oxyc:
Another option is to only set `type` if truthy.

### DIFF
--- a/docs/getting-started/syncing-folders.md
+++ b/docs/getting-started/syncing-folders.md
@@ -45,13 +45,13 @@ There are a number of issues people encounter with synced folders from time to t
 
 ### Using Native Synced Folders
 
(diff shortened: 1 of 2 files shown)


geerlingguy:
Thanks; that makes sense.

Issue #2154, opened 2021-03-10
Guest hangs after 5 minutes of inactivity on macOS Big Sur host


mcroyle:
### Issue Type

 - Bug Report / Support Request

### Your Environment

```
Vagrant 2.2.14
Virtualbox 6.1.18r142142
```

### Your OS
 - **macOS** (Big Sur 11.2.3)

### Summary
I'm using the "Drupal VM as a Composer Dependency" setup for a project. Multiple colleagues are experiencing issues where when the VM is left idle (no interaction in the guests terminal or accessing a site via the web) for approximately 5 minutes, everything hangs. Commands such as vagrant ssh from the host gets you into the guest as far as showing you the last login but no prompt, interacting within the guest just hangs, or accessing the website yields a timeout. I am running the same project on my machine with the same versions of Vagrant and Virtualbox, however macOS (Catalina 10.15.7). I have not experienced any of these issues.

Running vagrant halt && vagrant up gets us back to a functional state again until these periods of inactivity. However, running vagrant suspend && vagrant up does not and it remains non-functional.

My focus in troubleshooting / issue searching has been around NFS timeouts but this only turns up results for Windows machines. I've compared macOS Catalina /etc/exports to the /etc/exports in macOS Big Sur with no differences, I've permitted VIrtualbox to have Full Disk access, and reviewed the Virtualbox logs but nothing stands out.

At this point I'm unsure of what else I can check / test and hoping for some ideas or direction.

purplecones:
I also get this same exact issue. Also running the same vagrant and virtualbox versions but still on version 11.2.1 of MacOS. Running `vagrant reload` fixes it but it's annoying because it will hang again if I don't interact with the VM within a few minutes.

gregj0:
Multiple people on our team are seeing something similar on these combinations
**Your Environment**
Vagrant 2.2.14
Virtualbox 6.1.16r140961 (we have other issues with 6.1.18)
**Your OS**
macOS (Big Sur 11.2.3)

Seeing it on 
drupal-vm 6.0.2/Centos7
and
drupal-vm 6.0.3/Centos8

geerlingguy:
Strange; it seems like something that could be a bug in Vagrant, but haven't seen any bug reports in that issue queue that would help give any clues.

oxyc:
I have a colleague with the same problem and we couldn't figure it out :/ I think it was the NFS mount that became unavailable.

[...]

PR #2162, opened 2021-05-04
Issue #2154: Add vagrant-nfs-fix role to fix NFS timeout issues on macOS

oxyc:
It's been a long time since I did anything ansible let alone init.d scripts. Therefore this needs reviews.

It's based on the firewall role, with the inital fix from @MokDevelopment and the systemd service by @susannecoates.

1. Is there an issue running it as root?
2. Can we somehow detect if the provisioning host is macos?
3. I'm just using `touch` since I'm guessing a timestamp modification should be enough? As I type this, I'm just realizing maybe I should add back the `rm` so the file doesnt accidentally get committed to git on the host.
4. I dont have diskspace to build any of the RHEL boxes :/ Could someone try it out?
5. I don't actually see the issue so someone should test it.

Co-authored-by: MokDevelopment
Co-authored-by: Susanne Coates


provisioning/roles/drupalvm.vagrant-nfs-fix/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+vagrant_nfs_fix_keepalive_file: "/vagrant/.vagrant-nfs-fix-keepalive.tmp"
```
geerlingguy, line 2:
```suggestion
vagrant_nfs_fix_keepalive_file: "/vagrant/.vagrant-nfs-fix-keepalive.tmp"

```

provisioning/roles/drupalvm.vagrant-nfs-fix/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+    mode: 0755
+
+- name: Copy Vagrant NFS fix init script into place.
+  template:
+    src: vagrant-nfs-fix.init.j2
+    dest: /etc/init.d/vagrant-nfs-fix
+    owner: root
+    group: root
+    mode: 0755
+  when: ansible_service_mgr != 'systemd'
```
geerlingguy, line 17: I believe all supported OSes now include systemd, so we can drop the init script... right?

provisioning/roles/drupalvm.vagrant-nfs-fix/tasks/main.yml
```diff
@@ -0,0 +1,32 @@
...
+    owner: root
+    group: root
+    mode: 0644
+  when: ansible_service_mgr == 'systemd'
+
+- name: Configure the NFS fix service.
+  service:
+    name: vagrant-nfs-fix
+    state: started
+    enabled: true
```
geerlingguy, line 32:
```suggestion
    enabled: true

```

geerlingguy (changes requested):
Looks fine, though I think we can ditch the init script as Drupal VM doesn't support any OS version old enough to not have systemd anymore.

provisioning/roles/drupalvm.vagrant-nfs-fix/templates/vagrant-nfs-fix.unit.j2
```diff
@@ -9,4 +9,4 @@ RestartSec=1
 ExecStart=/usr/sbin/vagrant-nfs-fix
 
 [Install]
-WantedBy=multi-user.target
\ No newline at end of file
+WantedBy=multi-user.target
```
oxyc, line 12: @geerlingguy are this fine or do you want the actual empty line at the end?

default.config.yml
```diff
@@ -42,6 +42,10 @@ vagrant_plugins:
   - name: vagrant-vbguest
   - name: vagrant-hostsupdater
 
+# Automatically fix timeout issues with NFS on macOS. See
+# https://github.com/geerlingguy/drupal-vm/issues/2154
+vagrant_nfs_fix_enabled: "{{ vagrant_synced_folder_default_type == 'nfs' }}"
```
oxyc, line 47: Still not sure if this should really run by default just because it's nfs. Maybe it should be opt-in?

susannecoates: Since it seems to only be a problem on Big Sur, maybe it should be opt in?

geerlingguy: Yeah... maybe just set it false and we can add a note in the docs under this page? http://docs.drupalvm.com/en/latest/getting-started/installation-macos/ like:

## Big Sur

If you are using NFS shares on macOS Big Sur, and you have problems with the NFS mount going away, please set `vagrant_nfs_fix_enabled: true` in your `config.yml`.

Something like that.

susannecoates: Yep, sounds good!

oxyc:
While I have you, it would be much appreciated if you could give https://github.com/geerlingguy/packer-drupal-vm/issues/5 another try btw! It built successfully for me back when I tried at least.

geerlingguy (approved):

marcelovani:
Had the same issue after upgrading to BigSur. Thanks for the hack to fix it. I wonder how lot its going to take for Vagrant to do a proper fix.

PR #2186, opened 2021-11-05
Fixes #2185

### LINKED ISSUE #2185: Re-wording of the recipe
### Issue Type

### Your Environment

```

```

### Your OS

 - **macOS** (version)
 - **Windows** (version)
 - **Linux** (version)

### Summary

The recipe for adminer reads `Ensure paths to place symlinks exist` which is poor wording. It should be `Determine paths to place existing symlinks.` I also put a full stop at the end for consistency. Patch along with PR is following.

kartagis:


geerlingguy:
Role fixes should be filed against role repositories directly (for geerlingguy.adminer)—but note that the task name is correct but should have a full stop. If the directories do not exist, it will create them. If they exist, it will just make sure and move on.

